PR #22, opened 2017-01-27
Add data to summary tab

nwrogers:


src/mainwindow.cpp
```diff
@@ -33,15 +33,16 @@
...
 #include "flamegraph.h"
+#include "models/summarydata.h"
 
 MainWindow::MainWindow(QWidget *parent) :
     QMainWindow(parent),
     ui(new Ui::MainWindow),
     m_bottomUpCostModel(new CostModel(this)),
-    m_parser(new PerfParser(this))
+    m_parser(new PerfParser(this)),
+    m_topHotspotsCostModel(new CostModel(this))
```
milianw: this is not needed, the top 'N' entries uses the same data, and thus the same model, as the bottomUpCostModel. only a proxy is missing (cf. topproxy in heaptrack)

src/mainwindow.cpp
```diff
@@ -122,3 +141,38 @@ void MainWindow::hideLoadingResults()
     ui->openFileProgressBar->hide();
     ui->loadingResultsLabel->hide();
 }
+
+QString MainWindow::formatTimeString(quint64 nanoseconds)
```
milianw: move this out of the MainWindow class context into an anonymous namespace above the first method implementation in this file. Then first construct temporaries for the integral values and only construct the strings later on as needed

```
namespace {
QString formatTimeString(quint64 nanoseconds)
{
    quint64 totalSeconds = nanoseconds / 1000000000;
    quint64 days = ...;
    quint64 hours = ...;
 
    auto format = [] (quint64 fragment, int precision = 2) -> QString {
        return QString::number(fragment).rightJustified(precision, '0');
    };
    auto optional = [format] (quint64 fragment) -> QString {
        return fragment > 0 ? format(fragment) + QLatin1Char(":") : QString();
    };
    return optional(days) + optional(hours) + optional(minutes) + format(seconds) + QLatin1Char('.') + format(milliseconds, 3);
}
}
```

src/mainwindow.cpp
```diff
@@ -122,3 +141,38 @@ void MainWindow::hideLoadingResults()
...
+    return dataA.selfCost > dataB.selfCost;
+}
+
+const FrameData& MainWindow::getTopHotspots(const FrameData& data)
+{
+    int numberOfTopHotspots = 5;
+
+    // Sort and limit the number of top hotspots
+    QVector<FrameData> orderedHotspotVector = data.children;
+    qSort(orderedHotspotVector.begin(), orderedHotspotVector.end(), compareHotspotsBySelfCost);
```
milianw: for future code: prefer the STL, i.e. std::sort

src/mainwindow.cpp
```diff
@@ -122,3 +141,38 @@ void MainWindow::hideLoadingResults()
...
+    QString seconds = QString::number(totalSeconds % 60).rightJustified(2, '0');
+    QString milliseconds = QString::number((nanoseconds / 1000000) % 1000).rightJustified(3, '0');
+
+    return (days == "00" ? "" : days + ":")
+           + (hours == "00" ? "" : hours + ":")
+           + (minutes == "00" ? "" : minutes + ":")
+           + seconds + "." + milliseconds;
+}
+
+bool compareHotspotsBySelfCost(FrameData dataA, FrameData dataB)
```
milianw: remove, sorting should be handled by a proxy model

src/mainwindow.cpp
```diff
@@ -56,10 +57,28 @@ MainWindow::MainWindow(QWidget *parent) :
...
+    ui->topHotspotsTableView->hideColumn(2);
+    ui->topHotspotsTableView->hideColumn(3);
+    ui->topHotspotsTableView->hideColumn(5);
+    ui->topHotspotsTableView->sortByColumn(CostModel::SelfCost);
+
     connect(m_parser, &PerfParser::bottomUpDataAvailable,
             this, [this] (const FrameData& data) {
                 m_bottomUpCostModel->setData(data);
                 ui->flameGraph->setBottomUpData(data);
+                m_topHotspotsCostModel->setData(getTopHotspots(data));
```
milianw: shouldn't be needed

src/mainwindow.cpp
```diff
@@ -122,3 +141,38 @@ void MainWindow::hideLoadingResults()
...
+           + (minutes == "00" ? "" : minutes + ":")
+           + seconds + "." + milliseconds;
+}
+
+bool compareHotspotsBySelfCost(FrameData dataA, FrameData dataB)
+{
+    return dataA.selfCost > dataB.selfCost;
+}
+
+const FrameData& MainWindow::getTopHotspots(const FrameData& data)
```
milianw: remove

src/mainwindow.h
```diff
@@ -56,8 +59,13 @@ private slots:
 private:
     Ui::MainWindow *ui;
     CostModel* m_bottomUpCostModel;
+    CostModel* m_topHotspotsCostModel;
     PerfParser* m_parser;
+    FrameData m_topHotspotResult;
```
milianw: remove, this should be part of the hotspots cost model

src/mainwindow.cpp
```diff
@@ -56,10 +57,28 @@ MainWindow::MainWindow(QWidget *parent) :
     ui->bottomUpTreeView->sortByColumn(CostModel::SelfCost);
     ui->bottomUpTreeView->setModel(proxy);
 
+    auto topHotspotsProxy = new QSortFilterProxyModel(this);
```
milianw: this sorts, and we need another level of proxy: a TopProxy (cf. heaptrack) that removes the child rows to make a flat table out of a tree. Additionally, it should hide the unwanted columns. So we have this chain:

view -> qsfpm -> topproxy -> m_bottomUpCostModel

src/models/summarydata.h
```diff
@@ -0,0 +1,41 @@
+/*
+  summarydata.h
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2016-2017 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
```
milianw: remove 2016 from here

src/models/summarydata.h
```diff
@@ -0,0 +1,41 @@
...
+struct SummaryData
+{
+    quint64 applicationRunningTime;
+    quint32 threadCount;
+    quint32 processCount;
+};
+
+Q_DECLARE_TYPEINFO(SummaryData, Q_MOVABLE_TYPE);
+
+
```
milianw: remove empty line

src/models/summarydata.h
```diff
@@ -0,0 +1,41 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#include <QVector>
```
milianw: include QTypeInfo (afaik?)

src/parsers/perf/perfparser.cpp
```diff
@@ -636,6 +663,11 @@ struct PerfParserPrivate
     QVector<QString> strings;
     QProcess process;
     QString parserBinary;
+    SummaryData summaryResult;
+    quint64 applicationStartTime;
```
milianw: missing initialization, I'd set both to = 0;

src/parsers/perf/perfparser.cpp
```diff
@@ -636,6 +663,11 @@ struct PerfParserPrivate
...
+    quint64 applicationEndTime;
+    QVector<quint32> uniqueThreads;
```
milianw: these two should be QSet

src/parsers/perf/perfparser.cpp
```diff
@@ -606,6 +610,29 @@ struct PerfParserPrivate
...
 
+    void addSampleToSummary(const Sample& sample)
+    {
+        if (sample.time < applicationStartTime || applicationStartTime == 0) {
+            applicationStartTime = sample.time;
+        }
+        else if (sample.time > applicationEndTime || applicationEndTime == 0) {
+            applicationEndTime = sample.time;
+        }
+        if (!uniqueThreads.contains(sample.tid)) {
```
milianw: once these are sets, simply do `uniqueThreads.insert(sample.tid)` - it will either insert when not exist, or error out. Doing a contains check first then inserting is going to be just as slow/fast, except when we actually insert - then we do the contains check twice.

src/parsers/perf/perfparser.cpp
```diff
@@ -606,6 +610,29 @@ struct PerfParserPrivate
...
+            uniqueThreads.append(sample.tid);
+        }
+        if (!uniqueProcess.contains(sample.pid)) {
```
milianw: dito

src/parsers/perf/perfparser.cpp
```diff
@@ -606,6 +610,29 @@ struct PerfParserPrivate
...
+            uniqueProcess.append(sample.pid);
+        }
+    }
+
+    void calculateSummary()
+    {
+        summaryResult.applicationRunningTime = applicationEndTime - applicationStartTime;
+        summaryResult.threadCount = uniqueThreads.count();
```
milianw: prefer the generic `size` method, which also exists in STL containers

src/mainwindow.cpp
```diff
@@ -33,9 +33,30 @@
...
+    quint64 seconds = totalSeconds % 60;
+    quint64 milliseconds = (nanoseconds / 1000000) % 1000;
+
+    auto format = [] (quint64 fragment, int precision = 2) -> QString {
+        return QString::number(fragment).rightJustified(precision, '0');
+    };
+    auto optional = [format] (quint64 fragment) -> QString {
+        return fragment > 0 ? format(fragment) + QLatin1Char(':') : QString();
+    };
+    return optional(days) + optional(hours) + optional(minutes) + format(seconds) + QLatin1Char('.') + format(milliseconds, 3);
```
milianw: add a newline in here somewhere (in the code) to break this long line, i.e. something like

```
    return optional(days) + optional(hours) + optional(minutes)
           + format(seconds) + QLatin1Char('.') + format(milliseconds, 3);
```

src/models/topproxy.cpp
```diff
@@ -0,0 +1,46 @@
...
+TopProxy::~TopProxy() = default;
+
+bool TopProxy::filterAcceptsColumn(int source_column, const QModelIndex& /*source_parent*/) const
+{
+    return source_column == CostModel::Symbol || source_column == CostModel::Binary || source_column == CostModel::SelfCost;
+}
+
+int TopProxy::rowCount(const QModelIndex& parent) const
+{
+    return 5;
```
milianw: it will return at most 5, no? so this has to be something like:

```
if (parent.isValid()) {
    return 0; // this is a flat list, not a tree
}
return std::min(5, sourceModel()->rowCount(parent));
```

PR #26, opened 2017-02-14
Add Caller / Callee functionality

### LINKED ISSUE #5: aggregated caller/callee view
Performance call graph data should also be displayed in a way similar to what KCachegrind does, or what VTune shows in it's caller/callee view:

- aggregate data by symbol name
- add one entry per symbol to a flat table view showing inclusive/exclusive cost

nwrogers:
Implement logic to perfparser to aggregate by symbol the caller/calle performance results from the perf data.  Add support to GUI to display the results in a table view with a new callercalleemodel.  Fixes #5

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -45,6 +45,7 @@
 #include "models/costmodel.h"
 #include "models/summarydata.h"
 #include "models/topproxy.h"
+#include "models/callercalleemodel.h"
 
 namespace {
 QString formatTimeString(quint64 nanoseconds)
@@ -119,6 +120,12 @@ MainWindow::MainWindow(QWidget *parent) :
     ui->topHotspotsTableView->setSortingEnabled(false);
     ui->topHotspotsTableView->setModel(topHotspotsProxy);
 
+    auto callerCalleeCostModel = new CallerCalleeModel(this);
+    auto proxy = new QSortFilterProxyModel(this);
+    proxy->setSourceModel(callerCalleeCostModel);
+    ui->callerCalleeTableView->setSortingEnabled(true);
+    ui->callerCalleeTableView->setModel(proxy);
+
     setStyleSheet(QStringLiteral("QMainWindow { background: url(:/images/kdabproducts.png) top right no-repeat; }"));
 
     connect(m_parser, &PerfParser::bottomUpDataAvailable,
@@ -149,6 +156,12 @@ MainWindow::MainWindow(QWidget *parent) :
                 } else {
                     ui->lostMessage->setVisible(false);
                 }
+            });
+
+    connect(m_parser, &PerfParser::callerCalleeDataAvailable,
+            this, [this, callerCalleeCostModel] (const FrameData& data) {
+                callerCalleeCostModel->setData(data);
+                ui->callerCalleeTableView->sortByColumn(CallerCalleeModel::InclusiveCost);
             });
 
     connect(m_parser, &PerfParser::parsingFinished,
--- a/src/models/callercalleemodel.cpp
+++ b/src/models/callercalleemodel.cpp
@@ -1,0 +1,140 @@
+/*
+  callercalleemodel.cpp
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2017 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Nate Rogers <nate.rogers@kdab.com>
+
+  Licensees holding valid commercial KDAB Hotspot licenses may use this file in
+  accordance with Hotspot Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "callercalleemodel.h"
+
+CallerCalleeModel::CallerCalleeModel(QObject* parent)
+    : QAbstractTableModel(parent)
+{
+}
+
+CallerCalleeModel::~CallerCalleeModel() = default;
+
+int CallerCalleeModel::columnCount(const QModelIndex& parent) const
+{
+    return parent.isValid() ? 0 : NUM_COLUMNS;
+}
+
+int CallerCalleeModel::rowCount(const QModelIndex& parent) const
+{
+    return parent.isValid() ? 0 : m_root.children.size();
+}
+
+QVariant CallerCalleeModel::headerData(int section, Qt::Orientation orientation, int role) const
+{
+    if (role == Qt::InitialSortOrderRole) {
+        if (section == SelfCost || section == InclusiveCost)
+        {
+            return Qt::DescendingOrder;
+        }
+    }
+    if (role != Qt::DisplayRole || orientation != Qt::Horizontal) {
+        return {};
+    }
+
+    switch (static_cast<Columns>(section)) {
+        case Symbol:
+            return tr("Symbol");
+        case Binary:
+            return tr("Binary");
+        case Location:
+            return tr("Location");
+        case Address:
+            return tr("Address");
+        case SelfCost:
+            return tr("Self Cost");
+        case InclusiveCost:
+            return tr("Inclusive Cost");
+        case NUM_COLUMNS:
+            // fall-through
+            break;
+    }
+
+    return {};
+}
+
+QVariant CallerCalleeModel::data(const QModelIndex& index, int role) const
+{
+    if (!hasIndex(index.row(), index.column(), index.parent())) {
+        return {};
+    }
+
+    const auto& item = m_root.children.at(index.row());
+
+    if (role == SortRole) {
+        switch (static_cast<Columns>(index.column())) {
+            case Symbol:
+                return item.symbol;
+            case Binary:
+                return item.binary;
+            case Location:
+                return item.location;
+            case Address:
+                return item.address;
+            case SelfCost:
+                return item.selfCost;
+            case InclusiveCost:
+                return item.inclusiveCost;
+            case NUM_COLUMNS:
+                // do nothing
+                break;
+        }
+    } else if (role == FilterRole) {
+        // TODO: optimize this
+        return QString(item.symbol + item.binary + item.location);
+    } else if (role == Qt::DisplayRole) {
+        // TODO: show fractional cost
+        switch (static_cast<Columns>(index.column())) {
+            case Symbol:
+                return item.symbol;
+            case Binary:
+                return item.binary;
+            case Location:
+                return item.location;
+            case Address:
+                return item.address;
+            case SelfCost:
+                return item.selfCost;
+            case InclusiveCost:
+                return item.inclusiveCost;
+            case NUM_COLUMNS:
+                // do nothing
+                break;
+        }
+    }
+
+    // TODO: tooltips
+
+    return {};
+}
+
+void CallerCalleeModel::setData(const FrameData& data)
+{
+    beginResetModel();
+    m_root = data;
+    endResetModel();
+}
--- a/src/models/callercalleemodel.h
+++ b/src/models/callercalleemodel.h
@@ -1,0 +1,67 @@
+/*
+  callercalleemodel.h
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2017 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Nate Rogers <nate.rogers@kdab.com>
+
+  Licensees holding valid commercial KDAB Hotspot licenses may use this file in
+  accordance with Hotspot Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#include <QVector>
+#include <QAbstractItemModel>
+
+#include "framedata.h"
+
+class CallerCalleeModel : public QAbstractTableModel
+{
+    Q_OBJECT
+public:
+    explicit CallerCalleeModel(QObject* parent = nullptr);
+    ~CallerCalleeModel();
+
+    enum Columns {
+        Symbol = 0,
+        Binary,
+        Address,
+        Location,
+        SelfCost,
+        InclusiveCost,
+        NUM_COLUMNS
+    };
+
+    enum Roles {
+        SortRole = Qt::UserRole,
+        FilterRole
+    };
+
+    int columnCount(const QModelIndex& parent = {}) const override;
+    int rowCount(const QModelIndex &parent = {}) const override;
+
+    QVariant headerData(int section, Qt::Orientation orientation = Qt::Horizontal,
+                        int role = Qt::DisplayRole) const override;
+    QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+
+    void setData(const FrameData& data);
+private:
+    FrameData m_root;
+};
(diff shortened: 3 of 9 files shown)


src/main.cpp
```diff
@@ -39,6 +40,7 @@ int main(int argc, char** argv)
     QApplication app(argc, argv);
     qRegisterMetaType<FrameData>();
     qRegisterMetaType<SummaryData>();
+    qRegisterMetaType <QVector<FrameData>>();
```
milianw: remove space in front of the <

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,143 @@
...
+            break;
+    }
+
+    return {};
+}
+
+QVariant CallerCalleeModel::data(const QModelIndex& index, int role) const
+{
+    const auto* item = &m_rows.at(index.row());
+    //const auto& item = (role == MaxCostRole) ? m_maxCost : m_rows.at(index.row());
```
milianw: remove

src/models/callercalleemodel.h
```diff
@@ -0,0 +1,71 @@
...
+*/
+
+#pragma once
+
+#include <QVector>
+#include <QAbstractItemModel>
+
+#include "framedata.h"
+
+//using CallerCalleeRows = QVector<CallerCalleeData>;
```
milianw: remove

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +863,38 @@ struct PerfParserPrivate
         setParents(&topDownResult.children, nullptr);
     }
 
+    static void buildCallerCalleeResult(const QVector<FrameData>& bottomUpData, QVector<FrameData>& callerCalleeData)
+    {
+        static QSet<QString> callerCalleeRecursionGuard;
```
milianw: This must not be static, I actually doubt that this algorithm produces the correct results. Please check the current master of heaptrack for how to to this, and test with a recursive test application.

If you need help, let me know. I'll try to find some time later tonight to write a proper unit test for this which you can use then as basis. It's about time I start with unit test coverage, it will make your life much easier.

The problem is the following: The bottom up data contains all call graphs in a "merged" way, but the recursion guard must only be used for individual call graphs. You can get from the "merged"  tree of call stacks to a single stack by doing a depth-first search and then bubbling up the parent chain. I.e. the individual stacks are the parent-chains of every leaf in the bottomUpData tree. Now for every individual stack you'll have to guard against repeated counts of costs in the face of recursion, so you'll clear the set whenever you hit the next leaf before bubbling up the parent chain.

The comment I wrote initially didn't explicitly say that, so sorry if that wasn't clear.

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+    int idx = path.lastIndexOf(QLatin1Char('/'));
+    return path.mid(idx + 1);
+}
+}
+
+CallerCalleeModel::CallerCalleeModel(QObject* parent)
+    : QAbstractTableModel(parent)
+{
+    //qRegisterMetaType<CallerCalleeRows>();
```
milianw: remove

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,143 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "callercalleemodel.h"
+
+namespace{
+QString basename(const QString& path)
```
milianw: unused, remove

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+QVariant CallerCalleeModel::data(const QModelIndex& index, int role) const
+{
+    const auto* item = &m_rows.at(index.row());
+    //const auto& item = (role == MaxCostRole) ? m_maxCost : m_rows.at(index.row());
+
+    if (!item) {
+        return {};
+    }
+
```
milianw: add a check on `role == Qt::InitialSortOrderRole` and `return Qt::DescendingOrder;` for the cost columns

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +863,38 @@ struct PerfParserPrivate
...
+                newFrame.location = row.location;
+                newFrame.address = row.address;
+                newFrame.selfCost = 1;
+                newFrame.inclusiveCost = 1;
+                callerCalleeData.push_back(newFrame);
+                callerCalleeRecursionGuard.insert(row.location);
+            } else {
+                for (FrameData& callerCalleeRow : callerCalleeData) {
+                    if (callerCalleeRow.location == row.location) {
+                        callerCalleeRow.inclusiveCost++;
```
milianw: this is still potentially counting the same thing multiple times, i.e. when you have a recursive stack

    a > b > a > b

then the first two frames will not yet be in the set and thus added to the list, but then the last two frames will get their inclusive cost incremented while that shouldn't be the case for the caller/callee data representation.

src/models/callercalleemodel.cpp
nwrogers: Made the requested change in headerData

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,138 @@
...
+}
+
+CallerCalleeModel::~CallerCalleeModel() = default;
+
+QVariant CallerCalleeModel::headerData(int section, Qt::Orientation orientation, int role) const
+{
+    if (role != Qt::DisplayRole || orientation != Qt::Horizontal) {
+        return {};
+    }
+    if (role == Qt::InitialSortOrderRole) {
```
milianw: this will never be called b/c you have an early return above (`role != Qt::DisplayRole`)

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,138 @@
...
+            case SelfCost:
+                return item->selfCost;
+            case InclusiveCost:
+                return item->inclusiveCost;
+            case NUM_COLUMNS:
+                // do nothing
+                break;
+        }
+    } else if (role == FilterRole) {
+        return item->symbol + item->binary + item->location;
```
milianw: please add a `// TODO: optimize this`

src/models/callercalleemodel.h
```diff
@@ -0,0 +1,67 @@
...
+    QVariant headerData(int section, Qt::Orientation orientation = Qt::Horizontal,
+                        int role = Qt::DisplayRole) const override;
+    QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+
+    int columnCount(const QModelIndex& parent = {}) const override;
+    int rowCount(const QModelIndex &parent = {}) const override;
+
+    void resetData(const QVector<FrameData>& rows);
+private:
+    QVector<FrameData> m_rows;
```
milianw: I'd say use a FrameData here as "root" which holds the rows in it's `children`. The advantage is that this will allow us to track the total cost in the root object, for use in calculating the fractional cost.

src/main.cpp
```diff
@@ -39,6 +40,7 @@ int main(int argc, char** argv)
     QApplication app(argc, argv);
     qRegisterMetaType<FrameData>();
     qRegisterMetaType<SummaryData>();
+    qRegisterMetaType<QVector<FrameData>>();
```
milianw: if you use FrameData as root in the model, these change is not needed

src/mainwindow.cpp
```diff
@@ -151,6 +158,12 @@ MainWindow::MainWindow(QWidget *parent) :
                 }
             });
 
+    connect(m_parser, &PerfParser::callerCalleeDataAvailable,
+            this, [this, callerCalleeCostModel] (const QVector<FrameData>& data) {
+                callerCalleeCostModel->resetData(data);
+                ui->callerCalleeTableView->sortByColumn(CallerCalleeModel::InclusiveCost);
```
milianw: this shouldn't be needed

src/models/callercalleemodel.cpp
milianw: also, wrap it in a `QString(...)` to make the conversion from QStringBuilder explicit. I just pushed some changes which will make this necessary

src/mainwindow.cpp
nwrogers: If this isn't here it sorts by symbol, is that what we want?

src/main.cpp
```diff
@@ -28,6 +28,7 @@
 #include <QApplication>
 #include <QCommandLineParser>
 #include <QFile>
+#include <QVector>
```
milianw: not needed anymore

src/mainwindow.cpp
```diff
@@ -151,6 +158,12 @@ MainWindow::MainWindow(QWidget *parent) :
                 }
             });
 
+    connect(m_parser, &PerfParser::callerCalleeDataAvailable,
+            this, [this, callerCalleeCostModel] (const FrameData& data) {
+                callerCalleeCostModel->setData(data);
+                ui->callerCalleeTableView->sortByColumn(CallerCalleeModel::InclusiveCost);
```
milianw: still shouldn't be required

src/models/callercalleemodel.cpp
```diff
@@ -0,0 +1,203 @@
...
+CallerCalleeModel::CallerCalleeModel(QObject* parent)
+    : QAbstractTableModel(parent)
+{
+}
+
+CallerCalleeModel::~CallerCalleeModel() = default;
+
+int CallerCalleeModel::rowCount(const QModelIndex& parent) const
+{
+    if (parent.column() > 1) {
```
milianw: `>=`

the test found this issue in the existing model for me

src/models/callercalleemodel.h
```diff
@@ -0,0 +1,73 @@
...
+
+    QVariant headerData(int section, Qt::Orientation orientation = Qt::Horizontal,
+                        int role = Qt::DisplayRole) const override;
+    QVariant data(const QModelIndex& index, int role = Qt::DisplayRole) const override;
+
+    int columnCount(const QModelIndex& parent = {}) const override;
+    int rowCount(const QModelIndex &parent = {}) const override;
+
+    QModelIndex index(int row, int column, const QModelIndex& parent) const override;
+    QModelIndex parent(const QModelIndex& child) const override;
```
milianw: you don't need to supply neither index nor parent at all for a QAbstractTableModel. It has default implementations that are fine for us (afaik?)

since this is not a tree, accessing the data is trivial: `m_rows.value(index.row())`

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,33 @@ struct SymbolData
...
 };
 
+struct CallerCalleeLocation
+{
+    QString symbol;
+    QString binary;
+
+    bool operator<(const CallerCalleeLocation& location) const
+    {
+        return std::tie(symbol, binary)
```
milianw: join with next line (or is it too long? in that case, one more space in the line below, so that the `std::tie` aligns

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,33 @@ struct SymbolData
     }
 };
 
+struct CallerCalleeLocation
+{
+    QString symbol;
+    QString binary;
+
+    bool operator<(const CallerCalleeLocation& location) const
```
milianw: & next to variable name

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,33 @@ struct SymbolData
...
+inline bool operator==(const CallerCalleeLocation &location1, const CallerCalleeLocation &location2)
+{
+    return location1.symbol == location2.symbol
+           && location1.binary == location2.binary;
+}
+
+inline uint qHash(const CallerCalleeLocation &key)
+{
+    return qHash(key.symbol) ^ qHash(key.binary);
+}
```
milianw: empty line after

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,33 @@ struct SymbolData
...
+
+inline bool operator==(const CallerCalleeLocation &location1, const CallerCalleeLocation &location2)
+{
+    return location1.symbol == location2.symbol
+           && location1.binary == location2.binary;
+}
+
+inline uint qHash(const CallerCalleeLocation &key)
+{
+    return qHash(key.symbol) ^ qHash(key.binary);
```
milianw: please use a hash combine

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,33 @@ struct SymbolData
...
+    }
+};
+
+inline bool operator==(const CallerCalleeLocation &location1, const CallerCalleeLocation &location2)
+{
+    return location1.symbol == location2.symbol
+           && location1.binary == location2.binary;
+}
+
+inline uint qHash(const CallerCalleeLocation &key)
```
milianw: add a `, uint seed = 0` at the end

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,33 @@ struct SymbolData
...
+{
+    return qHash(key.symbol) ^ qHash(key.binary);
+}
+inline uint qHash(const CallerCalleeLocation &key, uint seed)
```
milianw: remove

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +890,44 @@ struct PerfParserPrivate
...
+        for (const FrameData& row : bottomUpData) {
+            if (row.children.isEmpty()) {
+                // leaf node found, bubble up the parent chain to add cost for all frames
+                // to the caller/callee data. this is done top-down since we must not count
+                // symbols more than once in the caller-callee data
+                QSet<CallerCalleeLocation> recursionGuard;
+                auto node = &row;
+
+                while (node) {
+                    const auto needle = std::tie(node->symbol, node->binary);
```
milianw: use CallerCalleeLocation here instead of tuple, now that we have it

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +890,44 @@ struct PerfParserPrivate
...
+                    if (!recursionGuard.contains({node->symbol, node->binary})) { // aggregate caller-callee data
```
milianw: use needle

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +890,44 @@ struct PerfParserPrivate
...
+                        auto it = std::lower_bound(callerCalleeData->children.begin(), callerCalleeData->children.end(), needle,
+                            [](const FrameData& frame, const auto needle) { return std::tie(frame.symbol, frame.binary) < needle; });
```
milianw: you implemented `operator<` above, so use it:

    return CallerCalleeLocation(frame.symbol, frame.binary) < needle;

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +890,44 @@ struct PerfParserPrivate
...
+
+                        if (it == callerCalleeData->children.end() || std::tie(it->symbol, it->binary) != needle) {
```
milianw: probably also replace tie with CallerCalleeLocation

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +890,44 @@ struct PerfParserPrivate
...
+                            it = callerCalleeData->children.insert(it, {node->symbol, node->binary, node->location, node->address, 0, 0, {}, nullptr});
+                        }
+                        it->inclusiveCost += 1;
+                        if (!node->parent) {
+                            it->selfCost += 1;
+
```
milianw: remove empty line

src/parsers/perf/perfparser.cpp
```diff
@@ -863,24 +890,44 @@ struct PerfParserPrivate
...
+                        }
+                        recursionGuard.insert({node->symbol, node->binary});
```
milianw: needle

src/mainwindow.cpp
milianw: I should read better. this is fine of course, ignore the comment above

src/models/callercalleemodel.h
nwrogers: As it turns out this is half true.  True, I didn't need to override parent but I do need to override index, otherwise I get a segfaults in CallerCalleeModel::itemFromIndex.

src/parsers/perf/perfparser.cpp
nwrogers: This is consistent with your coding style in heaptrack, for example: bool operator<(const LocationData::Ptr& rhs) const

in analyze/gui/parser.cpp

And in Qt code for example: bool operator<(const Module& module) const

in perfsymboltable.cpp

or am I misunderstanding what you mean here?

milianw: Qt coding style has the & next to the variable name: https://wiki.qt.io/Qt_Coding_Style

src/models/callercalleemodel.h
milianw: you don't need itemFromIndex either, as I said it is: `m_rows.value(index.row())` for a simple table

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,35 @@ struct SymbolData
...
+{
+    return location1.symbol == location2.symbol && location1.binary == location2.binary;
+}
+
+inline bool operator!=(const CallerCalleeLocation &location1, const CallerCalleeLocation &location2)
+{
+    return !(location1.symbol == location2.symbol && location1.binary == location2.binary);
+}
+
+inline uint qHash(const CallerCalleeLocation &key, uint seed)
```
milianw: default initialize `seed = 0`

src/parsers/perf/perfparser.cpp
```diff
@@ -503,6 +503,35 @@ struct SymbolData
...
+{
+    Util::HashCombineCommutative hash;
```
milianw: just use HashCombine, not the Commutative version

PR #27, opened 2017-02-23
Add system information to summary tab

nwrogers:


src/mainwindow.cpp
```diff
@@ -256,3 +268,19 @@ void MainWindow::aboutHotspot()
...
 }
+
+void MainWindow::on_systemInfoGroupBox_toggled(bool arg1)
+{
+    hideShowSystemInfo(arg1);
+}
+
+void MainWindow::hideShowSystemInfo(bool visible)
+{
+    foreach (QLabel *label, ui->systemInfoGroupBox->findChildren<QLabel*>()) {
```
milianw: Looking at the result, I think it would be better to always show the system information, but make the overall summary page scrollable. I.e. remove the toggle switch, always show the contents, but if the window is not tall enough it will not be visible and the user has to scroll to see it.

src/parsers/perf/perfparser.cpp
```diff
@@ -953,8 +953,17 @@ struct PerfParserPrivate
...
-        // TODO: add system info to summary page
+        summaryResult.hostName = QString::fromUtf8(features.hostName);
+        summaryResult.linuxKernelVersion = QString::fromUtf8(features.osRelease);
+        summaryResult.perfVersion = QString::fromUtf8(features.version);
+        summaryResult.cpuDescription = QString::fromUtf8(features.cpuDesc);
+        summaryResult.cpuId = QString::fromUtf8(features.cpuId);
+        summaryResult.cpuArchitecture = QString::fromUtf8(features.arch);
+        summaryResult.cpusOnline = features.nrCpusOnline;
+        summaryResult.cpusAvailable = features.nrCpusAvailable;
+        summaryResult.cpuSiblingCores = QString::fromUtf8(features.siblingCores.join(','));
```
milianw: please join with ", ", i.e. add a space after the comma

src/parsers/perf/perfparser.cpp
```diff
@@ -953,8 +953,17 @@ struct PerfParserPrivate
...
+        summaryResult.cpuSiblingThreads = QString::fromUtf8(features.siblingThreads.join(','));
```
milianw: dito

src/parsers/perf/perfparser.cpp
```diff
@@ -953,8 +953,17 @@ struct PerfParserPrivate
...
+        summaryResult.totalMemoryInKiB = features.totalMem;
```
milianw: please use KFormat's formatByteSize, see heaptrack or https://api.kde.org/frameworks/kcoreaddons/html/classKFormat.html#ae7412420b70e2ca935d0ebed6770e313

src/parsers/perf/perfparser.cpp
milianw: also, looking at the output: could you please sort the sibling threads list before joining it? Currently, I get:

0,4,1,5,2,6,3,7

src/mainwindow.ui
```diff
@@ -465,6 +471,175 @@
...
+              </property>
+              <property name="checkable">
+               <bool>true</bool>
+              </property>
+              <property name="checked">
+               <bool>false</bool>
+              </property>
+              <layout class="QFormLayout" name="formLayout_2">
+               <item row="0" column="0">
+                <widget class="QLabel" name="hostNameLabel">
```
milianw: I just realize that none of the labels on the summary page are selectable. Can you add that feature (afaik it's a property on the label), such that we can copy'n'paste from the summary page?

src/mainwindow.cpp
```diff
@@ -185,6 +187,17 @@ MainWindow::MainWindow(QWidget *parent) :
...
+                ui->linuxKernelVersionValue->setText(data.linuxKernelVersion);
+                ui->perfVersionValue->setText(data.perfVersion);
+                ui->cpuDescriptionValue->setText(data.cpuDescription);
+                ui->cpuIdValue->setText(data.cpuId);
+                ui->cpuArchitectureValue->setText(data.cpuArchitecture);
+                ui->cpusOnlineValue->setText(QString::number(data.cpusOnline));
+                ui->cpusAvailableValue->setText(QString::number(data.cpusAvailable));
+                ui->cpuSiblingCoresValue->setText(data.cpuSiblingCores);
+                ui->cpuSiblingThreadsValue->setText(data.cpuSiblingThreads);
+                ui->totalMemoryValue->setText(format.formatByteSize(data.totalMemoryInKiB, 1, KFormat::MetricBinaryDialect));
```
milianw: this is wrong, you need to pass in bytes, but totalMemoryInKiB, as the name indicates, is in KiB. So you'll have to multiply it by 1024.

src/parsers/perf/perfparser.cpp
```diff
@@ -811,8 +811,27 @@ struct PerfParserPrivate
...
+        summaryResult.cpuId = QString::fromUtf8(features.cpuId);
+        summaryResult.cpuArchitecture = QString::fromUtf8(features.arch);
+        summaryResult.cpusOnline = features.nrCpusOnline;
+        summaryResult.cpusAvailable = features.nrCpusAvailable;
+        summaryResult.cpuSiblingCores = QString::fromUtf8(features.siblingCores.join(", "));
+
+        QList<QByteArray> siblingThreads;
+        for (QByteArray thread : features.siblingThreads) {
+            QList<QByteArray> siblings = thread.split(',');
+            for (QByteArray sibling : siblings) {
```
milianw: FYI this could be simplified, but see above

```
siblingThreads += thread.split(',');
```

src/parsers/perf/perfparser.cpp
```diff
@@ -811,8 +811,27 @@ struct PerfParserPrivate
...
+        summaryResult.linuxKernelVersion = QString::fromUtf8(features.osRelease);
+        summaryResult.perfVersion = QString::fromUtf8(features.version);
+        summaryResult.cpuDescription = QString::fromUtf8(features.cpuDesc);
+        summaryResult.cpuId = QString::fromUtf8(features.cpuId);
+        summaryResult.cpuArchitecture = QString::fromUtf8(features.arch);
+        summaryResult.cpusOnline = features.nrCpusOnline;
+        summaryResult.cpusAvailable = features.nrCpusAvailable;
+        summaryResult.cpuSiblingCores = QString::fromUtf8(features.siblingCores.join(", "));
+
+        QList<QByteArray> siblingThreads;
```
milianw: Looking at the output from perf again, I think I misunderstood this and only realize it now that I see how you parse this here. So, sorry, but this needs to be revised again:

the sibling threads seems to indicate hardware threads, i.e. CPU ids, that share the same core. So for my Intel 8 core e.g. I have:

- siblingCores=(0-7)
- siblingThreads=(0,4, 1,5, 2,6, 3,7)

this means, I have one CPU node with 8 cores on them, but actually two of them always share the same hardware (hyperthreading). So I have only 4 "real" cores, and 8 "virtual" ones.

So splitting and sorting like I proposed is very wrong. Sorry again. Instead, I'd suggest we change the format to something like this:

Sibling Cores: [0-7]
Sibling Threads: [0,4], [1,5], [2,6], [3,7]

So for both of these lists, wrap every entry in '[' and ']', then join everything together with commas.

src/mainwindow.cpp
```diff
@@ -197,7 +197,7 @@ MainWindow::MainWindow(QWidget *parent) :
                 ui->cpusAvailableValue->setText(QString::number(data.cpusAvailable));
                 ui->cpuSiblingCoresValue->setText(data.cpuSiblingCores);
                 ui->cpuSiblingThreadsValue->setText(data.cpuSiblingThreads);
-                ui->totalMemoryValue->setText(format.formatByteSize(data.totalMemoryInKiB, 1, KFormat::MetricBinaryDialect));
+                ui->totalMemoryValue->setText(format.formatByteSize(data.totalMemoryInKiB, 1, KFormat::IECBinaryDialect));
```
milianw: this is still wrong, and actually worse than before ;-)

we want MetricBinaryDialect, but we need to convert totalMemoryInKib to bytes by adding a * 1024 here

src/parsers/perf/perfparser.cpp
```diff
@@ -819,18 +819,8 @@ struct PerfParserPrivate
...
-        for (QByteArray thread : features.siblingThreads) {
-            QList<QByteArray> siblings = thread.split(',');
-            for (QByteArray sibling : siblings) {
-                siblingThreads.append(sibling);
-            }
-        }
-        std::sort(siblingThreads.begin(), siblingThreads.end());
-        summaryResult.cpuSiblingThreads = QString::fromUtf8(siblingThreads.join(", "));
-
+        summaryResult.cpuSiblingCores = QString::fromUtf8(features.siblingCores.join(", ")).prepend(QLatin1Char('[')).replace(QLatin1Literal(", "), QLatin1Literal("], [")).append(QLatin1Char(']'));
```
milianw: use QLatin1String instead of QLatin1Literal please, the latter is Qt 4 stuff afaik

milianw:
Also, the commit message is still not good, see comment up top

src/parsers/perf/perfparser.cpp
```diff
@@ -811,8 +811,17 @@ struct PerfParserPrivate
...
-        // TODO: add system info to summary page
+        summaryResult.hostName = QString::fromUtf8(features.hostName);
+        summaryResult.linuxKernelVersion = QString::fromUtf8(features.osRelease);
+        summaryResult.perfVersion = QString::fromUtf8(features.version);
+        summaryResult.cpuDescription = QString::fromUtf8(features.cpuDesc);
+        summaryResult.cpuId = QString::fromUtf8(features.cpuId);
+        summaryResult.cpuArchitecture = QString::fromUtf8(features.arch);
+        summaryResult.cpusOnline = features.nrCpusOnline;
+        summaryResult.cpusAvailable = features.nrCpusAvailable;
+        summaryResult.cpuSiblingCores = QString::fromUtf8(features.siblingCores.join(", ")).prepend(QLatin1Char('[')).replace(QLatin1String(", "), QLatin1String("], [")).append(QLatin1Char(']'));
```
milianw: this and the following line needs to be reformatted to not be so long. I think something like this would be good, and it would also simplify the code a bit (note how you insert ", " and then replace it again later). Or am I missing something?

        auto formatCpuList = [](const QByteArrayList& list) -> QString {
            return QString::fromUtf8('[' + list.join("], [") + ']');
        };
        summaryResult.cpuSiblingCores = formatCpuList(features.siblingCores);
        summaryResult.cpuSiblingThreads = formatCpuList(features.siblingThreads);

nwrogers:
OK, done

Issue #23, opened 2017-01-31
use heaptrack's costdelegate in views


milianw:
the costdelegate code should be copied over to allow us to graphically visualize the percentage cost in our views

PR #34, opened 2017-02-28
Add code for cost delegate

nwrogers:
Bring in and adapt Heaptracks cost delegate for Hotspot

#23

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -43,6 +43,8 @@
 #include "parsers/perf/perfparser.h"
 
 #include "models/summarydata.h"
+#include "models/hashmodel.h"
+#include "models/costdelegate.h"
 #include "models/treemodel.h"
 #include "models/topproxy.h"
 #include "models/callercalleemodel.h"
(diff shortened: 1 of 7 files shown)


src/mainwindow.cpp
```diff
@@ -132,11 +133,15 @@ MainWindow::MainWindow(QWidget *parent) :
     ui->mainPageStack->setCurrentWidget(ui->startPage);
     ui->openFileButton->setFocus();
 
+    auto costDelegate = new CostDelegate(this);
+
     auto bottomUpCostModel = new BottomUpModel(this);
     setupTreeView(ui->bottomUpTreeView,  ui->bottomUpSearch, bottomUpCostModel);
+    ui->bottomUpTreeView->setItemDelegate(costDelegate);
```
milianw: uhm, don't we only want to set that for the cost columns? This sets it for all columns - is that correct? I don't think so.

src/models/costdelegate.cpp
```diff
@@ -0,0 +1,85 @@
...
+    : QStyledItemDelegate(parent)
+{
+}
+
+CostDelegate::~CostDelegate() = default;
+
+void CostDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, const QModelIndex& index) const
+{
+    // TODO: handle negative values
+    const int64_t cost = index.data(AbstractTreeModel::SortRole).toULongLong();
```
milianw: use auto, note how you get an unsigned value but store it in a signed int64_t

src/models/costdelegate.cpp
```diff
@@ -0,0 +1,85 @@
...
+    if (cost == 0) {
+        QStyledItemDelegate::paint(painter, option, index);
+        return;
+    }
+
+    const int64_t totalCost = index.data(AbstractTreeModel::TotalCostRole).toULongLong();
```
milianw: dito

src/models/costdelegate.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // top-down can miscalculate the peak cost
```
milianw: I think this might actually not be required anymore - could you check whether removing the std::min(1, ) wrapper? the abs and division is still fine though.

src/models/costdelegate.cpp
```diff
@@ -0,0 +1,85 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "costdelegate.h"
+
+#include "treemodel.h"
```
milianw: because this delegate also needs to support the hashmodel (for caller/callee), we must not let it depend on the treemodel

instead, pass in the roles to query via the ctor, store them in member variables, and then use that below.

this also means you'll need two delegates in the mainwindow: one for all tree models, and one for all hashmodels.

nwrogers:
Fixes made, comment updated, and I tested std:min. At least with the perf.data file I am testing with the results were identical for every result so I went ahead and removed the std::min wrapper

PR #38, opened 2017-03-07
Add text selection + copy support on summary page

### LINKED ISSUE #35: text selection + copy support on summary page
the text on the summary page needs to be selectable

Maybe we need to use HTML definition lists like I did in heaptrack within a QLabel that has selection enabled.  Currently we have individual labels, which is a mess to extend, I don't think it's worth it. Just have one label for the summary, and one for the system information. Then build a HTML `<dl>` with the contents and set it on these labels. Then we can copy'n'paste the contents into a text editor

nwrogers:
Update text on the summary screen to make it easier to select and copy the text.

Fixes #35

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -223,11 +223,44 @@ MainWindow::MainWindow(QWidget *parent) :
 
     connect(m_parser, &PerfParser::summaryDataAvailable,
             this, [this, bottomUpCostModel, topDownCostModel, callerCalleeCostModel, calleesModel, callersModel] (const SummaryData& data) {
-                KFormat format;
-                ui->appRunTimeValue->setText(formatTimeString(data.applicationRunningTime));
-                ui->threadCountValue->setText(QString::number(data.threadCount));
-                ui->processCountValue->setText(QString::number(data.processCount));
-                ui->sampleCountValue->setText(QString::number(data.sampleCount));
+                auto formatSummaryText = [] (const QString& description, const QString& value) -> QString {
+                    return QString(QLatin1String("<tr><td>") + description + QLatin1String(": </td><td>")
+                                   + value + QLatin1String("</td></tr>"));
+                };
+
+                QString summaryText;
+                {
+                    QTextStream stream(&summaryText);
+                    stream << "<qt><table>"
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -223,20 +223,51 @@ MainWindow::MainWindow(QWidget *parent) :
 
     connect(m_parser, &PerfParser::summaryDataAvailable,
             this, [this, bottomUpCostModel, topDownCostModel, callerCalleeCostModel, calleesModel, callersModel] (const SummaryData& data) {
-                KFormat format;
-                ui->appRunTimeValue->setText(formatTimeString(data.applicationRunningTime));
-                ui->threadCountValue->setText(QString::number(data.threadCount));
-                ui->processCountValue->setText(QString::number(data.processCount));
-                ui->sampleCountValue->setText(QString::number(data.sampleCount));
+                auto formatSummaryText = [] (QString description, QString value) -> QString {
```
milianw: const& the QStrings

src/mainwindow.ui
```diff
@@ -290,170 +290,12 @@
...
+                    <widget class="QLabel" name="summaryLabel">
                      <property name="toolTip">
                       <string>The number of threads found in the profile.</string>
                      </property>
                      <property name="text">
-                      <string/>
-                     </property>
-                     <property name="wordWrap">
-                      <bool>true</bool>
+                      <string>summary text</string>
```
milianw: please set the "translatable" property to false for this filler text

src/mainwindow.ui
```diff
@@ -495,285 +337,12 @@
...
-                    </widget>
-                   </item>
-                   <item row="10" column="1">
-                    <widget class="QLabel" name="totalMemoryValue">
-                     <property name="toolTip">
-                      <string>The total amount of RAM on the host machine.</string>
-                     </property>
-                     <property name="text">
-                      <string/>
+                      <string>system info</string>
```
milianw: dito

nwrogers:
The Qt::TextSelectableByMouse value is being set through the ui file...
<set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>

PR #39, opened 2017-03-14
Add context menu with option to jump to Caller/Callee

### LINKED ISSUE #33: jump to callers/callee view from stack views
The stack views, i.e. bottom-up, top-down and flame-graph should get context menus that show the selected symbol in the caller/callee view. I.e. set the current index, and ensure it's visible.

nwrogers:
Add context menu with option to jump to Caller/Callee tab from Bottom-Up, Top-Down, and Flame-Graph.

Fixes #33

### DIFF
--- a/src/flamegraph.cpp
+++ b/src/flamegraph.cpp
@@ -44,6 +44,7 @@
 #include <QCheckBox>
 #include <QDoubleSpinBox>
 #include <QCursor>
+#include <QMenu>
 
 #include <ThreadWeaver/ThreadWeaver>
 #include <KLocalizedString>
@@ -447,6 +448,19 @@ bool FlameGraph::eventFilter(QObject* object, QEvent* event)
         updateTooltip();
     } else if (event->type() == QEvent::Hide) {
         setData(nullptr);
+    } else if (event->type() == QEvent::ContextMenu) {
+        QContextMenuEvent *contextEvent = static_cast<QContextMenuEvent*>(event);
+        auto item = static_cast<FrameGraphicsItem*>(m_view->itemAt(m_view->mapFromGlobal(contextEvent->globalPos())));
+        if (!item) {
+            return ret;
+        }
+
+        QMenu contextMenu;
+        auto *viewCallerCallee = contextMenu.addAction(tr("View Caller/Callee"));
+        QAction *action = contextMenu.exec(QCursor::pos());
+        if (action == viewCallerCallee) {
+            emit jumpToCallerCallee(item->symbol());
+        }
     }
     return ret;
 }
--- a/src/flamegraph.h
+++ b/src/flamegraph.h
@@ -56,6 +56,9 @@ protected:
 private slots:
     void setData(FrameGraphicsItem* rootItem);
 
+signals:
+    void jumpToCallerCallee(const Data::Symbol& symbol);
+
 private:
     void setTooltipItem(const FrameGraphicsItem* item);
     void updateTooltip();
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -89,6 +89,8 @@ void setupTreeView(QTreeView* view, KFilterProxySearchLine* filter, Model* model
     view->sortByColumn(Model::InitialSortColumn);
     view->setModel(proxy);
     stretchFirstColumn(view);
+
+    view->setContextMenuPolicy(Qt::CustomContextMenu);
 }
 
 template<typename Model>
@@ -150,11 +152,13 @@ MainWindow::MainWindow(QWidget *parent) :
     auto bottomUpCostModel = new BottomUpModel(this);
     setupTreeView(ui->bottomUpTreeView,  ui->bottomUpSearch, bottomUpCostModel);
     ui->bottomUpTreeView->setItemDelegateForColumn(BottomUpModel::Cost, treeViewCostDelegate);
+    connect(ui->bottomUpTreeView, &QTreeView::customContextMenuRequested, this, &MainWindow::onBottomUpContextMenu);
 
     auto topDownCostModel = new TopDownModel(this);
     setupTreeView(ui->topDownTreeView, ui->topDownSearch, topDownCostModel);
     ui->topDownTreeView->setItemDelegateForColumn(TopDownModel::SelfCost, treeViewCostDelegate);
     ui->topDownTreeView->setItemDelegateForColumn(TopDownModel::InclusiveCost, treeViewCostDelegate);
+    connect(ui->topDownTreeView, &QTreeView::customContextMenuRequested, this, &MainWindow::onTopDownContextMenu);
 
     auto topHotspotsProxy = new TopProxy(this);
     topHotspotsProxy->setSourceModel(bottomUpCostModel);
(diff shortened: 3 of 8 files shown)


src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
     dialog.adjustSize();
     dialog.exec();
 }
+
+void MainWindow::customContextMenu(const QPoint &point, QTreeView* view, quint32 symbolRole)
```
milianw: use int instead of quint32

src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
...
+{
+    QModelIndex index = view->indexAt(point);
+    if (index.isValid()) {
```
milianw: invert and do early-return when it's not valid, decreases the indentation depth below

src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
...
+        QMenu contextMenu;
+        QAction *invokeAction = nullptr;
```
milianw: merge with next line (see flamegraph comment above)

milianw: also rename

src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
...
+{
+    customContextMenu(point, ui->bottomUpTreeView, BottomUpModel::SymbolRole);
+}
+
+void MainWindow::onTopDownContextMenu(const QPoint &point)
+{
+    customContextMenu(point, ui->topDownTreeView, TopDownModel::SymbolRole);
+}
+
+void MainWindow::onJumpToCallerCallee(const Data::Symbol& symbol)
```
milianw: remove the on prefix

milianw: move & next to arg name instead of type name

src/mainwindow.h
```diff
@@ -30,6 +30,11 @@
 #include <QMainWindow>
 #include <QScopedPointer>
 #include <QString>
+#include <QSortFilterProxyModel>
+#include <QTreeView>
+
+#include "callercalleemodel.h"
```
milianw: foward-declare the model and the symbol, that should work - no need to include the fat headers here

src/mainwindow.h
```diff
@@ -30,6 +30,11 @@
 #include <QMainWindow>
 #include <QScopedPointer>
 #include <QString>
+#include <QSortFilterProxyModel>
```
milianw: forward-declare both

src/models/hashmodel.h
```diff
@@ -104,6 +106,16 @@ class HashModel : public QAbstractTableModel
         return index(row, column);
     }
 
+    QModelIndex indexForSymbol(Data::Symbol symbol)
```
milianw: const&

milianw: this method should be moved into the SymbolHashModel, in `callercalleemodel.h`

src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
...
+{
+    QModelIndex index = view->indexAt(point);
+    if (index.isValid()) {
+        QMenu contextMenu;
+        QAction *invokeAction = nullptr;
+        invokeAction = contextMenu.addAction(tr("View Caller/Callee"));
+
+        QAction *action = contextMenu.exec(QCursor::pos());
+        if (action == invokeAction) {
+            const auto symbol = index.data(symbolRole).template value<Data::Symbol>();
```
milianw: the `template ` part shouldn't be required here, just `value` should be enough

src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
...
+
+void MainWindow::customContextMenu(const QPoint &point, QTreeView* view, quint32 symbolRole)
+{
+    QModelIndex index = view->indexAt(point);
+    if (index.isValid()) {
+        QMenu contextMenu;
+        QAction *invokeAction = nullptr;
+        invokeAction = contextMenu.addAction(tr("View Caller/Callee"));
+
+        QAction *action = contextMenu.exec(QCursor::pos());
```
milianw: auto *

src/mainwindow.cpp
```diff
@@ -367,3 +375,36 @@ void MainWindow::aboutHotspot()
     dialog.adjustSize();
     dialog.exec();
 }
+
+void MainWindow::customContextMenu(const QPoint &point, QTreeView* view, quint32 symbolRole)
+{
+    QModelIndex index = view->indexAt(point);
```
milianw: const auto

src/flamegraph.h
```diff
@@ -80,6 +83,7 @@ private slots:
     bool m_collapseRecursion = false;
     // cost threshold in percent, items below that value will not be shown
     double m_costThreshold = 0.1;
+    const FrameGraphicsItem* m_contextMenuItem = nullptr;
```
milianw: remove once you cleaned up the event handler

src/models/hashmodel.h
nwrogers: m_rows is private, is there another way to get the rows from outside the HashModel without having to make m_rows protected? Should I make m_rows protected? Also what is SymbolHashModel?

src/mainwindow.cpp
```diff
@@ -214,10 +218,12 @@ MainWindow::MainWindow(QWidget *parent) :
             });
 
     auto calleesModel = setupCallerOrCalleeView<CalleeModel>(ui->calleesView, ui->callerCalleeTableView,
-                                                             callerCalleeCostModel, callerCalleeProxy);
+                                                             m_callerCalleeCostModel, m_callerCalleeProxy);
+    auto calleeCostDelegate = new CostDelegate(CalleeModel::SortRole, CalleeModel::TotalCostRole, this);
```
nwrogers: Messed up the merge :/

src/flamegraph.cpp
```diff
@@ -447,6 +448,24 @@ bool FlameGraph::eventFilter(QObject* object, QEvent* event)
         updateTooltip();
     } else if (event->type() == QEvent::Hide) {
         setData(nullptr);
+    } else if (event->type() == QEvent::MouseButtonPress) {
```
nwrogers: I originally tried to do it all in event->type() == QEvent::ContextMenu but When I tried to get the FrameGraphicsItem I always got NULL. I checked the position returned from mapFromGlobal and it was off by one pixel in x and one pixel in y, so I made the one pixel adjustment and voila! I get our item... not sure what is worse the old way or the new way...

QPoint itemPosition(QCursor::pos().x()-1, QCursor::pos().y()-1);
auto item = static_cast<FrameGraphicsItem*>(m_view->itemAt(m_view->mapFromGlobal(itemPosition)));

milianw: odd, could you try http://doc.qt.io/qt-5/qcontextmenuevent.html#globalPos - does that work?

src/models/hashmodel.h
milianw: sorry, I mean `SymbolCostModelImpl`, and yes make m_rows protected for that purpose

nwrogers: Did you mean I should move it to CallerCalleeModel? I don't call indexForSymbol on the CostModel, at least not at this time

src/flamegraph.cpp
nwrogers: That did the trick

src/flamegraph.cpp
```diff
@@ -447,6 +448,16 @@ bool FlameGraph::eventFilter(QObject* object, QEvent* event)
         updateTooltip();
     } else if (event->type() == QEvent::Hide) {
         setData(nullptr);
+    } else if (event->type() == QEvent::ContextMenu) {
+        QMenu contextMenu;
+        auto *viewCallerCallee = contextMenu.addAction(tr("View Caller/Callee"));
+
+        QContextMenuEvent *contextEvent = static_cast<QContextMenuEvent*>(event);
+        auto item = static_cast<FrameGraphicsItem*>(m_view->itemAt(m_view->mapFromGlobal(contextEvent->globalPos())));
```
milianw: actually, when no item is found, we shouldn't even offer the context menu, right? so please change this to check for an item first before creating the menu and action

src/mainwindow.cpp
```diff
@@ -367,3 +373,36 @@ void MainWindow::aboutHotspot()
     dialog.adjustSize();
     dialog.exec();
 }
+
+void MainWindow::customContextMenu(const QPoint &point, QTreeView* view, int symbolRole)
+{
+    const auto index = view->indexAt(point);
+    if (!index.isValid()) {
```
milianw: this is the correct way of doing it, the flamegraph should do the same (see comment above)

milianw:
oh, you'll need to resolve the conflicts and resubmit - I added some new API in mainwindow.h

PR #40, opened 2017-03-16
Add an Open in Editor feature

### LINKED ISSUE #10: open editor feature
Every entry in the hotspot gui that resembles a code location should offer a "open in editor" context menu. This should then open the preferred editor of the user. Use `QDesktopServices::openUrl` by default, but allow the user to configure an editor/ide of his choice with potentially custom command line arguments via hotspot settings. GammaRay has this feature as well.

nwrogers:
Add an Open in Editor feature to allow the user to setup a default editor and open a source file from the Location section of the Caller/Callee tab in that editor.

This commit introduces the Open in Editor from GammaRay and adapts it
for usage in hotspot.

Fixes #10

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -32,6 +32,10 @@
 #include <QFileDialog>
 #include <QSortFilterProxyModel>
 #include <QApplication>
+#include <QSettings>
+#include <QStandardPaths>
+#include <QProcess>
+#include <QInputDialog>
 
 #include <KRecursiveFilterProxyModel>
 #include <KStandardAction>
@@ -124,6 +128,31 @@ Model* setupCallerOrCalleeView(QTreeView* view, QTreeView* callersCalleeView,
                     });
     return model;
 }
+
+struct IdeSettings {
+    const char * const app;
+    const char * const args;
+    const char * const name;
+    const char * const icon;
+};
+
+static const IdeSettings ideSettings[] = {
+#if defined(Q_OS_WIN) || defined(Q_OS_OSX)
+    {"", "", "", ""} // Dummy content, because we can't have empty arrays.
+#else
+    { "kdevelop", "%f:%l:%c", QT_TRANSLATE_NOOP("MainWindow", "KDevelop"), "kdevelop" },
+    { "kate", "%f --line %l --column %c", QT_TRANSLATE_NOOP("MainWindow", "Kate"), "kate" },
+    { "kwrite", "%f --line %l --column %c", QT_TRANSLATE_NOOP("MainWindow", "KWrite"), nullptr },
+    { "gedit", "%f +%l:%c", QT_TRANSLATE_NOOP("MainWindow", "gedit"), nullptr },
+    { "gvim", "%f +%l", QT_TRANSLATE_NOOP("MainWindow", "gvim"), nullptr },
+    { "qtcreator", "%f", QT_TRANSLATE_NOOP("MainWindow", "Qt Creator"), nullptr }
+#endif
+};
+#if defined(Q_OS_WIN) || defined(Q_OS_OSX) // Remove this #if branch when adding real data to ideSettings for Windows/OSX.
+    static const int ideSettingsSize = 0;
+#else
+    static const int ideSettingsSize = sizeof(ideSettings) / sizeof(IdeSettings);
+#endif
 }
 
 MainWindow::MainWindow(QWidget *parent) :
@@ -224,6 +253,8 @@ MainWindow::MainWindow(QWidget *parent) :
                                                              m_callerCalleeCostModel, m_callerCalleeProxy);
 
     auto sourceMapModel = setupModelAndProxyForView<SourceMapModel>(ui->sourceMapView);
+    ui->sourceMapView->setContextMenuPolicy(Qt::CustomContextMenu);
+    connect(ui->sourceMapView, &QTreeView::customContextMenuRequested, this, &MainWindow::onSourceMapContextMenu);
 
     connect(ui->callerCalleeTableView->selectionModel(), &QItemSelectionModel::currentRowChanged,
             this, [calleesModel, callersModel, sourceMapModel] (const QModelIndex& current, const QModelIndex& /*previous*/) {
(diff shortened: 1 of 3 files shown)


src/mainwindow.cpp
```diff
@@ -293,6 +325,64 @@ MainWindow::MainWindow(QWidget *parent) :
...
+    menu->addAction(action);
+#endif
+
+    QObject::connect(group, SIGNAL(triggered(QAction*)), this, SLOT(setCodeNavigationIDE(
+                                                                         QAction *)));
+
+    configAction->setMenu(menu);
+    ui->settingsMenu->addMenu(menu);
+
+    connect(this, SIGNAL(navigateToCode(QUrl,int,int)), this,
```
milianw: use new-style connect

src/mainwindow.cpp
```diff
@@ -293,6 +325,64 @@ MainWindow::MainWindow(QWidget *parent) :
...
+    action = new QAction(menu);
+    action->setText(tr("Automatic (No Line numbers)"));
+    action->setCheckable(true);
+    action->setChecked(currentIdx == -2);
+    action->setData(-2);
+    group->addAction(action);
+    menu->addAction(action);
+#endif
+
+    QObject::connect(group, SIGNAL(triggered(QAction*)), this, SLOT(setCodeNavigationIDE(
```
milianw: use new-style connect

src/mainwindow.cpp
```diff
@@ -293,6 +325,64 @@ MainWindow::MainWindow(QWidget *parent) :
         ui->resultsTabWidget->setTabToolTip(i, ui->resultsTabWidget->widget(i)->toolTip());
     }
 
+    QSettings settings(QStringLiteral("KDAB"), QStringLiteral("Hotspot"));
```
milianw: please extract this stuff into a separate function, to make it clear that this code is only about code navigation / settings

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
...
+        QDesktopServices::openUrl(QUrl(url));
+    }
+
+    const QString filePath = url.isLocalFile() ? url.toLocalFile() : url.toString();
+    command.replace(QStringLiteral("%f"), filePath);
+    command.replace(QStringLiteral("%l"), QString::number(std::max(1, lineNumber)));
+    command.replace(QStringLiteral("%c"), QString::number(std::max(1, columnNumber)));
+
+    if (!command.isEmpty()) {
+        qDebug() << "Detaching: " << command;
```
milianw: remove

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
...
+    if (ideIdx >= 0 && ideIdx < ideSettingsSize) {
+        command += QString::fromUtf8(ideSettings[ideIdx].app);
+        command += QString::fromUtf8(" ");
+        command += QString::fromUtf8(ideSettings[ideIdx].args);
+    } else
+#endif
+    if (ideIdx == -1) {
+        command = settings.value(QStringLiteral("CustomCommand")).toString();
+    } else {
+        QDesktopServices::openUrl(QUrl(url));
```
milianw: this should return early, as we open the file here. no need to open it below via qprocess a second time?

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
...
+    if (!index.isValid()) {
+        return;
+    }
+    const auto sourceMap = index.data(SourceMapModel::LocationRole).value<QString>();
+    auto seperator = sourceMap.lastIndexOf(QLatin1Char(':'));
+    if (seperator <= 0) {
+        return;
+    }
+
+    QString fileName = m_sysroot + m_appPath + sourceMap.left(seperator);
```
milianw: only prepend the appPath _or_ the sysroot, not both. i.e. extract this code that finds the url into a helper and call it twice, once with the sysroot and once with the app path prefix

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
...
+
+void MainWindow::onSourceMapContextMenu(const QPoint &point)
+{
+    const auto index = ui->sourceMapView->indexAt(point);
+    if (!index.isValid()) {
+        return;
+    }
+    const auto sourceMap = index.data(SourceMapModel::LocationRole).value<QString>();
+    auto seperator = sourceMap.lastIndexOf(QLatin1Char(':'));
+    if (seperator <= 0) {
```
milianw: I wonder if we could ever get a situation where we know a filename, but no line number? in such a case, opening the file would still be useful. but probably it's OK to keep this as-is and if we ever see such a case we can improve its handling later on

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
...
+    }
+
+    QString fileName = m_sysroot + m_appPath + sourceMap.left(seperator);
+    if (QFileInfo::exists(fileName)) {
+        QMenu contextMenu;
+        auto *viewCallerCallee = contextMenu.addAction(tr("Open in editor"));
+        auto *action = contextMenu.exec(QCursor::pos());
+        if (action == viewCallerCallee) {
+            int lineNumber = sourceMap.mid(seperator+1).toInt();
+            navigateToCode(QUrl(fileName), lineNumber, 0);
```
milianw: please pass the filename as QString here, see comments above

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
     dialog.adjustSize();
     dialog.exec();
 }
+
+void MainWindow::navigateToCode(const QUrl &url, int lineNumber, int columnNumber)
```
milianw: I don't see us supporting remote urls, so please change the url to a `const QString &path`

src/mainwindow.cpp
milianw: once you use a QString instead of passing the url, use QUrl::fromLocalFile here

src/mainwindow.cpp
```diff
@@ -398,3 +488,80 @@ void MainWindow::aboutHotspot()
...
+        command += QString::fromUtf8(ideSettings[ideIdx].args);
+    } else
+#endif
+    if (ideIdx == -1) {
+        command = settings.value(QStringLiteral("CustomCommand")).toString();
+    } else {
+        QDesktopServices::openUrl(QUrl(url));
+    }
+
+    const QString filePath = url.isLocalFile() ? url.toLocalFile() : url.toString();
```
milianw: we will always deal with local files (see above)

PR #55, opened 2017-06-08
Perf: Reserve memory for QSet

krf:
Summary:
32 seems to be a good candidate here, deduced from taking the highest
number from a 500 MB perf.data snapshot.


krf:
Just noticed this patch still lying around in my checkout, please check.

milianw:
this is not actually reserving memory for qset, it's mostly preventing rehashing - but if you measure the performance of this code path that's rarely if ever an issue. the much bigger issue, performance wise, is allocating the nodes for the hash contents - and those won't get allocated by a call to reserve. 

if you fixup your commit message to clarify what this patch is actually doing then I'm OK with it. but I bet you won't see a tangible difference in speed (did you check before/after?)

milianw:
ping?

krf:
Abandoning; not sure what I was trying to speed up here anymore.

Issue #11, opened 2016-12-23
launch perf from hotspot


milianw:
Allow running perf from within a nice launcher GUI in hotspot

This should allow multiple user-configurable analysis types, to select different event sources for perf, or do off-cpu VS. on-cpu profiling - or both at the same time.

As a first iteration, this only needs to run on localhost.

tmcguire:
For a novice, troubleshooting information is important - if something went wrong when launching perf, how to fix that? I believe perf itself already gives nice output in many cases - it would be useful to simply show that output to the hotspot user. Example:

> WARNING: Kernel address maps (/proc/{kallsyms,modules}) are restricted,
> check /proc/sys/kernel/kptr_restrict.
> 
> Samples in kernel functions may not be resolved if a suitable vmlinux
> file is not found in the buildid cache or in the vmlinux path.
> 
> Samples in kernel modules won't be resolved at all.
> 
> If some relocation was applied (e.g. kexec) symbols may be misresolved
> even with a suitable vmlinux or kallsyms file.
> 
> Perf session creation failed.

tmcguire:
In addition, warning the user when the perf version is to old and certain features won't be supported would be a nice touch.

milianw:
we need to support three cases:

- lauch new process
  - use file selector (e.g. KUrlRequester) to select an existing binary (executable file)
  - free-form line edit to add command line arguments for the binary
  - table of custom environment variables
- attach to process(es)
  - use KSysGuard to show active processes
  - then take the PID of the selected process(es) and attach to them via `--pid PID1,PID2` in the `perf record` call
- system wide
  - "simply" run `perf record -a` as root
    Note that systemwide will potentially require sudo rights, so we need to figure out how to do that correctly... Maybe use polkit (via the KDE libraries) for that? Needs to be researched.

for all the above, the user also needs to select the unwinding method, i.e. a drop down in this order:

- DWARF
- none
- fp
- lbr (only on intel hardware that actually has this feature, not sure how to detect that, needs to be researched)

additionally, the event type should be selectable. This should default to nothing, which internally uses the smart perf heuristics (i.e. cycles:ppp if available). Otherwise, if possible, show the categorized events form `perf list` in a smart way. Note that multiple events can be selected, and modifiers like `:k` or `:u` etc. can be applied. But for starters a free-form line edit could be used for that I guess (see below)

[...]

PR #57, opened 2017-06-29
Record perf data by launching an application

nwrogers:
Version 1.0 of the perf record GUI functionality.

Issue: #11

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -47,11 +47,16 @@
 #include <KFormat>
 #include <KConfigGroup>
 #include <KRecentFilesAction>
+#include <KUrlRequester>
+#include <Solid/Device>
+#include <Solid/Processor>
+#include <KShell>
 
 #include "aboutdialog.h"
 #include "flamegraph.h"
 
 #include "parsers/perf/perfparser.h"
+#include "perfrecord.h"
 
 #include "models/summarydata.h"
 #include "models/hashmodel.h"
@@ -165,19 +170,48 @@ static const IdeSettings ideSettings[] = {
 #else
     static const int ideSettingsSize = sizeof(ideSettings) / sizeof(IdeSettings);
 #endif
+
+bool isIntel()
+{
+    const auto list = Solid::Device::listFromType(Solid::DeviceInterface::Processor, QString());
+    Solid::Device device = list[0];
+    if(list.empty() || !device.is<Solid::Processor>()) {
+        return false;
+    }
+    const auto *processor= device.as<Solid::Processor>();
+    const auto instructionSets = processor->instructionSets();
+
+    return instructionSets.testFlag(Solid::Processor::IntelMmx) ||
+           instructionSets.testFlag(Solid::Processor::IntelSse) ||
+           instructionSets.testFlag(Solid::Processor::IntelSse2) ||
+           instructionSets.testFlag(Solid::Processor::IntelSse3) ||
+           instructionSets.testFlag(Solid::Processor::IntelSsse3) ||
+           instructionSets.testFlag(Solid::Processor::IntelSse4) ||
+           instructionSets.testFlag(Solid::Processor::IntelSse41) ||
+           instructionSets.testFlag(Solid::Processor::IntelSse42);
+}
 }
 
 MainWindow::MainWindow(QWidget *parent) :
     QMainWindow(parent),
     ui(new Ui::MainWindow),
     m_parser(new PerfParser(this)),
-    m_config(KSharedConfig::openConfig())
+    m_config(KSharedConfig::openConfig()),
+    m_perfRecord(new PerfRecord(this))
 {
     ui->setupUi(this);
 
     ui->lostMessage->setVisible(false);
     ui->parserErrorsBox->setVisible(false);
-    ui->fileMenu->addAction(KStandardAction::open(this, SLOT(on_openFileButton_clicked()), this));
+
+    auto *recordDataAction = new QAction(this);
+    recordDataAction->setText(QStringLiteral("&Record Data"));
+    recordDataAction->setIcon(QIcon::fromTheme(QStringLiteral("media-record")));
+    recordDataAction->setShortcut(Qt::CTRL + Qt::Key_R);
+    ui->fileMenu->addAction(recordDataAction);
+    connect(recordDataAction, &QAction::triggered, this, &MainWindow::onRecordDataButtonClicked);
+
+    ui->fileMenu->addAction(KStandardAction::open(this, SLOT(onOpenFileButtonClicked()), this));
     m_recentFilesAction = KStandardAction::openRecent(this, SLOT(openFile(QUrl)), this);
     m_recentFilesAction->loadEntries(m_config->group("RecentFiles"));
     ui->fileMenu->addAction(m_recentFilesAction);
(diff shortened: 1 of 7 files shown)


src/mainwindow.cpp
```diff
@@ -164,19 +169,47 @@ static const IdeSettings ideSettings[] = {
...
+bool isIntel()
+{
+    QList<Solid::Device> list = Solid::Device::listFromType(Solid::DeviceInterface::Processor, QString());
+    Solid::Device device = list[0];
+    if(!device.is<Solid::Processor>()) {
+        return false;
+    }
+    Solid::Processor *processor = device.as<Solid::Processor>();
+
+    return processor->instructionSets().testFlag(Solid::Processor::IntelMmx) ||
```
milianw: introduce a temporary to cleanup the code below:

```
const auto instructionSets = processor->instructionSets();
return instructionSets.testFlag(...) || ...
```

src/mainwindow.cpp
```diff
@@ -191,6 +224,21 @@ MainWindow::MainWindow(QWidget *parent) :
...
+    ui->workingDirectory->setMode(KFile::Directory);
+    ui->applicationRecordErrorLabel->hide();
+    connect(ui->openFileButton, &QPushButton::clicked, this, &MainWindow::onOpenFileButtonClicked);
+    connect(ui->recordDataButton, &QPushButton::clicked, this, &MainWindow::onRecordDataButtonClicked);
+    connect(ui->homeButton, &QPushButton::clicked, this, &MainWindow::onHomeButtonClicked);
+    connect(ui->applicationName, &KUrlRequester::textChanged, this, &MainWindow::onApplicationNameChanged);
+    connect(ui->startRecordingButton, &QPushButton::toggled, this, &MainWindow::onStartRecordingButtonClicked);
+
+    ui->recordTypeComboBox->addItem(tr("Launch Application"), QVariant::fromValue(LaunchApplication));
+    ui->callGraphComboBox->addItem(tr("None"), QVariant::fromValue(QStringLiteral("")));
```
milianw: `QVariant::fromValue(QString())`

src/mainwindow.cpp
```diff
@@ -413,6 +476,48 @@ void MainWindow::on_openFileButton_clicked()
...
+
+void MainWindow::onStartRecordingButtonClicked(bool checked)
+{
+    if (checked) {
+        ui->startRecordingButton->setIcon(QIcon::fromTheme(QStringLiteral("media-playback-stop")));
+        ui->startRecordingButton->setText(tr("Stop Recording"));
+
+        QStringList perfOptions;
+
+        if (!ui->callGraphComboBox->currentData().toString().isEmpty()) {
```
milianw: again, temporary to prevent repetition in code:

```
const auto callGraphOption = ui->callGraphComboBox->currentData().toString();
if (!callGraphOption.isEmpty()) {
    perfOptions << QStringLiteral("--call-graph") << callGraphOption;
}
```

milianw:
I have tested it now, some comments on the GUI:

- I dislike the "Home" button, but don't have a better idea on what to use there, so leave it as-is ;-)
- there's no live validation - i.e. inserting "foobar" into the application does not show me an error that no such executable can be found
- the validation on record is wrong, i.e. it does not check for binaries in the `PATH`. Use `QStandardDirs::findExecutable` for that purpose
- use `QDir::currentPath()` instead of `.` (when no application path is given, see above) for the default working directory
- the labels are not buddies of the edit widgets
- tooltips are missing
- the "start recording" button should have text beside icon, not below

I now tried to record this:

```
binary: /usr/bin/ls
params: /tmp/
working dir: <empty, i.e. /usr/bin>
Unwinding Method: None
Event Type: cycles,instructions
```

And got: "Failed to record perf data, error code 255" and nothing else. At least, show the error output. There wasn't even anything displayed in the command line. I suspect the issue is that the data file is being written to the working directory (`/usr/bin`) which is not writable. So this needs validation and a knob to configure where the file is getting written to, with a sensible default. Indeed, changing the working directory makes it work, but this is not helpful. Note that perf shows this on the command line:

```
/usr/bin$ perf record /usr/bin/ls /tmp -e cycles,instructions
failed to open perf.data : Permission denied
Perf session creation failed.
```

This is important to show to the user.

src/mainwindow.cpp
```diff
@@ -418,6 +493,85 @@ void MainWindow::on_openFileButton_clicked()
...
+    }
+    ui->applicationRecordErrorLabel->show();
+}
+
+void MainWindow::onWorkingDirectoryNameChanged(const QString& folderPath)
+{
+    QFileInfo folder(folderPath);
+
+    if (!folder.exists()) {
+        ui->applicationRecordErrorLabel->setText(tr("Working directory folder cannot be found: ") + folderPath);
```
mgerhardy: This might not work in all languages. It's better to use `tr("Foo %1").arg(folderPath)`

src/mainwindow.cpp
```diff
@@ -165,19 +170,48 @@ static const IdeSettings ideSettings[] = {
 #else
     static const int ideSettingsSize = sizeof(ideSettings) / sizeof(IdeSettings);
 #endif
+
+bool isIntel()
+{
+    QList<Solid::Device> list = Solid::Device::listFromType(Solid::DeviceInterface::Processor, QString());
+    Solid::Device device = list[0];
```
milianw: just to be pedantic - can you make sure we never crash here if the list returned is empty (shouldn't happen, but still). i.e. return false when the list is empty

src/mainwindow.cpp
```diff
@@ -165,19 +170,48 @@ static const IdeSettings ideSettings[] = {
 #else
     static const int ideSettingsSize = sizeof(ideSettings) / sizeof(IdeSettings);
 #endif
+
+bool isIntel()
+{
+    QList<Solid::Device> list = Solid::Device::listFromType(Solid::DeviceInterface::Processor, QString());
```
milianw: `const auto list = ...`

src/mainwindow.cpp
```diff
@@ -165,19 +170,48 @@ static const IdeSettings ideSettings[] = {
...
+    Solid::Device device = list[0];
+    if(!device.is<Solid::Processor>()) {
+        return false;
+    }
+    Solid::Processor *processor = device.as<Solid::Processor>();
```
milianw: `const auto *processor = ...`

src/mainwindow.cpp
```diff
@@ -165,19 +170,48 @@ static const IdeSettings ideSettings[] = {
...
+    m_config(KSharedConfig::openConfig()),
+    m_perfRecord(new PerfRecord(this))
 {
     ui->setupUi(this);
 
     ui->lostMessage->setVisible(false);
     ui->parserErrorsBox->setVisible(false);
-    ui->fileMenu->addAction(KStandardAction::open(this, SLOT(on_openFileButton_clicked()), this));
+
+    QAction* recordDataAction = new QAction(this);
```
milianw: `auto *recordDataAction = ...`

src/mainwindow.cpp
```diff
@@ -192,6 +226,25 @@ MainWindow::MainWindow(QWidget *parent) :
 
     ui->mainPageStack->setCurrentWidget(ui->startPage);
     ui->openFileButton->setFocus();
+    ui->applicationName->setText(QDir::currentPath());
```
milianw: the application name set to the current path? why?

src/mainwindow.h
```diff
@@ -46,6 +46,24 @@ class CallerCalleeModel;
...
+
+enum RecordType
+{
+    LaunchApplication,
+    AttachToProcess,
+    ProfileSystem
+};
+Q_DECLARE_METATYPE(RecordType)
+
+enum BacktraceType
```
milianw: this is unused, no? remove it please

src/perfrecord.cpp
```diff
@@ -43,38 +46,70 @@ PerfRecord::~PerfRecord() = default;
...
                         const QString &exePath, const QStringList &exeOptions)
 {
-    QFileInfo info(exePath);
-    if (!info.exists()) {
+    // Reset perf record process to avoid getting signals from old processes
+    m_perfRecordProcess->kill();
+    m_perfRecordProcess->deleteLater();
+    m_perfRecordProcess = new QProcess(this);
+
+    QFileInfo fileInfo(exePath);
```
milianw: name it `exeFileInfo`

milianw:
some functional comments:

- use a `KMessageWidget` for displaying the error(s) in the record page, and put the error on top or on bottom, not somewhere in the middle like right now (actually, this is OK at the bottom, but see below why I thought this is the case)
- the labels should be buddy widgets of the accompanying line edits
- the label should read `Event Type(s)`
- DWARF unwinding should be selected by default
- we need more tooltips, also for the line edits / combo boxes and individual items, but I can add some if you are at a loss for words
- I think it's odd to specify an output *directory*, it should be the output *file*, which should default to `<working directory>/perf.data`, then pass the file to `perf record` via it's `-o` parameter
- you can keep the existing "output directory" though but rename it to "working directory", i.e. the dir where the app will be launched
- perf output should only be shown when there is actual output, i.e. after starting the record (on a separate page, see also above)
- the perf output text edit should not be editable (currently it is)
- the perf output should use the monospace font (i.e. system default monospace)
- the perf output should color error output red (KColorScheme negative foreground color)
- when the application quits with an error, we still might have recorded some data which we could show the user (I'd prefer if we ask for that in such a case). to reproduce, run the `cpp-recursion` test client without passing any command line argument. it will return with an error, but you'll still have recorded some samples for that
- I notice that the command can become quite long, and our summary page will then scroll horizontally instead of splitting the command to multiple lines - probably a bug that should be fixed separately

nwrogers:
What will we use the "working directory" for? If we ask the user for the application name and an output file, then why do we need a working directory? (asking honestly, will I copy the application to the working directory and run it from there?)

Also, not sure how to use KColorScheme. Right now Perf Output is a QTextEdit, what do I have to promote it to to use KColorScheme? I looked at KTextEdit but I don't see that it uses KColorScheme.

milianw:
Some application may need to be run in a certain folder, this is what the working directory is there for. Imagine an `ls` without the ability to specify a folder as a CLI argument. If you want to `ls` in `/tmp`, you'd first need to do `cd /tmp;` before running it. There are some (you may say broken) applications like that out in the wild.

nwrogers:
Makes sense about the working directory. All requested changes have been made. Not sure why the CI failed, I don't think it was anything I did

milianw:
thanks, merged it now!

PR #59, opened 2017-07-06
Update README.md

htower:
On my system (kde neon) CMake fails without this commands.


milianw:
oh indeed, forgot that - thanks!

PR #69, opened 2017-07-10
Fix display of inclusive costs as tooltips in the top-down view

jktjkt:
Feel free to squash/edit/amend at will, or just to take the first commit if you preferred that.


jktjkt:
No problem, enjoy your holiday. You're right, I pushed a fixed version of the second change.

milianw:
thanks!

PR #75, opened 2017-07-24
TravisCI support

takacsd:
I created a simple configuration file for Travis CI if you are interested.

Sadly the latest Ubuntu environment is 16.04 on Travis CI and I couldn't make the build work on it, so I had to use docker to have the latest Ubuntu. The downside is that I can't install the necessary perf package there, so the tst_perfparser test is not working.

This is unfortunate, but I think the automatic build is still helpful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+sudo: required
+
+services:
+  - docker
+
+install:
+  - docker build -t hotspot .
+
+script:
+  - docker run -v $TRAVIS_BUILD_DIR:/hotspot hotspot /bin/sh -c "cd hotspot && mkdir build-hotspot && cd build-hotspot && cmake .. && make"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM ubuntu:17.04
+
+RUN apt-get update && apt-get install -y \
+    build-essential \
+    gettext \
+    libkf5threadweaver-dev \
+    libkf5i18n-dev \
+    libkf5configwidgets-dev \
+    libkf5coreaddons-dev \
+    libkf5itemviews-dev \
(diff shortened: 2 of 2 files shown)


milianw:
in principle I have no issue with doing this, but at KDAB we host a (non-public) CI that does post-commit compiles and unit test runs.

Before merging this, can you tell me more about the issue with perf? You'll need the packages linux-tools-common and linux-tools-generic, then perf should be available.

probonopd:
Please see https://github.com/KDAB/hotspot/pull/76, I think I managed to compile ot on Travis CI using trusty.

takacsd:
@milianw: AFAIK Perf needs some kernel specific package to work, something like linux-tools-`uname -r`-common or similar. Docker is not a full VM (under Linux), so even the latest ubuntu image will use trusty's kernel on travis, but there is no kernel specific linux-tools package for that old kernel in the repository of zesty.
(If the host is also zesty, it works fine if you starts a privileged docker container.)

My end goal would be to generate packages or something like AppImage for every release, and this PR is just the first step, but if you have this at KDAB already, and you can share it, then it' s even better. ;)

@probonopd: I really like your initiative on AppImage, but I kind of dislike the ppa and the prebuilt stuff in your script. Wouldn't be possible and somewhat cleaner to use docker there too?

probonopd:
@takacsd in which way do you think would using Docker be cleaner? Another option could be using the Open Build Service https://git.io/obs-ai.

takacsd:
@probonopd: Since every dependency is already in the zesty docker image, I thought it is much easier to bundle them, but I did not know that you need to build the AppImage on an old distribution to make it work on lots of distributions.

milianw:
I'm closing this now, as we have merged #76 now. I'm in the process of fixing the travis build.

PR #76, opened 2017-07-24
Generate AppImage

probonopd:
This PR, when merged, will compile this application on [Travis CI](https://travis-ci.org/) upon each `git push`, and upload an [AppImage](http://appimage.org/) to a temporary download URL on transfer.sh (available for 14 days). The download URL is toward the end of each Travis CI build log of each build (see below for how to set up automatic uploading to your GitHub Releases page).

For this to work, you need to enable Travis CI for your repository as [described here](https://travis-ci.org/getting_started) __prior to merging this__, if you haven't already done so.

Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions

[Here is an overview](https://github.com/probonopd/AppImageKit/wiki/AppImages) of projects that are already distributing upstream-provided, official AppImages.

__Please note:__ Instead of storing AppImage builds temporarily for 14 days each on transfer.sh, you could use GitHub Releases to store the binaries permanently. This way, they would be visible on the Releases page of your project. This is what I recommend. See https://docs.travis-ci.com/user/deployment/releases/. If you want to do this for continuous builds, also see https://github.com/probonopd/uploadtool.

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+language: cpp
+compiler: gcc
+sudo: require
+dist: trusty
+
+before_install:
+  - sudo add-apt-repository ppa:beineri/opt-qt59-trusty -y
+  - sudo apt-get update -qq
+    
+install: 
+  - sudo apt-get -y install qt59base qt59x11extras elfutils libelf-dev libdw-dev libasm-dev
+  - source /opt/qt*/bin/qt*-env.sh
+  - git clone git://anongit.kde.org/extra-cmake-modules
+  - cd extra-cmake-modules
+  - mkdir build
+  - cd build
+  - cmake .. # or run : cmake -DCMAKE_INSTALL_PREFIX=/usr .. &&
+  - make
+  - sudo make install
+  - cd ../..
+  - # Precompiled KF5
+  - wget -c "https://github.com/chigraph/precompiled-kf5-linux/releases/download/precompiled/kf5-gcc6-linux64-release.tar.xz"
+  - tar xf kf5-gcc6-linux64-release.tar.xz
+  - sudo cp -Rf root/kf5-release/* /opt/qt*/
+
+script:
+  - mkdir build
+  - cd build
+  - cmake .. -DCMAKE_INSTALL_PREFIX=/usr
+  - make -j$(nproc)
+  - make DESTDIR=appdir install ; find appdir/
+  - mkdir -p appdir/usr/share/applications/ # FIXME: Do in CMakeLists.txt
+  - cp ../hotspot.desktop appdir/usr/share/applications/ # FIXME: Do in CMakeLists.txt
+  - wget -c "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage" 
+  - chmod a+x linuxdeployqt*.AppImage
+  - unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
+  - ./linuxdeployqt*.AppImage ./appdir/usr/share/applications/*.desktop -bundle-non-qt-libs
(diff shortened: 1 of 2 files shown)


probonopd:
@milianw 
> at KDAB we host a (non-public) CI that does post-commit compiles and unit test runs

https://github.com/probonopd/hotspot/blob/feature/appimage/.travis.yml should be fairly straigtforward to carry over from Travis CI to whatever system you are using, if you prefer that.

milianw:
Thanks, but I have some questions:

What's the advantage over https://github.com/KDAB/hotspot/pull/75 ?

Why is it OK to use trusty for generating the AppImage? I was so far under the impression one should use an old distro like Centos 6 or so to build the AppImage. Will the one build on trusty run on older distros? I doubt it personally, but would like to be proven wrong.

probonopd:
Well, using CentOS 6 would make the AppImage run on even older distributions, but trusty is from 2014, so most distributions that came out in the last 3 years should be covered. The advantage is that you do not need to maintain a Docker image, builds are faster, and an AppImage is generated.

milianw:
let's try this out :)

milianw:
Ah, the older gcc doesn't default to C++11 yet, I'll push a fix to enable that explicitly.

milianw:
Odd, we do set

```
set(CMAKE_CXX_STANDARD 14)
```

already, is the cmake/gcc too old for that?

probonopd:
Strange, I think I had it working 10 days ago. Has something changed recently?

probonopd:
Yep, it used to build, as can be seen here:
https://travis-ci.org/probonopd/hotspot/builds/256950787

probonopd:
Do you have any ideas why it might be failing now?

urzds:
Are the download URLs for this advertised anywhere?  I.e. outside of the build job logs?

probonopd:
Currently they are only visible in the build logs.

If that is too cumbersome, check out https://github.com/probonopd/uploadtool for how to upload continuous builds to GitHub Releases. Advantage: Easier to find, constant URL.

milianw:
Hey guys,

sorry for the delay, but it was vacation time and now I'm busy catching up at work. I'll look into this probably at the beginning of next week.

Cheers

milianw:
I have finally fixed it:

https://travis-ci.org/KDAB/hotspot

I'll now have a look at the generated AppImage. Thanks again @probonopd

milianw:
Seems to work :)

probonopd:
Looking forward to the next release. If you would like to publish continuous builds, have a look at https://github.com/probonopd/uploadtool

milianw:
Yes, thanks for the heads up @probonopd, I'll have a look at this eventually. I'll also have to see how to leverage this whole setup to create AppImages for specific git tags - do you have any resources on that matter? I.e. once 1.1.0 is released, how can I get a AppImage build for it - or will Travis build *all* branches/tags, not just master?

probonopd:
By default, Travis will build all branches/tags, but it can be configured,

PR #77, opened 2017-07-25
Add Arch Linux dependency installation command

nilsdeppe:
PS Thanks for the awesome package!


milianw:
I'm an arch user, but I don't have `qt5ct` installed, I think you rather want to list the deps explicitly - something like `qt5-base qt5-tools`

nilsdeppe:
Do you have the deps at a finer scale? I don't, unfortunately and hopefully understandably don't want to uninstall things just to figure them out. If you have the list I'm happy to edit.

milianw:
It should be just `qt5-base`, amend that and we can take it from there

nilsdeppe:
Awesome, thanks! Just pushed the update.

milianw:
thanks

PR #82, opened 2017-08-03
Default to C++11 & require it under CMake

krf:
Related to #76


milianw:
cool, let's try this

@probonopd  any idea what the `rake` stuff is about on Travis (see above?).

probonopd:
"Could not find .travis.yml, using standard configuration." It cannot work without the [`.travis.yml`](https://github.com/KDAB/hotspot/blob/master/.travis.yml).

PR #98, opened 2017-10-05
Updated compile for Ubuntu (Linux Mint)

martinus:
The default description did not work for me. With these changes I could get it to compile and run.


milianw:
thanks!

PR #101, opened 2017-10-25
Simplify version numbering export

NewProggie:
If a project version is specified in the CMake project() command,
there's no need for manually adding the version information as it is
already provided by CMake.


krf:
Thanks!

milianw:
nice, I wasn't aware of this - thank you

PR #122, opened 2018-02-07
Add include directories for libelf.h

tmcguire:
This makes sure a custom libelf.h is picked instead of the system
libelf.h when available.

### DIFF
--- a/3rdparty/perfparser.cmake
+++ b/3rdparty/perfparser.cmake
@@ -1,5 +1,5 @@
 include_directories(
-    ${LIBELF_INCLUDE_DIR}
+    ${LIBELF_INCLUDE_DIRS}
     ${LIBDW_INCLUDE_DIR}/elfutils
     ${LIBDWARF_INCLUDE_DIRS}
     perfparser/app


3rdparty/perfparser.cmake
```diff
@@ -1,5 +1,6 @@
 include_directories(
-    ${LIBELF_INCLUDE_DIR}
+    ${LIBELF_INCLUDE_DIRS}
+    ${DWARF_INCLUDE_DIR}
```
milianw: this should be `LIBDWARF_INCLUDE_DIRS`

milianw:
Now that I'm on my laptop, I'll push the changes I have here. That includes one which will conflict with this changeset - namely the one which uses `LIBDWARF_INCLUDE_DIRS`. Please amend your changeset to fix the `LIBELF_INCLUDE_DIR` trailing `s` then.

3rdparty/perfparser.cmake
tmcguire: FindElfUtils.cmake uses DWARF_INCLUDE_DIR:

```find_path (DWARF_INCLUDE_DIR
    NAMES
      dwarf.h
    PATHS
      /usr/include
      /usr/local/include
      /opt/local/include
      /sw/include
      ENV CPATH) # PATH and INCLUDE will also work
```

tmcguire: Nevermind, later there is a `set (LIBDWARF_INCLUDE_DIRS  ${DWARF_INCLUDE_DIR} ${LIBDW_INCLUDE_DIR})`

milianw:
thanks!

Issue #131, opened 2018-04-16
Recording does not start (when perf executable is missing)


pager72:
Added screencast, zipped video file.
[issue.zip](https://github.com/KDAB/hotspot/files/1913726/issue.zip)

Nothing happens after I hit record, and it wont even stop after I hit it again. No logfile is made in specified directory also.

pager72:
Ran as root issue persists. Refer to screencast.
[bla.zip](https://github.com/KDAB/hotspot/files/1917422/bla.zip)

krf:
@milianw This is likely caused by a missing `perf` executable. I can reproduce this when the `perf` binary is missing.

milianw:
@pager72 in your screencast you pasted the shell command into the "event type" command line which is utterly wrong. Please paste the command I have shown *into a console*. Outside of hotspot. Can you confirm Kevins idea? Do you even have perf installed?

pager72:
Trying to build perf, but no instructions in
https://github.com/KDAB/perfparser

pager72:
Installed perf, hotspot works.

pager72:
Its out of the scope for this bug.

PR #132, opened 2018-04-17
Disable recording if perf is not installed

nwrogers:
Fixes: #131

### DIFF
--- a/src/recordpage.cpp
+++ b/src/recordpage.cpp
@@ -40,6 +40,7 @@
 #include <QtConcurrent/QtConcurrentRun>
 #include <QKeyEvent>
 #include <QShortcut>
+#include <QFileInfo>
 
 #include <KUrlRequester>
 #include <Solid/Device>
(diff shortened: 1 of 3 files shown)


src/recordpage.cpp
```diff
@@ -88,8 +89,26 @@ RecordType selectedRecordType(const QScopedPointer<Ui::RecordPage> &ui)
     return ui->recordTypeComboBox->currentData().value<RecordType>();
 }
 
+bool isPerfInstalled()
+{
+    QByteArray envPath = qgetenv("PATH");
```
krf: Should probably just use http://doc.qt.io/qt-5/qstandardpaths.html#findExecutable

=> `const bool isPerfInstalled = !QStandardPaths::findExecutable("perf").isEmpty()`

krf:
There should probably be a central function for getting the actual path to the perf executable? Dunno Hotspot code enough, though.

milianw:
+1, but please resolve the conflicts (I've added a clang-format file now)

milianw:
merged now, thanks @nwrogers

PR #139, opened 2018-07-11
Updated README.md

PatilChetan:
Building Hotspot had few simple steps missing and cmake path should be previous directory not ../hotspot. I think this will help avoid confusion to someone who is new to Linux.


krf:
I don't think that makes anything easier, does it?

Having source & build directories side-by-side (which is what the README suggests) is very common.

PatilChetan:
I thought `cmake ../hotspot` is incorrect command if previous step is asking to do `cd build-hotspot`. There is not `hotspot` folder in the parent directory. 

`cmake ../../hotspot` would be correct too.

PR #150, opened 2018-08-20
Update readme with gentoo installation information

DaveeFTW:
Do we need more than this?


milianw:
I think that's enough.

milianw:
thanks!

PR #152, opened 2018-09-11
Update dependencies for Arch Linux

alendit:
The package dependencies in Arch have changed, so we have to specify additional packages explicitly.


krf:
This is wrong, Hotspolt indeed only looks for:

```
    find_package(KF5 COMPONENTS
        ThreadWeaver
        I18n
        ConfigWidgets
        CoreAddons
        ItemViews
        ItemModels
        KIO 
        Solid
        WindowSystem
        REQUIRED
    )
```

Everything else from KF5 are transitive dependencies.

alendit:
Interesting, the original packages are enough on a clean docker image. Must have been an issue with my installation. 

Sorry for the noise.

krf:
Thanks for the confirmation!

PR #172, opened 2019-02-21
Set up path settings menu after CLI args have been applied

### LINKED ISSUE #171: Command line arguments not reflected in GUI
Values set with e.g. `--sysroot` and `--debugPaths` are not shown in the open file settings in the GUI.

Also, last time I tried with `--sysroot` it didn't work with a relative path. I had to provide the full path to get it to work. Haven't tested this with the latest version so I might have been fixed.

erijo:
Otherwise the values from the command line aren't shown in the GUI, even though
they take effect.

Fixes #171.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -117,6 +117,9 @@ MainWindow::MainWindow(QWidget* parent)
     layout->addWidget(m_pageStack);
     centralWidget()->setLayout(layout);
 
+    connect(this, &MainWindow::sysrootChanged, m_resultsPage, &ResultsPage::setSysroot);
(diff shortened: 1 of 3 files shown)


milianw:
I'd prefer we add signals to the setters and use those to update the GUI edits, can you update your patch accordingly?

erijo:
Won't that give a strange circular behavior where the line edit is edited, it signals textChanged, which calls e.g. setSysroot, which emits a signal that's connected to the line edit's setText slot, which triggers textChanged, and so on? Or I'm I missing something?

milianw:
in that case the text won't change, and thus no signal will be emitted, so that's safe

erijo:
Was it something like this that you had in mind?

milianw:
lgtm, thanks!

PR #180, opened 2019-04-17
Update hotspot build instructions

apelisse:


milianw:
your patch isn't correct, I think you misunderstood the current documentation which I believe is correct:

1) `git clone ...` gives you a folder named `hotspot`, but we don't `cd` into it
2) we create the build folder as a sibling to `hotspot`, then `cd` into it
3) then we need to run cmake on the source folder, which is `../cmake`

so -> this patch can be abandoned, no?

apelisse:
Oh I see, it's because I cd'd into hotspot, then mkdir the build-hotspot inside the hotspot dir, which is why it didn't work for me. Sorry for the noise :-)

PR #183, opened 2019-05-14
Add custom environments setting

pangqingyuan1991:


src/perfrecord.cpp
```diff
@@ -172,6 +172,7 @@ void PerfRecord::startRecording(const QStringList& perfOptions, const QString& o
     }
     m_perfRecordProcess = new QProcess(this);
     m_perfRecordProcess->setProcessChannelMode(QProcess::MergedChannels);
+    m_perfRecordProcess->setEnvironment(QProcess::systemEnvironment() + env);
```
milianw: please use QProcessEnvironment, cf. https://doc.qt.io/qt-5/qprocess.html#systemEnvironment

src/recordpage.ui
```diff
@@ -117,13 +117,36 @@
...
+        <property name="text">
+         <string>Custom Environments:</string>
+        </property>
+        <property name="buddy">
+         <cstring>applicationParametersBox</cstring>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="1">
+       <widget class="QLineEdit" name="applicationEnvBox">
```
milianw: please add a place holder text with some example, e.g. `FOO=1 BAR=2`

src/perfrecord.cpp
pangqingyuan1991: QProcessEnvironment is a bit complex. And I got compiling error when I use QString to parse env.
‘QString::QString(const char*)’ is private within this context

milianw: you get the compile error when you try to parse a raw string (e.g. utf8) into a QString (utf16) - you are probably missing a QString::fromUtf8 or similar - can you post your code?

I also don't understand what's so complex about QPE, it's basically still a list of env vars?

PR #184, opened 2019-05-19
Add process environment setting support

pangqingyuan1991:
Hi,
I have recreate a new one.

### DIFF
--- a/src/perfrecord.cpp
+++ b/src/perfrecord.cpp
@@ -76,7 +76,7 @@ static QStringList sudoOptions(const QString& sudoBinary)
 }
 
 void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfOptions, const QString& outputPath,
-                                const QStringList& recordOptions, const QString& workingDirectory)
+                                const QStringList& recordOptions, const QStringList& env, const QString& workingDirectory)
 {
     if (elevatePrivileges) {
         // elevate privileges temporarily as root
@@ -128,7 +128,7 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
                 });
         // poll the file for new input, readyRead isn't being emitted by QFile (cf. docs)
         auto* readTimer = new QTimer(outputFile);
-        auto readSlot = [this, outputFile, perfOptions, outputPath, recordOptions, workingDirectory]() {
+        auto readSlot = [this, outputFile, perfOptions, outputPath, recordOptions, env, workingDirectory]() {
             const auto data = outputFile->readAll();
             if (data.isEmpty()) {
                 return;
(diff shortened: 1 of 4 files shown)


pangqingyuan1991:
I notice CFLAGS has -DQT_NO_CAST_FROM_ASCII, but I cannot find which file define this flag. This is the reason I can't use QString("abcd") directly

src/perfrecord.cpp
```diff
@@ -172,6 +172,13 @@ void PerfRecord::startRecording(const QStringList& perfOptions, const QString& o
     }
     m_perfRecordProcess = new QProcess(this);
     m_perfRecordProcess->setProcessChannelMode(QProcess::MergedChannels);
+    auto e = QProcessEnvironment::systemEnvironment();
+    foreach(const QString &str, env){
```
milianw: for (const auto &str : env) {

src/perfrecord.cpp
```diff
@@ -172,6 +172,13 @@ void PerfRecord::startRecording(const QStringList& perfOptions, const QString& o
...
+        QStringList var = str.split(QChar::fromLatin1('='));
```
milianw: const auto var = str.split(QLatin1Char('='));

milianw: hm no, that's actually not enough, when the (string) value contains another = then we don't want to split again, so this should find the first `=` via indexOf and then pass the first and last part to `e.insert`

src/perfrecord.cpp
```diff
@@ -172,6 +172,13 @@ void PerfRecord::startRecording(const QStringList& perfOptions, const QString& o
     }
     m_perfRecordProcess = new QProcess(this);
     m_perfRecordProcess->setProcessChannelMode(QProcess::MergedChannels);
+    auto e = QProcessEnvironment::systemEnvironment();
```
milianw: please give it a more descriptive name, like `processEnvironment`

src/perfrecord.cpp
```diff
@@ -172,6 +172,13 @@ void PerfRecord::startRecording(const QStringList& perfOptions, const QString& o
...
+    foreach(const QString &str, env){
+        QStringList var = str.split(QChar::fromLatin1('='));
+        if(var.length()!=2) continue;
```
milianw: style:
```
if (var.size() != 2)
    continue;
```

src/recordpage.ui
```diff
@@ -117,13 +117,39 @@
...
+      <item row="2" column="1">
+       <widget class="QLineEdit" name="applicationEnvBox">
+        <property name="contextMenuPolicy">
+         <enum>Qt::NoContextMenu</enum>
+        </property>
+        <property name="toolTip">
+         <string>Optional environments to be set for the application being recorded</string>
+        </property>
+        <property name="placeholderText">
+         <string>FOO=1 BOO=2</string>
```
milianw: please use `FOO=1 BAR="asdf"` :)

src/recordpage.ui
```diff
@@ -117,13 +117,39 @@
...
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="1">
+       <widget class="QLineEdit" name="applicationEnvBox">
+        <property name="contextMenuPolicy">
+         <enum>Qt::NoContextMenu</enum>
+        </property>
+        <property name="toolTip">
+         <string>Optional environments to be set for the application being recorded</string>
```
milianw: Optional environment *variables*

pangqingyuan1991:
> pretty close, thanks for re-instantiating this patch! It's valuable and I'd like to get it in. can you fix the nits? otherwise I'll do that for you.
> 
> And yes, I'm setting `-DQT_NO_CAST_FROM_ASCII` deliberately to prevent the performance pitfalls of accidentally allocating utf16 QString data from utf8 string literals.

Hi, I think this is just a little feature and no much to do. So you can change it yourself.

milianw:
ok, I'll see when I get around to do that.

cheers

PR #199, opened 2019-08-15
Improve the XDG desktop file

Calinou:
This adds a generic name and comment, in English and in French.
This also adds keywords for easier searching.


milianw:
thanks!

PR #200, opened 2019-08-16
Add an AppStream metadata file

### LINKED ISSUE #191: AppStream information not available
https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

It would make it easier to find the application on software stores.
You can find plenty of examples in KDE and elsewhere. ;)

Calinou:
This closes #191.

PS: I added the email of the most active maintainer as the `update_contact`, let me know if this should be changed: slightly_smiling_face:

### DIFF
--- a/com.kdab.Hotspot.appdata.xml
+++ b/com.kdab.Hotspot.appdata.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright (C) 2013-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com -->
+<component type="desktop">
+  <id>com.kdab.Hotspot</id>
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>GPL-2.0-or-later</project_license>
+  <name>Hotspot</name>
+  <summary>Linux perf GUI for performance analysis</summary>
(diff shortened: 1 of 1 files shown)


com.kdab.Hotspot.appdata.xml
```diff
@@ -0,0 +1,45 @@
...
+  <url type="homepage">https://github.com/KDAB/hotspot</url>
+  <url type="bugtracker">https://github.com/KDAB/hotspot/issues</url>
+
+  <releases>
+    <release version="1.2.0" date="2019-05-20"/>
+  </releases>
+
+  <content_rating type="oars-1.1" />
+  <developer_name>KDAB</developer_name>
+  <update_contact>milian.wolff_AT_kdab.com</update_contact>
```
milianw: is the _AT_ standard procedure? my mail is public, so you could use @ directly

milianw:
one question, otherwise lgtm - thanks!

com.kdab.Hotspot.appdata.xml
Calinou: The [AppStream metadata guidelines](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html) state that using `_AT_` is valid, but I'll replace it with `@` if that's OK with you :slightly_smiling_face:

milianw:
thanks!

PR #206, opened 2019-09-23
HTML escape symbol name in tooltip

erijo:
As the symbol can include e.g. < and > which isn't shown unless escaped.


milianw:
thanks!

PR #208, opened 2019-09-24
Add support for prettifying symbols in cpp-parallel

erijo:
This is done by hiding parts (i.e. replacing with ...) of the internal templates used by primarily libstdc++.

It was requested in #195 that the symbols in cpp-parallel be prettified as well. This is one suggestion that hides big parts of the internal templates which doesn't add much useful information.

Let me know what you think.

### DIFF
--- a/tests/modeltests/tst_models.cpp
+++ b/tests/modeltests/tst_models.cpp
@@ -470,6 +470,24 @@ private slots:
                " char const (&) [1], char const (&) [1],"
                " std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&"
                "> >, bool ()>::operator()()";
+        QTest::newRow("std::call_once")
+            << "void std::call_once<...>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<"
+               "std::unique_ptr<std::__future_base::_Result_base, ...> ()>*, bool*"
+               "), std::__future_base::_State_baseV2*&&, std::function<"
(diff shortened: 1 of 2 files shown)


tests/modeltests/tst_models.cpp
```diff
@@ -470,6 +470,24 @@ private slots:
                " char const (&) [1], char const (&) [1],"
                " std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&"
                "> >, bool ()>::operator()()";
+        QTest::newRow("std::call_once")
+            << "void std::call_once<...>(std::once_flag&, void (std::__future_base::_State_baseV2::*&&)(std::function<"
```
milianw: that doesn't really look much simpler imo :(

but then again, if we hide too much, we may obfuscate the unreadable backtrace even more, as it becomes impossible to google the "real" stuff?

what do you think? should we get this in? should we leave it as-is? do even more drastic simplifications?

erijo: I agree with you that this change doesn't actually add much value. As I said in #195 that I would do this I thought I should at least give it a shot. But I'm perfectly fine with dropping this PR.

milianw: It's a pity but I also think we should drop it :(

in the future we may want to rethink this eventually to get something like Visual Studio's "just my code" backtrace simplification, such that all of that internal cruft gets hidden optionally.

erijo: Agree, that sounds like a better way forward.

PR #209, opened 2019-09-25
fix typo in build instructions

Canop:


milianw:
this is not a typo, you have probably entered the checked-out folder which the commands listed there don't

we create a *sibling* build folder and to reference the source folder from there, we need `../hotspot` as written.

PR #213, opened 2019-09-28
Reset filter and zoom stack when reloading file

erijo:
Otherwise the timeline may end up in a state where it looks like it's empty and
requires a manual reset.


milianw:
thanks, I saw that myself recently and was quite puzzled by this!

PR #215, opened 2019-09-29
Include vertical scrollbar in root width calculation

erijo:
If width isn't reserved for the vertical scrollbar, then the root width will be
too wide if the scroll bar is needed after the graph has been painted, forcing
the horizontal scroll bar to appear.


milianw:
thanks!

PR #218, opened 2019-10-25
Fix some Qt deprecated calls

albert-astals-cid-kdab:


milianw:
forgot that I can merge directly :D done that now, thanks again!

PR #219, opened 2019-11-07
Remove deprecated calls to QTreeView::sortByColumn

albert-astals-cid-kdab:
On the two cases that it's setup calls, use Qt::DescendingOrder

In the one that is called from a "has data" signal, then reuse
the views header sortIndicatorOrder so to not override what the user may
had chosen


milianw:
thanks!

PR #231, opened 2020-03-16
Add ticks to the timeline

koenpoppe:
To improve the sense of scale used in the time line, I propose to add time ticks to the table header.

The pull request contains a two-level approach that should give both an idea about the point in time and the range of the ticks.

### DIFF
--- a/tests/test-clients/cpp-threadnames/main.cpp
+++ b/tests/test-clients/cpp-threadnames/main.cpp
@@ -1,7 +1,7 @@
 /*
   This file is part of Hotspot, the Qt GUI for performance analysis.
 
-  Copyright (C) 2017-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Copyright (C) 2017-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
   Author: Milian Wolff <milian.wolff@kdab.com>
 
   Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/LICENSE.GPL.txt
+++ b/LICENSE.GPL.txt
@@ -1,4 +1,4 @@
- The Hotspot software is Copyright (C) 2016-2019 Klaralvdalens Datakonsult AB.
+ The Hotspot software is Copyright (C) 2016-2020 Klaralvdalens Datakonsult AB.
 
  You may use, distribute and copy the Hotspot software under the terms of
  the GNU General Public License version 2 or under the terms of GNU General
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -5,7 +5,7 @@ Version 1.0
 Copyright of this license text (C) 2001 Trolltech AS and (C) 2002-2016
 Klarälvdalens Datakonsult AB.  All rights reserved. License text used
 with kind permission of Trolltech AS. The software and accompanying
-material is Copyright (C) 2010-2019 Klarälvdalens Datakonsult AB.
+material is Copyright (C) 2010-2020 Klarälvdalens Datakonsult AB.
 
 This non-exclusive non-transferable License Agreement ("Agreement") is
 between you ("Licensee") and Klarälvdalens Datakonsult AB (KDAB), and
--- a/com.kdab.Hotspot.appdata.xml
+++ b/com.kdab.Hotspot.appdata.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Copyright (C) 2013-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com -->
+<!-- Copyright (C) 2013-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com -->
 <component type="desktop">
   <id>com.kdab.Hotspot</id>
   <metadata_license>CC0-1.0</metadata_license>
--- a/hotspot-config.h.cmake
+++ b/hotspot-config.h.cmake
@@ -3,7 +3,7 @@
 
   This file is part of Hotspot, the Qt GUI for performance analysis.
 
-  Copyright (C) 2016-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Copyright (C) 2016-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
   Author: Milian Wolff <milian.wolff@kdab.com>
 
   Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/scripts/check_kallsyms.py
+++ b/scripts/check_kallsyms.py
@@ -4,7 +4,7 @@
 #
 # This file is part of Hotspot, the Qt GUI for performance analysis.
 #
-# Copyright (C) 2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+# Copyright (C) 2019-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
 # Author: Milian Wolff <milian.wolff@kdab.com>
 #
 # Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/scripts/check_mmaps.py
+++ b/scripts/check_mmaps.py
@@ -4,7 +4,7 @@
 #
 # This file is part of Hotspot, the Qt GUI for performance analysis.
 #
-# Copyright (C) 2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+# Copyright (C) 2019-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
 # Author: Milian Wolff <milian.wolff@kdab.com>
 #
 # Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/scripts/create_tarballs.sh
+++ b/scripts/create_tarballs.sh
@@ -2,7 +2,7 @@
 #
 # This file is part of Hotspot, the Qt GUI for performance analysis.
 #
-# Copyright (C) 2017-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+# Copyright (C) 2017-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
 # Author: Milian Wolff <milian.wolff@kdab.com>
 #
 # Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/scripts/elevate_perf_privileges.sh
+++ b/scripts/elevate_perf_privileges.sh
@@ -7,7 +7,7 @@
 #
 # This file is part of Hotspot, the Qt GUI for performance analysis.
 # 
-# Copyright (C) 2018-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+# Copyright (C) 2018-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
 # Author: Milian Wolff <milian.wolff@kdab.com>
 # 
 # Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/src/aboutdialog.cpp
+++ b/src/aboutdialog.cpp
@@ -3,7 +3,7 @@
 
   This file is part of Hotspot, the Qt GUI for performance analysis.
 
-  Copyright (C) 2013-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Copyright (C) 2013-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
   Author: Volker Krause <volker.krause@kdab.com>
 
   Licensees holding valid commercial KDAB Hotspot licenses may use this file in
--- a/src/aboutdialog.h
+++ b/src/aboutdialog.h
@@ -3,7 +3,7 @@
 
   This file is part of Hotspot, the Qt GUI for performance analysis.
 
-  Copyright (C) 2013-2019 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Copyright (C) 2013-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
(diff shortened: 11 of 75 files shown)


src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+    connect(filterAndZoomStack, &FilterAndZoomStack::zoomChanged, this, &TimeAxisHeaderView::onZoomChanged);
+}
+
+void TimeAxisHeaderView::setTimeRange(const Data::TimeRange &timeRange)
+{
+    m_timeRange = timeRange;
+    emit headerDataChanged(orientation(), 1, 1);
+}
+
+void TimeAxisHeaderView::onFilterChanged(const Data::FilterAction &filterAction)
```
milianw: this could be merged with the onZoomChanged handler, and both could be moved into a lambda. I.e. in the ctor write:

```
auto emitHeaderDataChanged = [this]() {
    emit headerDataChanged(orientation(), 1, 1);
};
connect(filterAndZoomStack, &FilterAndZoomStack::filterChanged, this, emitHeaderDataChanged);
connect(filterAndZoomStack, &FilterAndZoomStack::zoomChanged, this, emitHeaderDataChanged);
```

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+    emit headerDataChanged(orientation(), 1, 1);
+}
+
+void TimeAxisHeaderView::paintSection(QPainter *painter, const QRect &rect, int logicalIndex) const
+{
+    if (painter == nullptr) return;
+    if (logicalIndex != 1) return; // TODO: role based?
+
+    auto zoomTime = m_filterAndZoomStack->zoom().time;
+    if (!zoomTime.isValid())
```
milianw: please format the change with `git-clang-format`

milianw: and remove the braces from the single line conditionals

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+    const double resolution = (end - start) / rect.width();
+    const auto xForTime = [rect, start, resolution](const double time) {
+        return rect.x() + qRound((time - start)/resolution);
+    };
+
+    const int fontSize = painter->fontMetrics().height();
+    const int endLabelWidth = painter->fontMetrics().horizontalAdvance(QLatin1String("-xXXXm"));
+    const int targetNbTicks = rect.width() / endLabelWidth;
+
+    qDebug() << __FUNCTION__ << start << end << targetNbTicks;
```
milianw: remove please. also, FYI, https://woboq.com/blog/nice-debug-output-with-qt.html - no need to add `__FUNCTION__`, just set `QT_MESSAGE_PATTERN` and include `%{function}` in its format string

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+    const PrefixTickLabels pfl(start, end, targetNbTicks);
+
+    painter->setPen(QColor(0,0,0));
+    if (pfl.hasPrefix())
+    {
+        const QString placeholder = QLatin1String("xxx");
```
milianw: `const auto placeholder = QStringLiteral("xxx");`

your version allocates a temporary `QString` for the conversion from `QLatin1String`

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+    };
+
+    const int fontSize = painter->fontMetrics().height();
+    const int endLabelWidth = painter->fontMetrics().horizontalAdvance(QLatin1String("-xXXXm"));
+    const int targetNbTicks = rect.width() / endLabelWidth;
+
+    qDebug() << __FUNCTION__ << start << end << targetNbTicks;
+    const PrefixTickLabels pfl(start, end, targetNbTicks);
+
+    painter->setPen(QColor(0,0,0));
```
milianw: please don't hardcode colors, refer to the `palette()` instead please

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+void TimeAxisHeaderView::onZoomChanged(const Data::ZoomAction &zoomAction)
+{
+    Q_UNUSED(zoomAction)
+    emit headerDataChanged(orientation(), 1, 1);
+}
+
+void TimeAxisHeaderView::paintSection(QPainter *painter, const QRect &rect, int logicalIndex) const
+{
+    if (painter == nullptr) return;
+    if (logicalIndex != 1) return; // TODO: role based?
```
milianw: I don't quite understand that comment, what has the role to do with the index?

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+        bounding.setWidth(prefixWidth);
+        painter->drawText(bounding, Qt::AlignBottom | Qt::AlignLeft, placeholder);
+    }
+
+    for (const auto &tickAndLabel : pfl.ticksAndLabel())
+    {
+        const auto x = xForTime(tickAndLabel.first);
+        if (std::abs(tickAndLabel.first - pfl.prefixValue()) < oneNanoSecond)
+        {
+            painter->setPen(QColor(0,0,0));
```
milianw: dito regarding hardcoded color

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+            painter->drawLine(x, rect.y() + fontSize, x, rect.y() + rect.height());
+        }
+        else
+        {
+            painter->setPen(QColor(0,0,255));
```
milianw: dito

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    const double niceMin = ceil(min / spacing) * spacing;
+    const int nbTicks = static_cast<int>(floor((max - niceMin) / spacing));
+
+    const auto siPrefix = PrefixTickLabels::siPrefix(scale_10);
+    m_ticksAndLabels.reserve(nbTicks);
+    for (int j = 0; j <= nbTicks; j++)
+    {
+        const double tick = niceMin + j * spacing;
+        const auto tickLabel = QString(QLatin1String("%1%2")).arg((tick - m_prefixValue) * inv_scale, 0, 'f', labelFraction_10).arg(tick != 0.0 ? siPrefix : QString());
```
milianw: what is this doing, formatting a fractional number? can't you use a floating point number instead?

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+const PrefixTickLabels::TicksAndLabel &PrefixTickLabels::ticksAndLabel() const
+{
+    return m_ticksAndLabels;
+}
+
+
+// Utility
+
+double PrefixTickLabels::niceNum(double value)
```
milianw: can you please add a comment, what is this doing?

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,257 @@
...
+    }
+    else if (fraction <= 5.0)
+    {
+        niceFraction = 5;
+    }
+
+    return niceFraction * pow(10, value_10);
+}
+
+QString PrefixTickLabels::siPrefix(int power)
```
milianw: this should go to `util.h/cpp`

src/models/timeaxisheaderview.h
```diff
@@ -0,0 +1,86 @@
...
+private:
+    int m_prefix_10 = 0;
+    double m_prefixValue = 0.0;
+    QString m_prefixLabel;
+
+
+    // Ticks
+public:
+    using TicksAndLabel = QVector<QPair<double, QString>>;
+    const TicksAndLabel &ticksAndLabel() const;
```
milianw: return by value instead of `const&`, `QVector` is COW so that's cheap

src/models/timeaxisheaderview.h
```diff
@@ -0,0 +1,86 @@
...
+private:
+    TicksAndLabel m_ticksAndLabels;
+
+
+    // Utility
+public:
+    /** Returns a power of 10 multiple of 1, 2 or 5 for the given value
```
milianw: ah, here's the comment. please move both helpers to either `util.h/cpp` or to a separate `plotutil.h/cpp` header

tests/test-clients/cpp-threadnames/main.cpp
```diff
@@ -34,7 +34,11 @@ int main()
 {
     for (int i = 0; i < 10; ++i) {
         thread t([i]() {
+#if defined(Q_OS_OSX)
             pthread_setname_np(pthread_self(), ("threadname" + to_string(i)).c_str());
```
milianw: that's a good change but ideally part of a separate merge request :)

src/models/timeaxisheaderview.h
```diff
@@ -0,0 +1,86 @@
...
+private:
+    FilterAndZoomStack* m_filterAndZoomStack = nullptr;
+public slots:
+    void onFilterChanged(const Data::FilterAction &filterAction);
+    void onZoomChanged(const Data::ZoomAction &zoomAction);
+protected:
+    void paintSection(QPainter *painter, const QRect &rect, int logicalIndex) const override;
+};
+
+class PrefixTickLabels
```
milianw: this class is private to the header view, no? so please move the declaration of this class either into an anon namespace within the .cpp file, or alternatively move it to a separate file .h/.cpp file pair

koenpoppe:
I will look into the requested changed. Meanwhile, here's a screenshot:

timeaxisheaderview

src/models/timeaxisheaderview.cpp
koenpoppe: Indeed, this is for debugging only and should be removed. Thank you for the hint related to the `QT_MESSAGE_PATTERN`, I wasn't aware the function name could be added that way.

src/models/timeaxisheaderview.cpp
koenpoppe: The timeline is only drawn above the events, this has been made clear in the code

src/models/timeaxisheaderview.cpp
koenpoppe: This is necessary to avoid rounding errors, the second part is a si prefix as can be seen on the screenshot below.

koenpoppe:
Remarks have been addressed. The screenshot illustrates the two-level approach.

koenpoppe:
I'm unsure if my cmake integration is up to par, but I believe the submodule approach more suited.

3rdparty/PrefixTickLabels.cmake
```diff
@@ -0,0 +1,14 @@
+add_library(PrefixTickLabels STATIC
+    PrefixTickLabels/src/PrefixTickLabels.cpp
+    )
+include_directories(
```
milianw: this shouldn't be required as you add the include dir to the target too

milianw:
finally some time to look at this. I've rebased this locally and also have made a few fixes locally - can you allow me to push to your branch such that we can work together on bringing this forward?

I'll add more notes to the individual commits

src/models/timeaxisheaderview.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include "eventmodel.h"
 #include "filterandzoomstack.h"
+#include "util.h"
```
milianw, line 9: ../util.h otherwise this doesn't compile for me

src/models/timeaxisheaderview.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include "../../3rdparty/PrefixTickLabels/src/PrefixTickLabels.h"
+
+TimeAxisHeaderView::TimeAxisHeaderView(FilterAndZoomStack* filterAndZoomStack, Qt::Orientation orientation,
+                                       QWidget* parent)
+    : QHeaderView(orientation, parent)
+    , m_filterAndZoomStack(filterAndZoomStack)
+{
+    setMinimumHeight(37);
```
milianw, line 18: this hardcoded number won't work, it's probably dependent on the font height? so something like using the double line height from the fontmetrics?

src/models/timeaxisheaderview.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "data.h"
+
+class FilterAndZoomStack;
+
+class TimeAxisHeaderView : public QHeaderView
+{
+    Q_OBJECT
+public:
+    explicit TimeAxisHeaderView(FilterAndZoomStack* filterAndZoomStack, Qt::Orientation orientation,
```
milianw, line 14: the orientation should always be horizontal, never vertical - so no need to pass it along

milianw:
I pushed the changes to the ticklabels branch as I didn't want to force-push to your master branch. could you change the source branch for the pull request to that repo, if possible? otherwise we'll have to create a new pull request

koenpoppe:
Feel free to push/merge to my master branch as there is no way to change the source branch of this merge request. I will wait with the other remarks until your changes are merged.

milianw:
done, would be cool if you could have another shot at it. most notably, the strange sizing issue of the timeline column is important. after that, I can actually play around with this new feature for real to see how it feels in practice

much appreciated, cheers

milianw:
hmm wtf why is this pull now closed? and why can I not reopen it?! sorry for that mess :(

milianw:
it says there are "no new commits" so potentially we can reopen after you've pushed some new commit/fixup to your master branch?

koenpoppe:
No problem, I'll go over it again with your changes and remarks and will re-open this pull request once done.

milianw:
hey, any news on this?

src/models/timeaxisheaderview.cpp
koenpoppe: Changed to be calculated based on the font size.

src/models/timeaxisheaderview.cpp
koenpoppe: Correct. This has been fixed.

src/models/timeaxisheaderview.h
koenpoppe: Header texts and sort indicators are restored. The layout of the ticks and the texts is improved.

koenpoppe:
I've found time to work on this and created the new pull request #238.

PR #232, opened 2020-03-26
Catskul/fix allow compile qt 5 9

catskul:
Allow compile with Qt 5.9:

 * Adds in preprocessor ifs around `RecursiveSort` to allow compilation with 18.04s default qt install qt-5.9. (Might be possible to drop down to qt 5.7 (debian 9s default)
 * appeases `qCompare` link error related to inexact type comparison
 * updates `perfparser` pointer to version which also resolves qCompare issues there
 * requires merge of https://github.com/KDAB/perfparser/pull/8 first

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ before_install:
     
 install: 
   - sudo apt-get -y install qt510base qt510svg qt510x11extras
+  - sudo apt-get -y install mesa-common-dev libglu1-mesa-dev
   - source /opt/qt*/bin/qt*-env.sh
   - git clone git://anongit.kde.org/extra-cmake-modules
   - cd extra-cmake-modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: cpp
 compiler: gcc
 sudo: require
-dist: trusty
+dist: xenial
 
 before_install:
-  - sudo add-apt-repository ppa:beineri/opt-qt-5.10.1-trusty -y
+  - sudo add-apt-repository ppa:beineri/opt-qt-5.10.1-xenial -y
   - sudo apt-get update -qq
     
 install: 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ set(HOTSPOT_VERSION_STRING "${hotspot_VERSION}")
 
 include(FeatureSummary)
 
-find_package(Qt5 5.10 COMPONENTS Core Widgets Network Test Svg REQUIRED)
+find_package(Qt5 5.9 COMPONENTS Core Widgets Network Test Svg REQUIRED)
 find_package(LibElf REQUIRED)
 find_package(Elfutils REQUIRED) # TODO: make optional, use internal copy
 find_package(ECM 1.0.0 NO_MODULE REQUIRED)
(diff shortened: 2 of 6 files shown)


catskul:
CI failing with:
```
/home/travis/build/KDAB/hotspot/src/resultspage.cpp: In member function ‘void ResultsPage::navigateToCodeFailed(const QString&)’:
/home/travis/build/KDAB/hotspot/src/resultspage.cpp:272:29: error: ‘std::chrono’ has not been declared
     QTimer::singleShot(std::chrono::seconds(5), ui->errorWidget, &KMessageWidget::animatedHide);
```

Investigating.

catskul:
Arcane problem w.r.t. Qt chrono support..???

```
In file included from /opt/qt510/include/QtCore/QTimer:1:0,
                 from /home/travis/build/KDAB/hotspot/src/resultspage.cpp:51:
/opt/qt510/include/QtCore/qtimer.h: In instantiation of ‘static void QTimer::singleShot(Duration, const typename QtPrivate::FunctionPointer<Func2>::Object*, Func1) [with Duration = std::chrono::duration<long int>; Func1 = void (KMessageWidget::*)(); typename QtPrivate::FunctionPointer<Func2>::Object = KMessageWidget]’:
/home/travis/build/KDAB/hotspot/src/resultspage.cpp:274:95:   required from here
/opt/qt510/include/QtCore/qtimer.h:104:53: error: no matching function for call to ‘QTimer::defaultTypeFor(std::chrono::duration<long int>&)’
         singleShot(interval, defaultTypeFor(interval), receiver, slot);
                                                     ^
/opt/qt510/include/QtCore/qtimer.h:104:53: note: candidate is:
/opt/qt510/include/QtCore/qtimer.h:208:43: note: static constexpr Qt::TimerType QTimer::defaultTypeFor(int)
     static Q_DECL_CONSTEXPR Qt::TimerType defaultTypeFor(int msecs) Q_DECL_NOTHROW
                                           ^
/opt/qt510/include/QtCore/qtimer.h:208:43: note:   no known conversion for argument 1 from ‘std::chrono::duration<long int>’ to ‘int’

make[2]: *** [src/CMakeFiles/hotspot.dir/resultspage.cpp.o] Error 1
```

https://development.qt-project.narkive.com/cOriRCay/qtimer-setinterval-and-chrono

investigating

catskul:
apparently some problem with old compiler in 14.04, but I can't figure out why this PR would make any difference. All previous builds using 14.04 should have failed in a similar way AFAIK.

catskul:
Fixed it but not sure if my fix will be acceptable as I moved the ubuntu version forward to 16.04

milianw:
Sorry, but I don't think this is a change I like to merge upstream. Most notably, removing the recursive filtering essentially cripples the search and makes it completely unusable.

Can't we leave it as-is and rely on AppImages instead? Or alternatively, download a newer Qt (e.g. from the official website, despite the annoying login) and build against that one instead of the distro-provided one?

milianw:
ping? any input on my remarks above?

catskul:
Your objection makes sense.

I can simply point them at my branch for the near future until we upgrade our base development environment.

I've not tried the appiamge. I'll look into it. I imagine I'll have a hard time getting my colleagues to use app images mostly because they're (and I am) not used to it. They seem to prefer either a deb package, or compiling from source. Unfortunately installing a newer Qt on 18.04 can cause a cascade of compatibility problems for other build from source packages so the Appimage or my branch will have to do for us.

milianw:
Thank you for the input. Please try the AppImages - it sounds like an ideal solution to your problem, really.

PR #233, opened 2020-03-29
Keep current reults page when reloading perf data

erijo:
It's nice to stay at e.g the flame graph after pressing F5 to get the latest data.


milianw:
thanks!

PR #248, opened 2020-08-04
Report number of missing/total symbols per module

### LINKED ISSUE #54: Display which debug symbols are missing in summary
Without debug symbols, it is hard to interpret the data because the function name can not be displayed. Even worse is that many backtraces will be broken because unwinding fails, resulting in wrong data being displayed.

Hotspot should show which debug symbols are missing in the summary view, to alert the user about the problem.

Idealy it would even show how many samples make use of the library without debug symbols, so one can judge how important the debug symbols for that library are.

raphinesse:
This change extends the information contained in the error message displayed to the user when a module is missing debug symbols.

Suppose that, in the recorded samples, there are 40 locations in module `ld-2.30.so`, two of which could not be resolved to a symbol. Then a message will be displayed in under _Summary/Parser Errors_.

#### Message before this change:

    Module "ld-2.30.so" is missing (some) debug symbols.

#### Message after this change:

    Module "ld-2.30.so" is missing 2 of 40 debug symbols.

This additional information should be sufficient to help the user determine whether a relevant amount of debug symbols are missing or not. So, IMHO this resolves #54

### DIFF
--- a/src/parsers/perf/perfparser.cpp
+++ b/src/parsers/perf/perfparser.cpp
@@ -814,6 +814,14 @@ public:
         }
 
         eventResult.totalCosts = summaryResult.costs;
+
+        for (auto i = numMissingSymbolsByModule.begin(); i != numMissingSymbolsByModule.end(); ++i) {
+            const auto& moduleName = strings.value(i.key());
+            const auto numMissing = i.value();
+            const auto numSymbols = numTotalSymbolsByModule.value(i.key());
+            summaryResult.errors << PerfParser::tr("Module \"%1\" is missing %2 of %3 debug symbols.")
+                .arg(moduleName).arg(numMissing).arg(numSymbols);
+        }
     }
 
     qint32 addCostType(const QString& label, Data::Costs::Unit unit)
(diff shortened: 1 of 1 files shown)


src/parsers/perf/perfparser.cpp
```diff
@@ -1182,7 +1190,8 @@ class PerfParserPrivate : public QObject
     Data::EventResults eventResult;
     QHash<qint32, QHash<qint32, QString>> commands;
     QScopedPointer<QTextStream> perfScriptOutput;
-    QSet<qint32> reportedMissingDebugInfoModules;
+    QHash<qint32, qint32> numTotalSymbolsByModule;
```
milianw, line 1193: please introduce a

```
struct SymbolCount {
    qint32 total = 0;
    qint32 missing = 0;
};
```
and then use that in one single hash, instead of having two hashes

src/parsers/perf/perfparser.cpp
```diff
@@ -909,10 +917,10 @@ class PerfParserPrivate : public QObject
         const auto binaryString = strings.value(symbol.symbol.binary.id);
         const auto pathString = strings.value(symbol.symbol.path.id);
         bottomUpResult.symbols[symbol.id] = {symbolString, binaryString, pathString};
-        if (symbolString.isEmpty() && !binaryString.isEmpty()
-            && !reportedMissingDebugInfoModules.contains(symbol.symbol.binary.id)) {
-            reportedMissingDebugInfoModules.insert(symbol.symbol.binary.id);
-            summaryResult.errors << PerfParser::tr("Module \"%1\" is missing (some) debug symbols.").arg(binaryString);
+
+        ++numTotalSymbolsByModule[symbol.symbol.binary.id];
```
milianw, line 921: once there's just one hash (see below), rewrite this like this:

```
auto &counter = numSymbolsByModule[symbol.symbol.binary.id];
++counter.total;
if (...)
    ++counter.missing;
```

src/parsers/perf/perfparser.cpp
raphinesse: I actually thought about doing that. The minor drawback is that you have to iterate over all modules in the end, filtering to those where `missing > 0`. In contrast, right now I can just iterate over `numMissingSymbolsByModule` and only get modules with missing symbols. That's why I kept it the way it is now.

Of course, that efficiency gain is surely negligible with the expected number of modules in a data set. So if you still prefer your suggested version I can change the code to use it. Should I?

milianw: yes please change it as proposed. in the iteration loop, just do something like:

```
if (!counter.missing)
    continue;
```

raphinesse: Sure, will do

milianw:
thanks

PR #261, opened 2020-10-14
Adding libkf5notifications-dev dependency

dmikushin:
I'm not tracking other distros, there similar fixes might be needed.


milianw:
thanks

PR #273, opened 2020-12-18
Further work on the disassembly feature.

dknysh:
Hello, Milian.
You requested some Further work on disassembly feature.

a) Hoist DisassemblyOutput::disassemble code chunk out into a separate
file and then build a simple test which runs this on some of the known symbols
in one of our existing test binaries, such as

../3rdparty/perfparser/tests/auto/perfdata/vector_static_gcc/
vector_static_gcc_v9.1.0

Could you please review the implementation for the point.

P.S. Could you please also clarify what do you mean as 'proper item model' in the second point - 

       b) We need a proper item model implementation on top of the DisassemblyOutput.
           Together with a model test then too.

Thank you,
Darya.


src/models/disassemblyoutput.h
```diff
@@ -0,0 +1,48 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <QString>
+#include "data.h"
```
milianw, line 29: don't include the header here, forward declare the Symbol struct instead

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -0,0 +1,105 @@
...
+    void testFiles()
+    {
+        QFETCH(QString, binaryFile);
+
+        const auto actualBinaryFile = QFINDTESTDATA(binaryFile);
+        QVERIFY(!actualBinaryFile.isEmpty() && QFile::exists(actualBinaryFile));
+        const auto expectedOutputFile = actualBinaryFile + ".expected.txt";
+        const auto actualOutputFile = actualBinaryFile + ".actual.txt";
+
+        Data::Symbol curSymbol = {QLatin1String("__cos_fma"),
```
milianw, line 64: pass the symbol in as another QTest column, then only patch/overwrite the actualBinaryFile here

milianw: also: here and below these should be QStringLiteral instead of QLatin1String, or remove it altogether - in a test that's OK imo

src/resultsdisassemblypage.h
```diff
@@ -29,6 +29,7 @@
 
 #include "data.h"
 #include "models/costdelegate.h"
+#include "models/disassemblyoutput.h"
```
milianw, line 32: forward declare instead and only include in .cpp file

milianw (changes requested):
thank you!

regarding the model: I mean implementing a real model instead of using QStandardItemModel which is notoriously bad and should only be used for prototyping purposes. Basically it would be a custom class that inherits from QAbstractTableModel and operaters directly on the `QVector<DisassemblyOutput::DisassemblyLine>` and corresponding costs internally then

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+        DisassemblyOutput disassemblyOutput = DisassemblyOutput::disassemble("objdump","x86_64", symbol);
+        QStringList actualTextList = QStringList();
+        for (int row = 0; row < disassemblyOutput.disassemblyLines.size(); row++) {
+            const auto &disassemblyLine = disassemblyOutput.disassemblyLines.at(row);
+            actualTextList << QString::number(disassemblyLine.addr) << disassemblyLine.disassembly
+                           << QLatin1String("\n");
+        }
+        QString actualText = actualTextList.join(" ");
+        QVector<DisassemblyOutput::DisassemblyLine> disassemblyLines = disassemblyOutput.disassemblyLines;
```
milianw, line 81: unused, remove

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -0,0 +1,107 @@
...
+        {
+            QFile actual(actualOutputFile);
```
milianw, line 83: open the file earlier above, then write directly to it via a QTextStream in the loop above instead of first building a list and then joining it

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -0,0 +1,107 @@
...
+        const auto actualBinaryFile = QFINDTESTDATA(symbol.binary);
+        symbol.actualPath = actualBinaryFile;
+
+        QVERIFY(!actualBinaryFile.isEmpty() && QFile::exists(actualBinaryFile));
+        const auto expectedOutputFile = actualBinaryFile + ".expected.txt";
+        const auto actualOutputFile = actualBinaryFile + ".actual.txt";
+
+        DisassemblyOutput disassemblyOutput = DisassemblyOutput::disassemble("objdump","x86_64", symbol);
+        QStringList actualTextList = QStringList();
+        for (int row = 0; row < disassemblyOutput.disassemblyLines.size(); row++) {
```
milianw, line 75: for (const auto &disassemblyLine : disassemblyOutput.disassemblyLines);

milianw (changes requested):
cool, three more little cleanups then this can go in

dknysh:
Updated accordingly your remarks.

milianw:
thank you

PR #297, opened 2021-06-16
Offer "off CPU recording" also when the perf man page is missing

dfaure-kdab:
The code calls `perf record --help` to detect the supported features,
but when building perf from sources, if asciidoc (and its million
dependencies) isn't installed, the man page isn't built, and `perf
record --help` only says "No manual entry for perf-record".

When in doubt, offer the features, the user will find out if it works or
not.


milianw:
thanks

PR #298, opened 2021-06-18
Cache sysroot and other paths

lievenhey:
I added an option to save different environments in a config and added an option to quickly switch between them.

this makes it easier to work with different sysroots

### DIFF
--- a/src/multiconfigwidget.cpp
+++ b/src/multiconfigwidget.cpp
@@ -1,0 +1,135 @@
+/*
+  multiconfigwidget.cpp
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2016-2021 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Lieven Hey <lieven.hey@kdab.com>
+
+  Licensees holding valid commercial KDAB Hotspot licenses may use this file in
+  accordance with Hotspot Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "multiconfigwidget.h"
+
+#include <QComboBox>
+#include <QHBoxLayout>
+#include <QLineEdit>
+#include <QPushButton>
+
+MultiConfigWidget::MultiConfigWidget(QWidget* parent)
+    : QWidget(parent)
+{
+    auto* layout = new QHBoxLayout(this);
+    layout->setMargin(0);
+
+    m_comboBox = new QComboBox(this);
+    m_comboBox->setEditable(true);
+    m_comboBox->setInsertPolicy(QComboBox::InsertAtCurrent);
+    m_comboBox->setDisabled(true);
+    layout->addWidget(m_comboBox);
+
+    connect(m_comboBox->lineEdit(), &QLineEdit::editingFinished, this, [this] {
+        m_config.deleteGroup(m_comboBox->currentData().toString());
+        saveConfigAs(m_comboBox->currentText());
+        m_comboBox->setItemData(m_comboBox->currentIndex(), m_comboBox->currentText());
+        m_config.sync();
+    });
+
+    connect(m_comboBox, QOverload<int>::of(&QComboBox::currentIndexChanged), this,
+            [this] { selectConfig(m_comboBox->currentData().toString()); });
+
+    m_copyButton = new QPushButton(this);
+    m_copyButton->setText(tr("Copy Config"));
+    layout->addWidget(m_copyButton);
+
+    connect(m_copyButton, &QPushButton::clicked, this, [this] {
+        const QString name = tr("Config %1").arg(m_comboBox->count() + 1);
+        saveConfigAs(name);
+        m_comboBox->addItem(name, name);
+        m_comboBox->setDisabled(false);
+    });
+
+    m_removeButton = new QPushButton(this);
+    m_removeButton->setText(tr("Remove Config"));
+    layout->addWidget(m_removeButton);
+
+    connect(m_removeButton, &QPushButton::clicked, this, [this] {
+        m_config.deleteGroup(m_comboBox->currentData().toString());
+        m_comboBox->removeItem(m_comboBox->currentIndex());
+
+        if (m_comboBox->count() == 0) {
+            m_comboBox->setDisabled(true);
+        } else {
+            selectConfig(m_comboBox->currentData().toString());
+        }
+    });
+
+    setLayout(layout);
+}
+
+MultiConfigWidget::~MultiConfigWidget() = default;
+
+QString MultiConfigWidget::currentConfig() const
+{
+    return m_comboBox->currentData().toString();
+}
+
+void MultiConfigWidget::setConfig(const KConfigGroup& group)
+{
+    m_comboBox->clear();
+    m_config = group;
+
+    const auto groups = m_config.groupList();
+    for (const auto& config : groups) {
+        if (m_config.hasGroup(config)) {
+            // item data is used to get the old name after renaming
+            m_comboBox->addItem(config, config);
+            m_comboBox->setDisabled(false);
+        }
+    }
+}
+
+void MultiConfigWidget::saveConfigAs(const QString& name)
+{
+    if (!name.isEmpty()) {
+        emit saveConfig(m_config.group(name));
+    }
+}
+
+void MultiConfigWidget::updateCurrentConfig()
+{
+    if (m_comboBox->currentIndex() != -1) {
+        saveConfigAs(m_comboBox->currentData().toString());
+    }
+}
+
+void MultiConfigWidget::selectConfig(const QString& name)
+{
+    m_config.sync();
+    if (!name.isEmpty() && m_config.hasGroup(name)) {
+        emit restoreConfig(m_config.group(name));
+    }
+}
+
+void MultiConfigWidget::restoreCurrent()
+{
+    if (m_comboBox->currentIndex() != -1) {
+        selectConfig(m_comboBox->currentData().toString());
+    }
+}
--- a/src/multiconfigwidget.h
+++ b/src/multiconfigwidget.h
@@ -1,0 +1,64 @@
+/*
+  multiconfigwidget.cpp
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2016-2020 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Lieven Hey <lieven.hey@kdab.com>
+
+  Licensees holding valid commercial KDAB Hotspot licenses may use this file in
+  accordance with Hotspot Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef MULTICONFIGWIDGET_H
+#define MULTICONFIGWIDGET_H
+
+#include <KConfigGroup>
+#include <QWidget>
+
+class QComboBox;
+class QPushButton;
+
+class MultiConfigWidget : public QWidget
+{
+    Q_OBJECT
+public:
+    MultiConfigWidget(QWidget* parent = nullptr);
+    ~MultiConfigWidget();
+
+    QString currentConfig() const;
+
+signals:
+    void saveConfig(KConfigGroup group);
+    void restoreConfig(const KConfigGroup& group);
+
+public slots:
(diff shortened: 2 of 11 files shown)


src/settingsdialog.cpp
```diff
@@ -28,10 +28,19 @@
...
-#include <KUrlRequester>
 #include <KComboBox>
+#include <KUrlRequester>
+#include <kconfiggroup.h>
+#include <ksharedconfig.h>
+#include <QKeyEvent>
+#include <QLineEdit>
 #include <QListView>
 
+KConfigGroup config()
```
milianw, line 39: wrap in anon namespace to give file local linkage

src/settingsdialog.cpp
```diff
@@ -58,10 +67,55 @@ SettingsDialog::SettingsDialog(QWidget* parent)
     };
     auto lastExtraLibsWidget = setupMultiPath(ui->extraLibraryPaths, ui->extraLibraryPathsLabel, ui->lineEditApplicationPath);
     setupMultiPath(ui->debugPaths, ui->debugPathsLabel, lastExtraLibsWidget);
+
+    const auto configGroups = config().groupList();
+    auto configfile = config();
+    for (const auto& configName : configGroups) {
+        if (configfile.hasGroup(configName)) {
```
milianw, line 74: this looks like it's duplicating a lot of code with the record page, can we share the logic somehow?

src/settingsdialog.cpp
```diff
@@ -124,3 +178,69 @@ QString SettingsDialog::objdump() const
 {
     return ui->lineEditObjdump->text();
 }
+
+void SettingsDialog::keyPressEvent(QKeyEvent* event)
+{
+    // disable the return -> accept policy since it prevents the user from confirming name changes in the combobox
+    // you can still press CTRL + Enter to close the dialog
+    switch (event->key()) {
```
milianw, line 186:
```
if (event->modifiers() != Qt::Key_Control && (event->key() == Qt::Key_Enter || event->key() == Qt::Key_Return)) {
    return;
}
QDialog::keyPressEvent(event);
```

milianw (changes requested):

src/settingsdialog.cpp
lievenhey: I created a custom widget that does this

lievenhey:
Please let me know if I need to add a License text

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,115 @@
+#include "multiconfigwidget.h"
```
milianw, line 1: missing license header

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    m_restoreFunction = restoreFunction;
+}
+
+QString MultiConfigWidget::currentConfig() const
+{
+    return m_comboBox->currentData().toString();
+}
+
+void MultiConfigWidget::setKConfigGroup(const KConfigGroup& group)
```
milianw, line 72: rename to `setConfig`

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,115 @@
...
+{
+    saveConfigAs(m_comboBox->currentData().toString());
+}
+
+void MultiConfigWidget::selectConfig(const QString& name)
+{
+    m_config.sync();
+    if (!m_restoreFunction)
+        return;
+    auto t = m_config.groupList();
```
milianw, line 104: unused statement, remove

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,115 @@
...
+    if (m_config.hasGroup(name) && !name.isEmpty()) {
```
milianw, line 105: flip conditionals, i.e. first check the cheap operation (`isEmpty`), then the potentially more expensive one (`hasGroup`)

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,115 @@
...
+        m_restoreFunction(m_config.group(name));
+    }
+}
+
+void MultiConfigWidget::restoreCurrent()
+{
+    if (m_comboBox->currentIndex() != -1) {
```
milianw, line 112: check isn't needed, if `currentIndex` is `-1` then `currentData.toString()` will return an empty string which is checked in `selectConfig`.

but it would be fine to keep it here, but then it should probably added to  `updateCurrentConfig` too e.g., no? i.e. to make this code somewhat more consistent?

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,115 @@
...
+            // item data is used to get the old name after renaming
+            m_comboBox->addItem(config, config);
+            m_comboBox->setDisabled(false);
+        }
+    }
+}
+
+void MultiConfigWidget::saveConfigAs(const QString& name)
+{
+    if (m_saveFunction) {
```
milianw, line 89: should this check the name for validity similar to how you are doing it in `selectConfig` below? if not, why not?

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
+#ifndef MULTICONFIGWIDGET_H
```
milianw, line 1: missing license header

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
...
+#include <QComboBox>
+#include <QPushButton>
+#include <QWidget>
+
+class MultiConfigWidget : public QWidget
+{
+    Q_OBJECT
+public:
+    explicit MultiConfigWidget(QWidget* parent = nullptr);
+
```
milianw, line 15: missing dtor, with `= default` in .cpp file

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
...
+    QString currentConfig() const;
+
+public slots:
+    void setKConfigGroup(const KConfigGroup& group);
+    void saveConfigAs(const QString& name);
+    void updateCurrentConfig();
+    void selectConfig(const QString& name);
+    void restoreCurrent();
+
+private slots:
```
milianw, line 27: remove

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
...
+
+private:
+    std::function<void(KConfigGroup)> m_saveFunction;
+    std::function<void(const KConfigGroup&)> m_restoreFunction;
+    KConfigGroup m_config;
+    QComboBox* m_comboBox;
```
milianw, line 33: here and below: `= nullptr`

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
+#ifndef MULTICONFIGWIDGET_H
+#define MULTICONFIGWIDGET_H
+
+#include <KConfigGroup>
+#include <functional>
+#include <QComboBox>
```
milianw, line 6: forward declare

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
...
+#include <QPushButton>
```
milianw, line 7: dito

src/recordpage.cpp
```diff
@@ -207,24 +207,30 @@ RecordPage::RecordPage(QWidget* parent)
...
-    const auto updateLaunchConfig = [this] {
-        if (ui->launchConfigComboBox->count() > 0) {
-            saveApplicationConfig(ui->launchConfigComboBox->currentText());
-        }
+    m_multiConfig = new MultiConfigWidget(this);
+
+    m_multiConfig->setKConfigGroup(applicationConfig(ui->applicationName->text()));
+
+    auto saveFunction = [this](KConfigGroup group) {
+        group.writeEntry(QLatin1String("params"), ui->applicationParametersBox->text());
```
milianw, line 215: why the QL1S here now suddenly? shouldn't be needed, and isn't used for `readEntry` below either?

src/multiconfigwidget.h
```diff
@@ -0,0 +1,38 @@
...
+#include <QWidget>
+
+class MultiConfigWidget : public QWidget
+{
+    Q_OBJECT
+public:
+    explicit MultiConfigWidget(QWidget* parent = nullptr);
+
+    void setSaveFunction(const std::function<void(KConfigGroup)>& saveFunction);
```
milianw, line 16: shouldn't both be set through the ctor already - I mean having this widget without either one leads to useless/broken behavior, no?

i.e. I think as-is this is bad API design. make both args of the ctor, _and_ assert the validity of both `std::function`'s, then remove the other runtime checks that check if either was set or not

src/recordpage.cpp
```diff
@@ -207,24 +207,30 @@ RecordPage::RecordPage(QWidget* parent)
...
+        group.writeEntry(QLatin1String("workingDir"), ui->workingDirectory->text());
+    };
+
+    m_multiConfig->setSaveFunction(saveFunction);
```
milianw, line 219: both callbacks should be moved instead of copied

src/settingsdialog.cpp
```diff
@@ -32,13 +32,16 @@
 #include <KUrlRequester>
 #include <kconfiggroup.h>
 #include <ksharedconfig.h>
+#include <QKeyEvent>
 #include <QLineEdit>
 #include <QListView>
 
+namespace {
```
milianw, line 41: unrelated change, but a good one - feel free to keep it as-is for now, but in the future try to keep your commits atomic and focused on one thing only

milianw (changes requested):
cool, mostly nits to cleanup the last commit, then this can go in 

thanks!

src/multiconfigwidget.h
lievenhey: I cant forward declare KConfigGroup but I did it with QPushButton and QComboBox

src/settingsdialog.cpp
lievenhey: This was your first change request

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,141 @@
...
+*/
+
+#include "multiconfigwidget.h"
+
+#include <QComboBox>
+#include <QHBoxLayout>
+#include <QLineEdit>
+#include <QPushButton>
+
+MultiConfigWidget::MultiConfigWidget(const std::function<void(KConfigGroup)>&& saveFunction,
```
milianw, line 35: rvalue args are bad practice, just move at the call side

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (m_comboBox->count() == 0) {
+            m_comboBox->setDisabled(true);
+        } else {
+            selectConfig(m_comboBox->currentData().toString());
+        }
+    });
+
+    setLayout(layout);
+
+    m_saveFunction = saveFunction;
```
milianw, line 86: use member initializer instead (see above)

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "multiconfigwidget.h"
+
+#include <QComboBox>
+#include <QHBoxLayout>
+#include <QLineEdit>
+#include <QPushButton>
+
+MultiConfigWidget::MultiConfigWidget(const std::function<void(KConfigGroup)>&& saveFunction,
+                                     const std::function<void(const KConfigGroup&)>&& restoreFunction, QWidget* parent)
+    : QWidget(parent)
```
milianw, line 37: directly initialize the save/restore functions here

```
    , m_saveFunction(std::move(saveFunction))
    , m_restoreFunction(std::move(restoreFunction))
```

src/multiconfigwidget.h
```diff
@@ -0,0 +1,64 @@
...
+#include <QWidget>
+
+class QComboBox;
+class QPushButton;
+
+class MultiConfigWidget : public QWidget
+{
+    Q_OBJECT
+public:
+    MultiConfigWidget(const std::function<void(KConfigGroup)>&& saveFunction,
```
milianw, line 42: take by value instead of by rvalue ref

src/recordpage.cpp
```diff
@@ -207,24 +207,26 @@ RecordPage::RecordPage(QWidget* parent)
...
+
+    auto restoreFunction = [this](const KConfigGroup& group) {
+        ui->applicationParametersBox->setText(group.readEntry("params", ""));
+        ui->workingDirectory->setText(group.readEntry("workingDir", ""));
+        setError({});
     };
 
-    connect(ui->applicationParametersBox, &QLineEdit::editingFinished, this, updateLaunchConfig);
-    connect(ui->workingDirectory, &KUrlRequester::textChanged, this, updateLaunchConfig);
+    m_multiConfig = new MultiConfigWidget(saveFunction, restoreFunction, this);
```
milianw, line 221: `std::move(saveFunction), std::move(restoreFunction)`

side note: does your code even compile?! you are taking an rvalue ref, but you are not moving here, so this shouldn't compile at all as-is?

src/settingsdialog.cpp
milianw: now I understand: you squashed this change into the wrong patch, it should be part of the first patch, not the second

milianw (changes requested):

src/recordpage.cpp
lievenhey: yes

lievenhey: seems the compiler can detect this

src/settingsdialog.cpp
lievenhey: done is now in the correct commit

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,140 @@
...
+void MultiConfigWidget::updateCurrentConfig()
+{
+    if (m_comboBox->currentIndex() != -1) {
+        saveConfigAs(m_comboBox->currentData().toString());
+    }
+}
+
+void MultiConfigWidget::selectConfig(const QString& name)
+{
+    Q_ASSERT(m_restoreFunction != nullptr);
```
milianw, line 128: just `Q_ASSERT(m_restoreFunction)`

src/multiconfigwidget.cpp
```diff
@@ -0,0 +1,140 @@
...
+            // item data is used to get the old name after renaming
+            m_comboBox->addItem(config, config);
+            m_comboBox->setDisabled(false);
+        }
+    }
+}
+
+void MultiConfigWidget::saveConfigAs(const QString& name)
+{
+    Q_ASSERT(m_saveFunction != nullptr);
```
milianw, line 113: just `Q_ASSERT(m_saveFunction)`

src/recordpage.cpp
milianw: still, please move it here explicitly if you stick to `std::function`

but, quite frankly, giving this a second thought I would believe using signals instead would be much nicer:

in the config widget, you'd just emit a signal instead of calling the save/restore function

and here after constructing the widget, you'd just connect to the signals to the lambdas you have above

milianw (changes requested):

milianw (approved):
lgtm, thanks

PR #299, opened 2021-06-18
Ansi colors

lievenhey:
some programs use ansi escape codes to display color. Currently this breaks the output a bit by inserting unprintable characters.

This change contains a small parser that turns the sequences into html to display the colors correctly.

Stuff that might not work correctly:
- combination of insertHtml and insertText
- some text formating might break in insertHtml

I would recommend adding an option to switch between removed escape sequences and display colors.

### DIFF
--- a/src/recordpage.cpp
+++ b/src/recordpage.cpp
@@ -36,23 +36,29 @@
 #include <QDebug>
 #include <QKeyEvent>
 #include <QListView>
+#include <QScrollArea>
 #include <QShortcut>
 #include <QStandardItemModel>
 #include <QStandardPaths>
+#include <QTemporaryFile>
 #include <QTimer>
-#include <QScrollArea>
 #include <QtConcurrent/QtConcurrentRun>
 
 #include <KColumnResizer>
 #include <KComboBox>
 #include <KConfigGroup>
+#include <KParts/ReadOnlyPart>
+#include <KService>
 #include <KSharedConfig>
 #include <KShell>
 #include <KUrlCompletion>
 #include <KUrlRequester>
 #include <Solid/Device>
 #include <Solid/Processor>
+#include <kde_terminal_interface.h>
 #include <kio_version.h>
+
+#include <QRegularExpression>
 
 #include "perfrecord.h"
 
@@ -459,6 +465,8 @@ RecordPage::RecordPage(QWidget* parent)
         if (index != -1)
             ui->compressionComboBox->setCurrentIndex(index);
     }
+
+    addKonsoleWidget();
 }
 
 RecordPage::~RecordPage() = default;
@@ -484,6 +492,9 @@ void RecordPage::onStartRecordingButtonClicked(bool checked)
         ui->startRecordingButton->setIcon(QIcon::fromTheme(QStringLiteral("media-playback-stop")));
         ui->startRecordingButton->setText(tr("Stop Recording"));
         ui->perfResultsTextEdit->clear();
+
+        // clear konsole
+        addKonsoleWidget();
 
         QStringList perfOptions;
 
@@ -739,15 +750,71 @@ void RecordPage::updateProcessesFinished()
 
 void RecordPage::appendOutput(const QString& text)
 {
+    if (m_konsolePart) {
+        m_konsoleFile->write(text.toUtf8());
+        m_konsoleFile->flush();
+        return;
+    }
+
+    // this regex finds ansi escapes
+    static const QRegularExpression regex(QStringLiteral("\\x1b\\[[0-9;]+m"));
+
     QTextCursor cursor(ui->perfResultsTextEdit->document());
     cursor.movePosition(QTextCursor::End);
-    cursor.insertText(text);
+
+    QString cleanOutput = text;
+    cursor.insertText(cleanOutput.replace(regex, QString()));
 }
 
 void RecordPage::setError(const QString& message)
 {
     ui->applicationRecordErrorMessage->setText(message);
     ui->applicationRecordErrorMessage->setVisible(!message.isEmpty());
+}
+
+void RecordPage::addKonsoleWidget()
+{
+    KService::Ptr service = KService::serviceByDesktopName(QStringLiteral("konsolepart"));
+
+    if (!service) {
+        return;
+    }
+
+    const auto tail = QStandardPaths::findExecutable(QStringLiteral("tail"));
+
+    if (tail.isEmpty()) {
+        return;
+    }
+
+    if (m_konsolePart) {
+        ui->recordOutputBox->layout()->removeWidget(m_konsolePart->widget());
+        m_konsolePart->deleteLater();
+    }
+    m_konsolePart = service->createInstance<KParts::ReadOnlyPart>(this);
+
+    if (!m_konsolePart) {
+        return;
+    }
+
+    const auto terminalInterface = qobject_cast<TerminalInterface*>(m_konsolePart);
+    if (!terminalInterface) {
+        qWarning("konsole kpart doesn't implement terminal interface");
+        delete m_konsolePart;
+        m_konsolePart = nullptr;
+        return;
+    }
+
+    if (m_konsoleFile) {
+        m_konsoleFile->deleteLater();
+    }
+    m_konsoleFile = new QTemporaryFile(this);
+    m_konsoleFile->open();
+
+    terminalInterface->startProgram(tail, {tail, QStringLiteral("-f"), m_konsoleFile->fileName()});
+    ui->recordOutputBoxLayout->insertWidget(1, m_konsolePart->widget());
+    ui->perfResultsTextEdit->setVisible(false);
+
+    m_konsolePart->widget()->setFocusPolicy(Qt::FocusPolicy::NoFocus);
 }
 
 void RecordPage::updateRecordType()
@@ -770,6 +837,9 @@ void RecordPage::updateRecordType()
     if (recordType == AttachToProcess) {
         updateProcesses();
     }
+
+    // clear konsole
+    addKonsoleWidget();
 
     updateStartRecordingButtonState(ui);
 }
--- a/src/recordpage.h
+++ b/src/recordpage.h
@@ -35,6 +35,7 @@
 #include "processlist.h"
 
 class QTimer;
+class QTemporaryFile;
 
 namespace Ui {
 class RecordPage;
@@ -43,6 +44,10 @@ class RecordPage;
 class PerfRecord;
 class ProcessModel;
 class ProcessFilterModel;
+
+namespace KParts {
+class ReadOnlyPart;
+}
 
 enum RecordType
 {
@@ -85,6 +90,7 @@ private:
     void updateRecordType();
     void appendOutput(const QString& text);
     void setError(const QString& message);
+    void addKonsoleWidget();
 
     QScopedPointer<Ui::RecordPage> ui;
 
@@ -92,6 +98,8 @@ private:
     QString m_resultsFile;
     QElapsedTimer m_recordTimer;
     QTimer* m_updateRuntimeTimer;
+    KParts::ReadOnlyPart* m_konsolePart = nullptr;
+    QTemporaryFile* m_konsoleFile = nullptr;
     MultiConfigWidget* m_multiConfig;
 
     ProcessModel* m_processModel;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,7 @@ find_package(KF5 COMPONENTS
     WindowSystem
     Notifications
     IconThemes
+    Parts
     REQUIRED
 )
 
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ As of now, you will need the following dependencies to build this project:
   - windowsystem
   - notifications
   - iconthemes
+  - parts
 - [KDDockWidgets](https://github.com/KDAB/KDDockWidgets)
   - this library is not yet packaged on most distributions, you'll have to compile it yourself from source
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -91,6 +91,7 @@ target_link_libraries(hotspot
     KF5::WindowSystem
     KF5::Notifications
     KF5::IconThemes
+    KF5::Parts
     models
     PrefixTickLabels
     KDAB::kddockwidgets
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -106,7 +106,7 @@ int firstAvailableIde()
 }
 
 MainWindow::MainWindow(QWidget* parent)
-    : QMainWindow(parent)
+    : KParts::MainWindow(parent)
     , ui(new Ui::MainWindow)
     , m_parser(new PerfParser(this))
     , m_config(KSharedConfig::openConfig())
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -27,10 +27,10 @@
 
 #pragma once
 
-#include <QMainWindow>
 #include <QScopedPointer>
 #include <QString>
 
+#include <KParts/MainWindow>
 #include <KSharedConfig>
 
 namespace Ui {
@@ -47,7 +47,7 @@ class ResultsPage;
 class RecordPage;
 class SettingsDialog;
 
-class MainWindow : public QMainWindow
+class MainWindow : public KParts::MainWindow
 {
     Q_OBJECT
 
--- a/src/recordpage.ui
+++ b/src/recordpage.ui
@@ -72,11 +72,11 @@
        </widget>
       </item>
       <item row="0" column="1">
-       <widget class="KUrlComboRequester" name="applicationName" native="true">
+       <widget class="KUrlComboRequester" name="applicationName">
         <property name="toolTip">
          <string>Path to the application to be recorded</string>
         </property>
-        <property name="text" stdset="0">
+        <property name="text">
          <string/>
         </property>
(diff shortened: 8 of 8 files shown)


lievenhey:
Example output:
grafik
Output from shell:
grafik

lievenhey:
I have added a combobox to let the user decide which method he wants to use.
It seems the font does not support bold characters

src/recordpage.cpp
```diff
@@ -801,9 +813,81 @@ QStringList RecordPage::applicationConfigurations()
 
 void RecordPage::appendOutput(const QString& text)
 {
+    // this regex finds ansi escapes and extracts all the numbers in it
+    QRegularExpression regex(QLatin1String("\\x1b\\[([0-9;]+)m"));
```
milianw, line 817: `static const auto regex = QRegularExpression(QStringLiteral(...));`

src/recordpage.cpp
```diff
@@ -801,9 +813,81 @@ QStringList RecordPage::applicationConfigurations()
...
+
     QTextCursor cursor(ui->perfResultsTextEdit->document());
     cursor.movePosition(QTextCursor::End);
-    cursor.insertText(text);
+
+    auto ansiToHtml = [](const QStringList& numbers) -> QString {
+        static bool openColor = false;
```
milianw, line 823: no mutable statics, move this out of the lambda and capture by ref instead

src/recordpage.cpp
```diff
@@ -801,9 +813,81 @@ QStringList RecordPage::applicationConfigurations()
...
+        // only foregroud colors are supported currently
+        const QHash<int, QColor> colors = {
```
milianw, line 825: move to separate function that has a simple switch statement to do the mapping, instead of requiring a qhash construction _every time this lambda is called_!

milianw: furthermore this won't work for systems using dark color schemes (which is what I use), we must not hard code the colors.

really, I think the proper solution here would be to figure out how to embed the konsole kpart to do this task for us, instead of trying to reinvent the wheel and suffering from all kinds of issues then

src/recordpage.cpp
```diff
@@ -801,9 +813,81 @@ QStringList RecordPage::applicationConfigurations()
...
+                    textList.pop_front();
+                }
+            }
+            colorText += textList.front();
+        } else {
+            colorText = text_;
+        }
+        // newline don't work in html
+        colorText = colorText.replace(QLatin1String("\n"), QLatin1String("<br>"));
+        qDebug() << colorText;
```
milianw, line 886: remove

src/recordpage.cpp
```diff
@@ -801,9 +813,81 @@ QStringList RecordPage::applicationConfigurations()
...
+        text_.replace(QLatin1String(" "), QLatin1String("&nbsp;"));
+        QString colorText;
+        auto matches = regex.globalMatch(text_);
+        if (matches.hasNext()) {
+            auto textList = text_.split(regex);
+            while (matches.hasNext()) {
+                auto match = matches.next();
+
+                if (textList.length() > 1) {
+                    colorText += textList.front() + ansiToHtml(match.captured(1).split(QLatin1String(";")));
```
milianw, line 876: capturedView(1) and https://www.kdab.com/qstringview-diaries-zero-allocation-string-splitting/

src/recordpage.cpp
milianw: all of this code should go to a separate function instead of pushing it tightly into the widget

we would also need separate unit testing

milianw: and instead of building lots of strings and concatenating them all the time, you should instead use a `QTextStream` and just put everything in there, then call insertHtml once on the document here in the widget

src/recordpage.cpp
```diff
@@ -801,9 +813,81 @@ QStringList RecordPage::applicationConfigurations()
...
+    case Default:
+        cursor.insertText(text);
+        break;
+    case RemoveAnsi: {
+        QString cleanOutput = text;
+        cursor.insertText(cleanOutput.replace(regex, QLatin1String("")));
+        break;
+    }
+    case DecodeAnsi: {
+        // html eats up spaces, replaces them to keep the formating
```
milianw, line 865: just wrap the whole output once in a `<pre>` instead

src/recordpage.cpp
milianw: you are also not properly escaping html-like output that may be available in the `text`

milianw (changes requested):
I'm very much on the fence about this once - I would much rather we don't reinvent the wheel and embed the konsole kpart somehow instead if possible and use that here. but I'm not sure if that's easily possible. if not, then working on that upstream to make it simpler would be really nice. alternatively maybe we can leverage https://github.com/lxqt/qterminal ?

alternatively, if you find a way to make this code work with a dark system color scheme (i.e. without hardcoding the color map), then I'd be OK with it, after some significant cleanup of the code

src/recordpage.cpp
lievenhey: As far as I know these colors are standardized

milianw: they are not:
Screenshot_20210810_101133

src/recordpage.cpp
```diff
@@ -52,7 +53,11 @@
 #include <KUrlRequester>
 #include <Solid/Device>
 #include <Solid/Processor>
+#include <kde_terminal_interface.h>
 #include <kio_version.h>
+#include <kservice.h>
```
milianw, line 58: `#include <KService>`, and move up above the `<Solid/` lines

src/recordpage.cpp
```diff
@@ -173,6 +178,8 @@ RecordPage::RecordPage(QWidget* parent)
     , ui(new Ui::RecordPage)
     , m_perfRecord(new PerfRecord(this))
     , m_updateRuntimeTimer(new QTimer(this))
+    , m_konsolePart(nullptr)
```
milianw, line 181: do this in the header already please

src/recordpage.cpp
```diff
@@ -454,6 +461,12 @@ RecordPage::RecordPage(QWidget* parent)
         if (index != -1)
             ui->compressionComboBox->setCurrentIndex(index);
     }
+
+#ifdef __linux__
```
milianw, line 465: this looks wrong, why do you check the platform? the konsole part could be available on other platforms too after all?

src/recordpage.cpp
```diff
@@ -454,6 +461,12 @@ RecordPage::RecordPage(QWidget* parent)
...
+    m_service = KService::serviceByDesktopName(QLatin1String("konsolepart"));
```
milianw, line 466: `QStringLiteral`

src/recordpage.cpp
```diff
@@ -823,9 +839,22 @@ QStringList RecordPage::applicationConfigurations()
 
 void RecordPage::appendOutput(const QString& text)
 {
+#ifdef __linux__
```
milianw, line 842: again, why?

src/recordpage.cpp
```diff
@@ -823,9 +839,22 @@ QStringList RecordPage::applicationConfigurations()
...
+    if (m_konsolePart) {
+        m_konsoleFile->write(text.toUtf8());
+        m_konsoleFile->flush();
+        return;
+    }
+#endif
+
+    // this regex finds ansi escapes and extracts all the numbers in it
+    QRegularExpression regex(QLatin1String("\\x1b\\[[0-9;]+m"));
```
milianw, line 851: `static const`, and use `QStringLiteral` instead

src/recordpage.cpp
```diff
@@ -823,9 +839,22 @@ QStringList RecordPage::applicationConfigurations()
...
 {
+#ifdef __linux__
+    if (m_konsolePart) {
+        m_konsoleFile->write(text.toUtf8());
+        m_konsoleFile->flush();
+        return;
+    }
+#endif
+
+    // this regex finds ansi escapes and extracts all the numbers in it
```
milianw, line 850: we don't extract anything

src/recordpage.cpp
```diff
@@ -823,9 +839,22 @@ QStringList RecordPage::applicationConfigurations()
...
+    QRegularExpression regex(QLatin1String("\\x1b\\[[0-9;]+m"));
+
     QTextCursor cursor(ui->perfResultsTextEdit->document());
     cursor.movePosition(QTextCursor::End);
-    cursor.insertText(text);
+
+    QString cleanOutput = text;
+    cursor.insertText(cleanOutput.replace(regex, QLatin1String("")));
```
milianw, line 857: `QString()` instead of `QL1S("")`

src/recordpage.cpp
```diff
@@ -834,6 +863,29 @@ void RecordPage::setError(const QString& message)
     ui->applicationRecordErrorMessage->setVisible(!message.isEmpty());
 }
 
+void RecordPage::addKonsoleWidget()
+{
+    if (m_service) {
```
milianw, line 868: reduce indent depth:

```
if (!m_service) {
    return;
}

const auto tail = ...;
if (tail.isEmpty()) {
    return;
}
```

src/recordpage.cpp
```diff
@@ -834,6 +863,29 @@ void RecordPage::setError(const QString& message)
...
+        const auto tail = QStandardPaths::findExecutable(QLatin1String("tail"));
+
+        if (!tail.isEmpty()) {
+            if (m_konsolePart) {
+                ui->recordOutputBox->layout()->removeWidget(m_konsolePart->widget());
+                m_konsolePart->deleteLater();
+            }
+            m_konsolePart = m_service->createInstance<KParts::ReadOnlyPart>(this);
+            m_konsoleFile = new QTemporaryFile(this);
```
milianw, line 877: check the validity of the kpart early instead of blindly using it and its casted variant:
```
if (!m_konsolePart) {
    return;
}
const auto terminalInterface = qobject_cast<TerminalInterface*>(m_konsolePart);
if (!terminalInterface) {
    qWarning("konsole kpart doesn't implement terminal interface");
    delete m_konsolePart;
    m_konsolePart = nullptr;
    return;
}
```

src/recordpage.cpp
```diff
@@ -834,6 +863,29 @@ void RecordPage::setError(const QString& message)
...
+            m_konsoleFile->open();
+            qobject_cast<TerminalInterface*>(m_konsolePart)
+                ->startProgram(tail, {QLatin1String("tail"), QLatin1String("-f"), m_konsoleFile->fileName()});
+            if (m_konsolePart) {
```
milianw, line 881: note how you already accessed the konsolePart - this check comes too late!

src/recordpage.h
```diff
@@ -27,13 +27,16 @@
 
 #pragma once
 
+#include <KParts/ReadOnlyPart>
```
milianw, line 30: forward declare

src/recordpage.cpp
```diff
@@ -173,6 +178,8 @@ RecordPage::RecordPage(QWidget* parent)
     , ui(new Ui::RecordPage)
     , m_perfRecord(new PerfRecord(this))
     , m_updateRuntimeTimer(new QTimer(this))
+    , m_konsolePart(nullptr)
+    , m_service(nullptr)
```
milianw, line 182: not needed for a smart ptr

src/recordpage.h
```diff
@@ -27,13 +27,16 @@
...
+#include <KService>
```
milianw, line 31: remove (see below)

src/recordpage.h
```diff
@@ -92,13 +95,17 @@ private slots:
...
+    void addKonsoleWidget();
 
     QScopedPointer<Ui::RecordPage> ui;
 
     PerfRecord* m_perfRecord;
     QString m_resultsFile;
     QElapsedTimer m_recordTimer;
     QTimer* m_updateRuntimeTimer;
+    KParts::ReadOnlyPart* m_konsolePart;
+    KService::Ptr m_service;
```
milianw, line 107: remove (see .cpp)

src/recordpage.cpp
milianw: and move this down to the single place where you are using the service, I don't see a reason why this needs to be cached

milianw (changes requested):
cool, this goes into the right direction, many thanks!

can you please check whether pressing `ctrl + c` when having the konsole part focused to stop the `tail` process e.g.? I think that would be acceptable at first and would then try to investigate how to improve upon that then

src/recordpage.cpp
lievenhey: As far as I know there is now tail on windows. Since it would not work I disabled it.

src/recordpage.h
lievenhey: I can remove the m_service but the m_konsolePart is required to clear the terminal before rerunning perf record

lievenhey:
ctrl+c stops the tail process. I think it might be possible to prevent this by using an event filter

src/recordpage.cpp
milianw: but its available on freebsd, mac and even on windows potentially one could supply `tail`

i.e.: this is the wrong way to handle this, instead you should only keep the failsafe mode to gracefully handle the situation where `tail` wasn't found (it may even not be available on linux after all)

src/recordpage.h
milianw: yes, that's fine - I only mean the m_service here

src/recordpage.cpp
```diff
@@ -834,6 +858,46 @@ void RecordPage::setError(const QString& message)
...
+
+    const auto terminalInterface = qobject_cast<TerminalInterface*>(m_konsolePart);
+    if (!terminalInterface) {
+        qWarning("konsole kpart doesn't implement terminal interface");
+        delete m_konsolePart;
+        m_konsolePart = nullptr;
+        return;
+    }
+
+    terminalInterface->startProgram(tail, {QLatin1String("tail"), QLatin1String("-f"), m_konsoleFile->fileName()});
```
milianw, line 896: use the `tail` here instead that you found above. and the `QLatin1String("-f")` should be a `QStringLiteral` instead

milianw (changes requested):

src/recordpage.cpp
lievenhey: good point. I removed the ifdef and tested both cases (konsolepart not found and tail not found). In both cases the fallback to the textarea worked.

src/recordpage.cpp
```diff
@@ -834,6 +856,46 @@ void RecordPage::setError(const QString& message)
...
+        ui->recordOutputBox->layout()->removeWidget(m_konsolePart->widget());
+        m_konsolePart->deleteLater();
+    }
+    m_konsolePart = service->createInstance<KParts::ReadOnlyPart>(this);
+
+    if (!m_konsolePart) {
+        return;
+    }
+
+    m_konsoleFile = new QTemporaryFile(this);
```
milianw, line 883: move down, below the cast check of the terminalInterface

also check if m_konsoleFile exists already and if so, delete it before recreating it here. otherwise we are "leaking" temporary files until hotspot is quit at the very end

src/recordpage.cpp
```diff
@@ -834,6 +856,46 @@ void RecordPage::setError(const QString& message)
...
+    const auto terminalInterface = qobject_cast<TerminalInterface*>(m_konsolePart);
+    if (!terminalInterface) {
+        qWarning("konsole kpart doesn't implement terminal interface");
+        delete m_konsolePart;
+        m_konsolePart = nullptr;
+        return;
+    }
+
+    terminalInterface->startProgram(tail, {tail, QStringLiteral("-f"), m_konsoleFile->fileName()});
+    qobject_cast<QVBoxLayout*>(ui->recordOutputBox->layout())->insertWidget(1, m_konsolePart->widget());
```
milianw, line 895: set the `layoutName` property of the `recordOutputBox` in `designer` to something stable, e.g. `recordOutputBoxLayout`. Then use it here directly without casting

src/recordpage.cpp
```diff
@@ -480,6 +488,9 @@ void RecordPage::onStartRecordingButtonClicked(bool checked)
         ui->startRecordingButton->setText(tr("Stop Recording"));
         ui->perfResultsTextEdit->clear();
 
+        // clear konsole
```
milianw, line 491: same should be done in `updateRecordType` below where `ui->perfResultsTextEdit->clear();` is also called

milianw (changes requested):

src/recordpage.h
```diff
@@ -92,13 +97,16 @@ private slots:
...
+    void addKonsoleWidget();
 
     QScopedPointer<Ui::RecordPage> ui;
 
     PerfRecord* m_perfRecord;
     QString m_resultsFile;
     QElapsedTimer m_recordTimer;
     QTimer* m_updateRuntimeTimer;
+    KParts::ReadOnlyPart* m_konsolePart = nullptr;
+    QTemporaryFile* m_konsoleFile;
```
milianw, line 109: either `= nullptr` it here, or make it a `std::unique_ptr` (which would help with the `.reset()` I proposed

milianw:
> ctrl+c stops the tail process. I think it might be possible to prevent this by using an event filter

can you please try to implement that then? or alternatively detect this and then close the debuggee too

lievenhey:
Sadly the event filter did not work as expected. But setting the focus policy to NoFocus prevents the user from pressing ctrl + c.

milianw (approved):

milianw:
cool, let's try this out and see how it goes from here on - many thanks Lieven!

PR #300, opened 2021-06-23
use kauth for privilege elevation

lievenhey:
kdesu is no longer maintained

I am still looking for a better name than authhelper because this is too generic


src/CMakeLists.txt
```diff
@@ -102,3 +103,11 @@ set_target_properties(hotspot
 
 install(TARGETS hotspot RUNTIME DESTINATION ${KDE_INSTALL_BINDIR})
 install(FILES hotspot.notifyrc DESTINATION ${KDE_INSTALL_KNOTIFY5RCDIR})
+
+add_executable(authhelper authhelper.cpp)
+target_link_libraries(authhelper KF5::AuthCore)
+
+install(TARGETS authhelper DESTINATION ${KAUTH_HELPER_INSTALL_DIR})
```
milianw, line 110: is the directory namespaced by `hotspot`? otherwise this name is far too generic, it should be called `hotspot-perf-authhelper` or similar

CMakeLists.txt
```diff
@@ -70,6 +70,8 @@ find_package(KF5 COMPONENTS
     REQUIRED
 )
 
+find_package(KF5Auth NO_MODULE)
```
milianw, line 73: `CONFIG` instead of `NO_MODULE`

README.md
```diff
@@ -113,6 +113,7 @@ As of now, you will need the following dependencies to build this project:
   - windowsystem
   - notifications
   - iconthemes
+  - kauth
```
milianw, line 116: kauth (optional)

src/CMakeLists.txt
```diff
@@ -90,6 +90,7 @@ target_link_libraries(hotspot
     KF5::WindowSystem
     KF5::Notifications
     KF5::IconThemes
+    KF5::AuthCore
```
milianw, line 93: make this optional

src/CMakeLists.txt
```diff
@@ -102,3 +103,11 @@ set_target_properties(hotspot
 
 install(TARGETS hotspot RUNTIME DESTINATION ${KDE_INSTALL_BINDIR})
 install(FILES hotspot.notifyrc DESTINATION ${KDE_INSTALL_KNOTIFY5RCDIR})
+
+add_executable(authhelper authhelper.cpp)
```
milianw, line 107: check if kauth was found

src/perfrecord.cpp
```diff
@@ -44,10 +44,11 @@
 
 #include <hotspot-config.h>
 
+#include <KAuth>
```
milianw, line 47: optional

src/perfrecord.cpp
```diff
@@ -62,72 +63,27 @@ PerfRecord::~PerfRecord()
     }
 }
 
-static QStringList sudoOptions(const QString& sudoBinary)
```
milianw, line 65: the old code needs to stay for people like me who don't pollute their `/usr` by installing stuff in there. similarly the appimage won't be able to use kauth either. afaik kauth is only an option for distro packages.

src/perfrecord.cpp
```diff
@@ -147,22 +103,27 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
...
+        QVariantMap args;
+        args[QLatin1String("script")] = elevateScript;
+        args[QLatin1String("output")] = outputFile->fileName();
+        action.setArguments(args);
+
+        KAuth::ExecuteJob* job = action.execute();
+
+        connect(job, &KAuth::ExecuteJob::percentChanged, this, [this, readTimer](KJob* job, unsigned long step) {
+            Q_UNUSED(job);
+            qDebug() << step;
```
milianw, line 118: remove

src/recordpage.cpp
```diff
@@ -356,12 +356,12 @@ RecordPage::RecordPage(QWidget* parent)
     if (m_perfRecord->currentUsername() == QLatin1String("root")) {
         ui->elevatePrivilegesCheckBox->setChecked(true);
         ui->elevatePrivilegesCheckBox->setEnabled(false);
-    } else if (m_perfRecord->sudoUtil().isEmpty()) {
-        ui->elevatePrivilegesCheckBox->setChecked(false);
-        ui->elevatePrivilegesCheckBox->setEnabled(false);
-        ui->elevatePrivilegesCheckBox->setText(
-            tr("(Note: Install pkexec, kdesudo or kdesu to temporarily elevate perf privileges.)"));
-    }
+    } /* else if (m_perfRecord->sudoUtil().isEmpty()) {
```
milianw, line 359: ?!

tests/integrationtests/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ ecm_add_test(
         KF5::CoreAddons
         KF5::WindowSystem
         KF5::KIOCore
+        KF5::AuthCore
```
milianw, line 19: optional

milianw (changes requested):

README.md
milianw: still open

src/CMakeLists.txt
```diff
@@ -102,3 +102,16 @@ set_target_properties(hotspot
 
 install(TARGETS hotspot RUNTIME DESTINATION ${KDE_INSTALL_BINDIR})
 install(FILES hotspot.notifyrc DESTINATION ${KDE_INSTALL_KNOTIFY5RCDIR})
+
+if (KF5Auth_FOUND)
+    target_link_libraries(hotspot KF5::AuthCore)
+    target_compile_definitions(hotspot PRIVATE USE_KAUTH)
```
milianw, line 108: use `hotspot-config.h.cmake` for this purpose with a new `#cmakedefine01` therein

src/authhelper.cpp
```diff
@@ -0,0 +1,36 @@
+#include <QObject>
```
milianw, line 1: missing license header

src/perfrecord.cpp
```diff
@@ -44,6 +44,10 @@
 
 #include <hotspot-config.h>
 
+#ifdef USE_KAUTH
```
milianw, line 47: `#if` (once you use `#cmakedefine01` as suggested above)

src/perfrecord.cpp
```diff
@@ -110,7 +114,7 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
 
         // I/O redirection of client scripts launched by kdesu & friends doesn't work, i.e. no data can be read...
         // so instead we use a temporary file and parse its contents via a polling timer :-/
-        auto* outputFile = new QTemporaryFile(m_elevatePrivilegesProcess);
+        auto* outputFile = new QTemporaryFile(this);
```
milianw, line 117: why this change?

src/perfrecord.cpp
```diff
@@ -128,6 +132,7 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
                         m_elevatePrivilegesProcess->deleteLater();
                     }
                 });
+
```
milianw, line 135: dito?

src/perfrecord.cpp
```diff
@@ -162,7 +167,36 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
                     }
                 });
 
+#ifdef USE_KAUTH
```
milianw, line 170: `#if` only (see above)

src/perfrecord.cpp
```diff
@@ -162,7 +167,36 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
...
+        KAuth::Action action(QLatin1String("com.kdab.hotspot.perf.elevate"));
```
milianw, line 171: here and below: I bet you are using `QL1S` wrongly, this should be `QStringLiteral` everywhere. generally, please acquaint yourself with the rules there - only use `QL1S` with overloads that support it, but don't use it when a `QString` is expected - then use `QStringLiteral` instead

src/perfrecord.cpp
```diff
@@ -162,7 +167,36 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
...
+        action.setHelperId(QLatin1String("com.kdab.hotspot.perf"));
+        QVariantMap args;
+        args[QLatin1String("script")] = elevateScript;
+        args[QLatin1String("output")] = outputFile->fileName();
+        action.setArguments(args);
+
+        KAuth::ExecuteJob* job = action.execute();
```
milianw, line 178:
```
auto job = action.execute();
```

src/authhelper.cpp
```diff
@@ -0,0 +1,36 @@
...
+public:
+    AuthHelper() = default;
+
+public slots:
+    KAuth::ActionReply elevate(const QVariantMap& args)
+    {
+        QEventLoop loop;
+        KAuth::ActionReply reply = KAuth::ActionReply::SuccessReply();
+
+        auto process = new QProcess();
```
milianw, line 21: this is leaked, construct it on the stack

src/authhelper.cpp
```diff
@@ -0,0 +1,36 @@
...
+class AuthHelper : public QObject
+{
+    Q_OBJECT
+public:
+    AuthHelper() = default;
+
+public slots:
+    KAuth::ActionReply elevate(const QVariantMap& args)
+    {
+        QEventLoop loop;
```
milianw, line 18: unused, remove

src/authhelper.cpp
```diff
@@ -0,0 +1,36 @@
...
+        KAuth::ActionReply reply = KAuth::ActionReply::SuccessReply();
```
milianw, line 19: use `return KAuth::ActionReply::SuccessReply` below directly, but also handle errors properly by returning non-success replies as needed

src/authhelper.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+        auto process = new QProcess();
+
+        QTimer::singleShot(1000, process, [process] {
```
milianw, line 23: remove, and instead use `process.waitForFinished` with a timeout of 1s, return a failure reply if that doesn't work

milianw (changes requested):

src/perfrecord.cpp
lievenhey: I can't remember correctly but it was something stupid in hindsight

src/authhelper.cpp
lievenhey: the only way this could fail during my testing was when the elevate script was not available. I added a check for this. In any other case the problem was handled in hotspot

src/perfrecord.cpp
```diff
@@ -44,6 +44,10 @@
 
 #include <hotspot-config.h>
 
+#if KF5Auth_FOUND == 1
```
milianw, line 47: no need for `== 1`

src/perfrecord.cpp
```diff
@@ -162,7 +166,39 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
                     }
                 });
 
+#if KF5Auth_FOUND == 1
```
milianw, line 169: dito

src/perfrecord.cpp
```diff
@@ -162,7 +166,39 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
...
+        KAuth::Action action(QStringLiteral("com.kdab.hotspot.perf.elevate"));
+        action.setHelperId(QStringLiteral("com.kdab.hotspot.perf"));
+        QVariantMap args;
+        args[QStringLiteral("script")] = elevateScript;
+        args[QStringLiteral("output")] = outputFile->fileName();
+        action.setArguments(args);
+
+        qDebug() << elevateScript << outputFile->fileName();
```
milianw, line 177: remove

src/perfrecord.cpp
```diff
@@ -162,7 +166,39 @@ void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfO
...
+
+        auto job = action.execute();
+
+        connect(job, &KAuth::ExecuteJob::result, this, [this, sudoBinary, options, readTimer](KJob* kjob) {
+            auto* job = qobject_cast<KAuth::ExecuteJob*>(kjob);
+
+            qDebug() << "Reply" << job->error() << job->errorString();
```
milianw, line 184: dito

src/authhelper.cpp
```diff
@@ -0,0 +1,51 @@
...
+class AuthHelper : public QObject
+{
+    Q_OBJECT
+public:
+    AuthHelper() = default;
+
+public slots:
+    KAuth::ActionReply elevate(const QVariantMap& args)
+    {
+        const QString script = args[QStringLiteral("script")].toString();
```
milianw, line 35: please don't pass the script here, instead use `Util::findLibexecBinary` or something similar to that, which works for the kauth helper

otherwise this kauth helper is super dangerous, as it could be used to run arbitrary code masquerading as our elevate script utility - one would just pass an arbitrary script as an argument and it would be executed as root...

src/recordpage.cpp
```diff
@@ -356,11 +357,11 @@ RecordPage::RecordPage(QWidget* parent)
     if (m_perfRecord->currentUsername() == QLatin1String("root")) {
         ui->elevatePrivilegesCheckBox->setChecked(true);
         ui->elevatePrivilegesCheckBox->setEnabled(false);
-    } else if (m_perfRecord->sudoUtil().isEmpty()) {
+    } else if (m_perfRecord->sudoUtil().isEmpty() && KF5Auth_FOUND == 0) {
```
milianw, line 360: use `!KF5Auth_FOUND` instead of the `== 0`

milianw (changes requested):

src/authhelper.cpp
```diff
@@ -0,0 +1,53 @@
...
+        // make install installs the helper to /usr/lib/kauth/ and the elevate script to /usr/lib/libexec
+        const QString script = qApp->applicationDirPath() + (QStringLiteral("/../libexec/elevate_perf_privileges.sh"));
+
+        if (!QFile::exists(script)) {
+            return KAuth::ActionReply::HelperErrorReply();
+        }
+
+        QProcess process;
+        process.start(script, {args[QStringLiteral("output")].toString()});
+        process.waitForFinished(1000);
```
milianw, line 45: check the return value for this call - even if you say "it cannot fail". it could be that someone edits the script, making it fail. or $whatever - no need to assume success when it's so easy to check it here!

src/authhelper.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    Q_OBJECT
+public:
+    AuthHelper() = default;
+
+public slots:
+    KAuth::ActionReply elevate(const QVariantMap& args)
+    {
+        // make install installs the helper to /usr/lib/kauth/ and the elevate script to /usr/lib/libexec
+        const QString script = qApp->applicationDirPath() + (QStringLiteral("/../libexec/elevate_perf_privileges.sh"));
```
milianw, line 37: qApp->applicationDirPath() + QLatin1String("/../libexec/elevate_perf_privileges.sh");

milianw (changes requested):

src/authhelper.cpp
lievenhey: I added a check that reports an error if the script is not stopped by waitForFinished.

milianw: did you forget to push? I can't see it?

milianw: oh now I see it - the connect above. that's far too complicated - just use the return value of `waitForFinished()!`

```
if (process.waitForFinished(1000))
    return KAuth::ActionReply::SuccessReply();
// can we include/show the `process.errorString()`?
return KAuth::ActionReply::HelperErrorReply();

**Review by milianw — CHANGES_REQUESTED** [review]
please simplify

File: src/authhelper.cpp (hunk 29 above)
Comment (lievenhey) [reply src/authhelper.cpp:55]: meh too easy

File: src/authhelper.cpp [hunk 31]
```diff
@@ -0,0 +1,58 @@
...
+        const QString script = qApp->applicationDirPath() + QLatin1String("/../libexec/elevate_perf_privileges.sh");
+
+        if (!QFile::exists(script)) {
+            return KAuth::ActionReply::HelperErrorReply();
+        }
+
+        QProcess process;
+        process.start(script, {args[QStringLiteral("output")].toString()});
+
+        if (!process.waitForFinished(1000))
```
milianw, line 46: remove the `!` - this whole branch logic here is flipped - did you even test this? :stuck_out_tongue:

src/recordpage.cpp
```diff
@@ -59,7 +59,7 @@
 #include <kio_version.h>
 
 #include <QRegularExpression>
-
+#include "hotspot-config.h"
```
milianw, line 62: add an empty line above again please

milianw (changes requested):

src/authhelper.cpp
lievenhey: The registers some signal handlers that will be executed when the script is terminated. After it elevated the privileges it will wait until it receives the kill signal. Then it will reset the privileges and terminate.
So while sounding wrong this is the correct behavior that will report an error if the script terminates before 1s is passed.
The same behavior can be found in perfrecord.cpp:121

milianw: it will report _success_ when the script terminates before 1s has passed!

and it will otherwise report an error when it finished successfully within 1s.

and if the script is supposed to be killed, we would need to launch the process in a detached manner and submit the PID as a reply from the KAuth helper, no?

lievenhey: The helper waits now for a killing signal to stop the script.

PR #303, opened 2021-06-30
use KPageDialog for settings

lievenhey:
the current settings dialog is not really expandable. By changing the
type from QDialog to QPageDialog it can be easier expanded

### DIFF
--- a/src/settingsdialog.ui
+++ b/src/settingsdialog.ui
@@ -1,14 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ui version="4.0">
  <class>SettingsDialog</class>
- <widget class="QDialog" name="SettingsDialog">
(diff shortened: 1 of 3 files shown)


src/settingsdialog.ui
```diff
@@ -84,6 +96,16 @@
      </property>
     </widget>
    </item>
+   <item row="4" column="1">
+    <widget class="KEditListWidget" name="debugPaths" native="true">
+     <property name="focusPolicy">
+      <enum>Qt::TabFocus</enum>
+     </property>
+     <property name="toolTip">
+      <string>List of that contain debug information.</string>
```
milianw, line 105: while at it, please complete this sentence:

> List of paths that contain debug information.

milianw (approved):

PR #308, opened 2021-08-09
optimize filtering

### LINKED ISSUE #30: optimize filtering
The current filtering is not as efficient as it could be. Most notably, we concatenate the symbol+binary strings. Instead we should:

- inherit from QSortFilterProxyModel
- in its `filterAcceptsRow` overload, ask the model directly whether it matches the given needle
- in the model, check against the symbol contents

Note that we cannot simply return a `Data::Symbol` variant, as that would still allocate the data on the heap for type erasure.

lievenhey:
this patch removes the filter role from AbstractTreeModel and adds a
QSortFilterProxyModel named CostProxy to remove memory allocations while
filtering

closes #30

### DIFF
--- a/src/models/costproxy.h
+++ b/src/models/costproxy.h
@@ -1,0 +1,70 @@
+/*
+  costproxy.h
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2016-2021 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Lieven Hey <lieven.hey@kdab.com>
+
+  Licensees holding valid commercial KDAB Hotspot licenses may use this file in
+  accordance with Hotspot Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef COSTPROXY_H
+#define COSTPROXY_H
+
+#include <QSortFilterProxyModel>
+
+template<typename Model>
+class CostProxy : public QSortFilterProxyModel
+{
+public:
+    explicit CostProxy(QObject* parent = nullptr)
+        : QSortFilterProxyModel(parent)
+    {
+        setRecursiveFilteringEnabled(true);
+        setDynamicSortFilter(true);
+    }
+
+protected:
+    bool filterAcceptsRow(int source_row, const QModelIndex& parent) const override
+    {
+        Q_UNUSED(source_row);
+
+        if (!parent.isValid())
+            return false;
+
+        const auto* model = qobject_cast<Model*>(sourceModel());
+        Q_ASSERT(model);
+
+        const auto* item = model->itemFromIndex(parent);
+        if (!item) {
+            return false;
+        }
+
+        const auto needle = filterRegExp().pattern();
+
+        if (item->symbol.symbol.indexOf(needle) == -1 && item->symbol.binary.indexOf(needle) == -1) {
+            return false;
+        }
+
+        return true;
+    }
+};
+
+#endif // COSTPROXY_H
--- a/src/models/treemodel.h
+++ b/src/models/treemodel.h
@@ -42,7 +42,6 @@ public:
     {
         SortRole = Qt::UserRole,
         TotalCostRole,
-        FilterRole,
         SymbolRole
     };
 };
@@ -151,10 +150,7 @@ public:
             return {};
         }
 
-        if (role == FilterRole) {
-            // TODO: optimize
-            return QString(Util::formatSymbol(item->symbol, false) + item->symbol.binary);
-        } else if (role == SymbolRole) {
+        if (role == SymbolRole) {
             return QVariant::fromValue(item->symbol);
         } else {
             auto ret = rowData(item, index.column(), role);
@@ -185,7 +181,6 @@ public:
         endResetModel();
     }
 
-private:
     const TreeNode* itemFromIndex(const QModelIndex& index) const
     {
         if (!index.isValid() || index.column() >= numColumns()) {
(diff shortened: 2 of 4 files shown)


src/models/treemodel.h
```diff
@@ -281,6 +278,46 @@ class CostTreeModel : public TreeModel<decltype(Results::root), ModelImpl>
     Results m_results;
 };
 
+template<typename Results, typename ModelImpl>
```
milianw, line 281: this should be in a separate header and not be a template. instead pass the respective SymbolRole and use that below

src/models/treemodel.h
```diff
@@ -281,6 +278,46 @@ class CostTreeModel : public TreeModel<decltype(Results::root), ModelImpl>
...
+    explicit CostProxy(QObject* parent = nullptr)
+    {
+        setRecursiveFilteringEnabled(true);
+        setDynamicSortFilter(true);
+    }
+
+protected:
+    bool filterAcceptsRow(int source_row, const QModelIndex& parent) const override
+    {
+        if (parent.isValid())
```
milianw, line 294: this filters out everything but the top level entries which sounds wrong to me?

src/models/treemodel.h
```diff
@@ -281,6 +278,46 @@ class CostTreeModel : public TreeModel<decltype(Results::root), ModelImpl>
...
+            return false;
+
+        const auto* model = static_cast<TreeModel<Results, ModelImpl>*>(sourceModel());
```
milianw, line 297: instead, use this:

```
const auto symbol = sourceModel()->index(source_row, 0, parent).data(m_symbolRole).value<Data::Symbol>();
```

src/models/treemodel.h
```diff
@@ -281,6 +278,46 @@ class CostTreeModel : public TreeModel<decltype(Results::root), ModelImpl>
...
+        const auto needle = filterRegExp().pattern();
+
+        if (item->symbol.symbol.indexOf(needle) == -1 && item->symbol.binary.indexOf(needle) == -1) {
+            return false;
+        }
+
+        return true;
+    }
+
+    bool lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const override
```
milianw, line 310: this is wrong and should be removed again - you are always sorting by symbol here, which totally breaks sorting by costs, no?

generally, sorting is unrelated to filtering, so please remove this hunk from this patch and instead explicitly take the sort role in the ctor in addition to the symbol role, but just keep the default implementation for lessThan

milianw (changes requested):

src/models/treemodel.h
lievenhey: yeah

src/models/costproxy.h
```diff
@@ -0,0 +1,43 @@
+#ifndef COSTPROXY_H
```
milianw, line 1: missing license header

src/models/costproxy.h
```diff
@@ -0,0 +1,43 @@
...
+#define COSTPROXY_H
+
+#include "treemodel.h"
+#include <QSortFilterProxyModel>
+
+template<typename Results, typename ModelImpl>
```
milianw, line 7: why pass both types and force TreeModel below (and include it above)? Can't you just take a singular `typename Model` here, and then cast to that directly? then you can remove the `treemodel.h` include, I don't think it should be needed here

src/models/costproxy.h
```diff
@@ -0,0 +1,43 @@
...
+
+protected:
+    bool filterAcceptsRow(int source_row, const QModelIndex& parent) const override
+    {
+        Q_UNUSED(source_row);
+
+        if (!parent.isValid())
+            return false;
+
+        const auto* model = static_cast<TreeModel<Results, ModelImpl>*>(sourceModel());
```
milianw, line 26:
```
const auto *model = qobject_cast<Model*>(sourceModel();
Q_ASSERT(model);
```

src/resultsbottomuppage.cpp
```diff
@@ -82,7 +82,8 @@ ResultsBottomUpPage::ResultsBottomUpPage(FilterAndZoomStack* filterStack, PerfPa
     ui->setupUi(this);
 
     auto bottomUpCostModel = new BottomUpModel(this);
-    ResultsUtil::setupTreeView(ui->bottomUpTreeView, ui->bottomUpSearch, bottomUpCostModel);
+    ResultsUtil::setupTreeView<Data::BottomUp, BottomUpModel>(ui->bottomUpTreeView, ui->bottomUpSearch,
```
milianw, line 85: by doing the change above, no change should be required here anymore and in the other similar hunks that follow

src/resultsutil.cpp
```diff
@@ -41,6 +41,7 @@
 #include "models/costdelegate.h"
 #include "models/data.h"
 #include "models/filterandzoomstack.h"
+#include "models/treemodel.h"
```
milianw, line 44: why?

src/resultsutil.h
```diff
@@ -29,7 +29,10 @@
 
 #include <functional>
 
+#include "models/costproxy.h"
+#include "models/treemodel.h"
```
milianw, line 33: why?

src/resultsutil.h
```diff
@@ -51,13 +54,23 @@ void setupHeaderView(QTreeView* view);
 
 void connectFilter(QLineEdit* filter, QSortFilterProxyModel* proxy);
 
-void setupTreeView(QTreeView* view, QLineEdit* filter, QAbstractItemModel* model, int initialSortColumn, int sortRole,
```
milianw, line 54: this should be kept as an implementation detail, and should take the proxy model as arg (but not the templated type, just a `QAbstractProxyModel*`)

src/resultsutil.h
```diff
@@ -51,13 +54,23 @@ void setupHeaderView(QTreeView* view);
...
+    view->setModel(proxy);
+    setupHeaderView(view);
+}
 
-template<typename Model>
+template<typename Results, typename ModelImpl, typename Model>
 void setupTreeView(QTreeView* view, QLineEdit* filter, Model* model)
 {
-    setupTreeView(view, filter, model, Model::InitialSortColumn, Model::SortRole, Model::FilterRole);
+    setupTreeView<Results, ModelImpl>(view, filter, model, Model::InitialSortColumn, Model::SortRole);
```
milianw, line 73: create the templated proxy here, then delegate to the implementation above (which shouldn't be inline)

all of this reduces template bloat

src/resultsutil.h
```diff
@@ -29,7 +29,10 @@
 
 #include <functional>
 
+#include "models/costproxy.h"
+#include "models/treemodel.h"
 #include <QFlags>
+#include <QTreeView>
```
milianw, line 35: remove once cleanup below has landed

milianw (changes requested):

milianw (approved):
lgtm, thanks!

PR #310, opened 2021-08-18
moved perf output in own widget

lievenhey:


lievenhey:
The code became a bit messy so I divided the messy part in two not messy widgets

src/perfoutputwidget.cpp
```diff
@@ -0,0 +1,198 @@
+#include "perfoutputwidget.h"
```
milianw, line 1: missing license header

src/perfoutputwidget.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include <QStandardPaths>
+#include <QTemporaryFile>
+#include <QTextEdit>
+#include <QVBoxLayout>
+
+#include <KParts/ReadOnlyPart>
+#include <KParts/kde_terminal_interface.h>
+#include <KService>
+
+PerfOutputWidget::PerfOutputWidget(QWidget* parent)
```
milianw, line 16: create two files for each widget please

src/perfoutputwidget.h
```diff
@@ -0,0 +1,69 @@
+#ifndef PERFOUTPUTWIDGET_H
```
milianw, line 1: missing license header

src/perfoutputwidget.h
```diff
@@ -0,0 +1,69 @@
...
+    virtual void addOutput(const QString&) = 0;
+    virtual void clear() = 0;
+    virtual void enableInput(bool enable) = 0;
+    virtual void setInputVisible(bool visible) = 0;
+
+signals:
+    void sendInput(const QByteArray& input);
+};
+
+class PerfOutputWidgetKonsole : public PerfOutputWidget
```
milianw, line 29: separate file please

src/perfoutputwidget.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace KParts {
+class ReadOnlyPart;
+}
+
+class QTemporaryFile;
+class QTextEdit;
+class QLineEdit;
+
+class PerfOutputWidget : public QWidget
```
milianw, line 14: separate file, but also don't forget to add a .cpp with the implementation of the dot `PerfOutputWidget::~PerfOutputWidget() = default`, to make sure that the vtable isn't inlined

src/recordpage.cpp
```diff
@@ -213,6 +210,17 @@ RecordPage::RecordPage(QWidget* parent)
     ui->outputFile->setMode(KFile::File | KFile::LocalOnly);
     ui->eventTypeBox->lineEdit()->setPlaceholderText(tr("perf defaults (usually cycles:Pu)"));
 
+    if (PerfOutputWidgetKonsole::isAvailable()) {
```
milianw, line 213: instead, I'd use a factory pattern that can fail, i.e. something like this:

```
m_perfOutput = PerfOutputWidgetKonsole::create(this);
if (!m_perfOutput) {
    m_perfOutput = new PerfOutputWidgetText(this);
}
```

src/perfoutputwidget.cpp
```diff
@@ -0,0 +1,198 @@
...
+        return nullptr;
+    }
+    return service->createInstance<KParts::ReadOnlyPart>();
+}
+
+PerfOutputWidgetText::PerfOutputWidgetText(QWidget* parent)
+    : PerfOutputWidget(parent)
+{
+    auto layout = new QVBoxLayout(this);
+    layout->setMargin(0);
```
milianw, line 151: ? shouldn't this be `setContentMargins(0, 0, 0, 0)` - where did you find that function? 

https://doc.qt.io/qt-5/qvboxlayout-members.html

milianw (changes requested):

src/perfoutputwidget.cpp
lievenhey: autocomplete. I remembered that you had to set one margin to 0 but not witch one. So I choose the first one

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+PerfOutputWidgetKonsole::~PerfOutputWidgetKonsole() = default;
+
+PerfOutputWidgetKonsole* PerfOutputWidgetKonsole::create(QWidget* parent)
+{
+    auto part = createPart();
+    if (!part) {
+        return nullptr;
+    }
+
+    delete part;
```
milianw, line 56: huh why is the part deleted here again? shouldn't it be passed on to the console widget?

also you should check that the part actually implements the terminal interface too

finally I guess you should only check findTail once here, and then set it as a member arg of the widget and pass it along through the ctor too?

i.e. one way or another, more of the `::clear` code below should be extracted and then called here to check that all dependencies are available before creating the widget

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+    Q_UNUSED(watched);
+    if (event->type() == QEvent::KeyPress) {
+        auto keyEvent = static_cast<QKeyEvent*>(event);
+
+        // ignore ctrl + c
+        if (keyEvent->key() == Qt::Key_C && keyEvent->modifiers() == Qt::ControlModifier) {
+            return true;
+        }
+
+        if (m_inputEnabled) {
```
milianw, line 76: hm this is going to interact horribly with TUI applications and or things that use readline internally

can't we just eat the events when input is disabled, but otherwise let the konsole handle the events itself ?

milianw (changes requested):

src/perfoutputwidgetkonsole.cpp
lievenhey: no, this would be too easy

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (m_tailPath.isEmpty()) {
+        return nullptr;
+    }
+
+    auto part = createPart();
+    if (!part) {
+        return nullptr;
+    }
+
+    const auto terminalInterface = qobject_cast<TerminalInterface*>(part);
```
milianw, line 66: why not pass that TerminalInterface directly to the PerfOutputWidgetKonsole? that would remove the need for casting below at...

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+    return service->createInstance<KParts::ReadOnlyPart>();
+}
+
+void PerfOutputWidgetKonsole::addPartToLayout()
+{
+    m_konsoleFile = new QTemporaryFile(this);
+    m_konsoleFile->open();
+
+    const auto terminalInterface = qobject_cast<TerminalInterface*>(m_konsolePart);
```
milianw, line 160: ...this place here

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+        m_inputEnabled = enable;
+    }
+}
+
+void PerfOutputWidgetKonsole::setInputVisible(bool visible)
+{
+    Q_UNUSED(visible);
+}
+
+QString PerfOutputWidgetKonsole::findTail()
```
milianw, line 138: this should be a plain free function in an anon namespace in this `.cpp` file - `private static` is a code smell to me, similar to `private slots`

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+{
+    const auto tail = QStandardPaths::findExecutable(QStringLiteral("tail"));
+
+    return tail;
+}
+
+KParts::ReadOnlyPart* PerfOutputWidgetKonsole::createPart()
```
milianw, line 145: this should also be a free function and return the `TerminalInterface*` casted result already

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    if (!service) {
+        return nullptr;
+    }
+    return service->createInstance<KParts::ReadOnlyPart>();
+}
+
+void PerfOutputWidgetKonsole::addPartToLayout()
+{
+    m_konsoleFile = new QTemporaryFile(this);
```
milianw, line 157: you could parent this to the konsole part, then you wouldn't have to manually delete it in the `::clear` method above

src/perfoutputwidgetkonsole.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+void PerfOutputWidgetKonsole::clear()
+{
+    if (m_konsolePart) {
+        m_konsolePart->deleteLater();
+    }
+
+    m_konsolePart = createPart();
+
+    if (m_konsoleFile) {
```
milianw, line 119: this wouldn't be needed if the file would be parented to the part

milianw (changes requested):
getting there ;-)

src/perfoutputwidgetkonsole.cpp
lievenhey: I still need the KParts::ReadOnlyPart to add the widget to the layout and I can't cast the TerminalInterface back to a KParts::ReadOnlyParts since it is missing the Q_OBJECT macro.
I guess adding a new member variable pointing to the same object would be the only option to remove the need for casting.

milianw: understood, I fogot the fact that this is using multiple inheritance with Q_INTERFACES

src/perfoutputwidgetkonsole.cpp
milianw: this point here could still be done - i.e. it should be a free function instead of a `private static` function

lievenhey: done

milianw:
please rebase, then I'm fine with merging this - thanks!

lievenhey:
I just noticed that it is impossible to send user input to the process. The characters you type are only added to the tail process not the perf process.

lievenhey:
I use this script to test this patch:
```
#!/usr/bin/bash
# Read the user input
echo "Enter the user name: "
read first_name
echo "The Current User Name is $first_name"
echo
echo "Enter other users'names: "
read name1 name2 name3
echo "$name1, $name2, $name3 are the other users."

printf "          "
for b in 0 1 2 3 4 5 6 7; do printf "  4${b}m "; done
echo
for f in "" 30 31 32 33 34 35 36 37; do
    for s in "" "1;"; do
        printf "%4sm" "${s}${f}"
        printf " \033[%sm%s\033[0m" "$s$f" "gYw "
        for b in 0 1 2 3 4 5 6 7; do
            printf " \033[4%s;%sm%s\033[0m" "$b" "$s$f" " gYw "
        done
        echo
     done
done
```

PR #311, opened 2021-08-19
add debuginfod config page

lievenhey:
hotspot supports debuginfod to download symbols on the fly. This change
add a config page to allow users to define their download servers.


lievenhey:
I tested this locally by creating debug symbols with
objcopy --only-keep-debug hotspot hotspot.debug
strip --strip-debug --strip-unneeded hotspot
moved hotspot.debug to /tmp/ (otherwise hotspot find it) and recorded hotspot. After making sure that some symbols are missing I started debuginfod in /tmp/ (debuginfod -F .) and added the server in the settings dialog. After reloading the symbols were available.

src/parsers/perf/perfparser.cpp
```diff
@@ -1405,7 +1407,14 @@ void PerfParser::startParseFile(const QString& path, const QString& sysroot, con
         }
 
         QProcess process;
-        process.setProcessEnvironment(Util::appImageEnvironment());
+        auto env = Util::appImageEnvironment();
+
+        if (!Settings::instance()->debuginfodUrls().isEmpty()) {
+            env.insert(QStringLiteral("DEBUGINFOD_URLS"),
```
milianw, line 1413: we shouldn't override any existing urls, but just prepend our custom urls on top, i.e.if hotspot is started with `DEBUGINFOD_URLS` already set to something

src/debuginfoddialog.ui
```diff
@@ -0,0 +1,38 @@
...
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QFormLayout" name="formLayout">
+   <item row="0" column="0">
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>Debuginfod Urls:</string>
```
milianw, line 20: here and elsewhere in this patch: please don't capitalize debuginfod, use it with a lowercase d

milianw (changes requested):
please also add a short comment documenting this behavior in the README.md and link to https://sourceware.org/elfutils/Debuginfod.html too there

src/parsers/perf/perfparser.cpp
lievenhey: done, tested by adding an invalid url in hotspot and defining a working one in DEBUGINFOD_URLS

README.md
```diff
@@ -360,6 +360,11 @@ notably, unwinding will fail when:
     Note that this can dramatically increase the size of the `perf.data` files - use it with care. Also have a look at `man perf record`.
   - For some scenarios, recursive function calls simply fail to be unwound. See also https://github.com/KDAB/hotspot/issues/93
 
+### debuginfod
+
+hotspot supports downloading debug symbols via [debuginfod](https://sourceware.org/elfutils/Debuginfod.html).
+This can be enabled by either adding download urls in the settings or launching hotspot with DEBUGINFOD_URLS defined in the environment.
```
milianw, line 366: add backticks to format the env var like this: `DEBUGINFOD_URLS`

src/parsers/perf/perfparser.cpp
```diff
@@ -1405,7 +1407,15 @@ void PerfParser::startParseFile(const QString& path, const QString& sysroot, con
         }
 
         QProcess process;
-        process.setProcessEnvironment(Util::appImageEnvironment());
+        auto env = Util::appImageEnvironment();
+
+        if (!Settings::instance()->debuginfodUrls().isEmpty()) {
```
milianw, line 1412: this is be much nicer to read when you introduce some locals:

```
const auto customUrls = Settings::instance()->debuginfodUrls();
if (!customUrls.isEmpty()) {
    const auto envVar = QStringLiteral("DEBUGINFOD_URLS");
    const auto defaultUrls = env.value(envVar);
    const auto separator = QLatin1Char(' ');
    env.insert(envVar, customurls.join(separator) + separator + defaultUrls);
}
```

milianw (changes requested):
two small nits, otherwise lgtm now - thanks!

PR #313, opened 2021-08-22
avoid calling elevate script for root

tufeigunchu:
There is no need to elevate privileges when user is root

### DIFF
--- a/src/perfrecord.cpp
+++ b/src/perfrecord.cpp
@@ -36,6 +36,7 @@
 #include <QTimer>
 
 #include <csignal>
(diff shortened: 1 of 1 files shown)


milianw:
in general it is not recommended to launch GUI applications as root.

and generally, it would be much nicer if we could just check if the user has sufficient permissions somehow. I wonder if that is doable somehow... maybe also query the capabilities set on the `perf` executable?

that said, I have nothing against merging this patch as-is, so I'll be doing that now. cheers, thanks for your contribution!

PR #314, opened 2021-08-23
apply unwind settings on startup

lievenhey:
the unwind setting only get applied when the user presses the accept
button in the settings dialog

this patch changes this to apply the last used settings on startup

this can be checked by adding a breakpoint in Mainwindow.cpp:324 (setDebugPaths)


milianw:
thanks!

PR #316, opened 2021-08-27
added custom proxies to remove allocations during filtering

lievenhey:
The caller callee Models still used the old filtering method of creating of concatenating strings and turning these into a QVariant.
This patch add some proxies that can directly filter the model without the need of creating a QVariant or string concatenation.

### DIFF
--- a/src/models/hashmodel.h
+++ b/src/models/hashmodel.h
@@ -83,6 +83,11 @@ public:
         return index(row, column);
     }
 
+    typename Rows::key_type key(int row) const
+    {
+        return m_keys.value(row);
+    }
+
 protected:
     void setRows(const Rows& rows)
     {
--- a/src/resultscallercalleepage.cpp
+++ b/src/resultscallercalleepage.cpp
@@ -38,6 +38,7 @@
 #include "resultsutil.h"
 
 #include "models/callercalleemodel.h"
+#include "models/callercalleeproxy.h"
 #include "models/costdelegate.h"
 #include "models/filterandzoomstack.h"
 #include "models/hashmodel.h"
@@ -48,7 +49,7 @@ template<typename Model>
 Model* setupModelAndProxyForView(QTreeView* view)
 {
     auto model = new Model(view);
-    auto proxy = new QSortFilterProxyModel(model);
+    auto proxy = new CallerCalleeProxy<Model>(model);
     proxy->setSourceModel(model);
     proxy->setSortRole(Model::SortRole);
     view->setModel(proxy);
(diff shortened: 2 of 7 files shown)


src/models/hashmodel.h
```diff
@@ -83,6 +83,11 @@ class HashModel : public QAbstractTableModel
         return index(row, column);
     }
 
+    typename Rows::key_type getKey(int row) const
```
milianw, line 86: rename to `key` or `keyForRow` - no `get` prefix

src/models/sourcemapproxy.h
```diff
@@ -0,0 +1,64 @@
...
+        Q_UNUSED(parent);
+
+        const auto* model = qobject_cast<Model*>(sourceModel());
+        Q_ASSERT(model);
+
+        const auto needle = filterRegExp().pattern();
+
+        const auto symbol = model->getKey(source_row);
+
+        if (symbol.indexOf(needle) == -1) {
```
milianw, line 56: I believe only this matching is different to the other proxy, I would appreciate if that could be merged

basically add two overloads and then call that here:

```
bool match(const QString &needle, const QString &symbol);
bool match(const QString &needle, const Data::Symbol &symbol);
```

the rest is all the same, no?

src/resultscallercalleepage.cpp
```diff
@@ -38,17 +38,30 @@
...
 #include "models/callercalleemodel.h"
+#include "models/callercalleproxy.h"
 #include "models/costdelegate.h"
 #include "models/filterandzoomstack.h"
 #include "models/hashmodel.h"
+#include "models/sourcemapproxy.h"
 #include "models/treemodel.h"
 
 namespace {
+QSortFilterProxyModel* getProxy(SourceMapModel* model)
```
milianw, line 49: no `get` prefix please - also this is not a getter, but a factory. rename to `createProxy`?

milianw (changes requested):

src/resultscallercalleepage.cpp
```diff
@@ -38,17 +38,24 @@
...
 
 #include "models/callercalleemodel.h"
+#include "models/callercalleeproxy.h"
 #include "models/costdelegate.h"
 #include "models/filterandzoomstack.h"
 #include "models/hashmodel.h"
 #include "models/treemodel.h"
 
 namespace {
+template<typename Model>
```
milianw, line 48: remove

src/resultscallercalleepage.cpp
```diff
@@ -38,17 +38,24 @@
...
+{
+    return new CallerCalleeProxy<Model>(model);
+}
+
 template<typename Model>
 Model* setupModelAndProxyForView(QTreeView* view)
 {
     auto model = new Model(view);
-    auto proxy = new QSortFilterProxyModel(model);
+    auto* proxy = createProxy(model);
```
milianw, line 58: inline it here, just like you did below

```
auto proxy = new CallerCalleeProxy<Model>(model);
```

milianw (changes requested):

milianw:
thanks!

PR #320, opened 2021-09-13
Detect debugge crashed

lievenhey:
detect if the debugge crashed and show a warning if so


src/perfrecord.h
```diff
@@ -68,6 +68,7 @@ class PerfRecord : public QObject
     void recordingFinished(const QString& fileLocation);
     void recordingFailed(const QString& errorMessage);
     void recordingOutput(const QString& errorMessage);
+    void debuggeCrashed();
```
milianw, line 71: two e: debugeeCrashed

milianw (changes requested):
I'm fine in general, but what happens if we stop the recording before the application has finished - is that detected as a crash (it should not)?

lievenhey:
If you stopped the recording it was detected as a crash (didn't thought of that). Fixed that now.

src/recordpage.cpp
```diff
@@ -352,6 +349,13 @@ RecordPage::RecordPage(QWidget* parent)
         ui->viewPerfRecordResultsButton->setEnabled(false);
     });
 
+    connect(m_perfRecord, &PerfRecord::debuggeeCrashed, this, [this]{
+        ui->applicationRecordWarningMessage->setText(tr("Debugge crashed. Results may be unusable."));
+        ui->applicationRecordWarningMessage->setVisible(true);
+
+        QTimer::singleShot(3000, this, [this]{ui->applicationRecordWarningMessage->setVisible(false);});
```
milianw, line 356: please don't auto hide messages. the user can hide it manually if desired. We should only clear/hide the message manually when the user re-starts a recording

milianw (changes requested):
one minor nit, otherwise lgtm

PR #321, opened 2021-09-13
Fix crash. Make sure we do not open an empty config group

aleixpol:
KConfig asserts on it and shuts hotspot down


milianw:
thanks!

PR #326, opened 2021-09-23
Support tracepoints

lievenhey:
This patch adds support for tracepoints. They will show up in the timeline like this: 
grafik


lievenhey:
In case that this should be added to the readme
How to add a custom tracepoint:

- add to source
```
#include <sys/sdt.h>
DTRACE_PROBE(hotspot, paint);```
- add to perf `sudo perf probe -x hotspot -a sdt_hotspot:paint`
- run with elevated privileges

milianw (changes requested):

milianw:
thansk!

PR #333, opened 2021-10-19
improve collapseTemplates

lievenhey:
collapseTemplates no longer breaks with operator<< and operator<

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -41,16 +41,34 @@
 #include "data.h"
 #include "settings.h"
 
-static QString collapseTemplate(const QString& str, int level)
+QString collapseTemplate(const QString& str, int level)
 {
+
     if (str.indexOf(QLatin1Char('<')) == -1) {
         return str;
     }
 
+    auto isSpaceOrAngleBracket = [](QChar c) {
+        if (c == QLatin1Char(' ')) {
+            return true;
+        }
+        if (c == QLatin1Char('<')) {
+            return true;
+        }
+        if (c == QLatin1Char('>')) {
+            return true;
+        }
+
+        return false;
+    };
+
     QString output;
     output.reserve(str.size());
+    const auto operatorKeyword = QLatin1String("operator");
+    const int size = str.size();
     int depth = 0;
-    for (auto c : str) {
+    for (int i = 0; i < size; i++) {
+        const auto c = str[i];
         if (c == QLatin1Char('<')) {
             depth++;
             if (depth == level) {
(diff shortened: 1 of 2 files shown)


lievenhey:
Now it looks for the operator keyword if < or > is found.

src/util.cpp
```diff
@@ -41,22 +41,55 @@
 #include "data.h"
 #include "settings.h"
 
-static QString collapseTemplate(const QString& str, int level)
+namespace {
+bool isOperator(const QStringRef& str, int pos)
```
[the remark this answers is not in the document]
lievenhey: thanks, I don't have much experience with text parsing

src/util.cpp
```diff
@@ -41,24 +41,51 @@
...
+        if (c == QLatin1Char('>')) {
+            return true;
+        }
+
+        return false;
+    };
+
     QString output;
     output.reserve(str.size());
+    const auto operatorKeyword = QStringLiteral("operator");
```
milianw, line 67: QLatin1String

src/util.cpp
```diff
@@ -41,24 +41,51 @@
...
             }
         } else if (c == QLatin1Char('>')) {
             depth--;
+        } else if (c == QLatin1Char('o') && str.midRef(i, operatorKeyword.size()) == operatorKeyword) {
+            i += operatorKeyword.size();
+            output.append(operatorKeyword);
+            for (; i < size && isSpaceOrAngelBracket(str[i]); i++) {
+                output.append(str[i]);
+            }
+            output.append(str[i]);
```
milianw, line 85: why is this line here, this seems to be eating one char too many? are you trying to compensate for the `i++` in the inner and outer `for` loops? maybe just do a `--i` after the last iteration of the inner loop? or use different loop vars?

src/util.cpp
```diff
@@ -41,24 +41,51 @@
...
 
-static QString collapseTemplate(const QString& str, int level)
+QString collapseTemplate(const QString& str, int level)
 {
+
     if (str.indexOf(QLatin1Char('<')) == -1) {
         return str;
     }
 
+    auto isSpaceOrAngelBracket = [](QChar c) {
```
milianw, line 51: Angle not :angel:

milianw (changes requested):
nearly done ;-)

src/util.cpp
lievenhey: I used a different loop var and added a comment explaining whats going on

lievenhey:
should be done now

milianw:
thanks!

PR #343, opened 2021-10-29
perfparser update fixing #177

### LINKED ISSUE #177: Lots of "unhandled event type 5" (and 6) (PERF_RECORD_{UN,}THROTTLE)
I just tried hotspot on a recorded perf run I just made, and I got lots of messages like:
unhandled event type 6
unhandled event type 5
unhandled event type 6

Perf command line was:
perf record -g -F max chrome --single-process foo.html

GitMensch:
fixes #177


milianw:
something is fishy here, seems like we forgot to update the `hotspot` branch previously and now lose some of @lievenhey's recent patches in perfparser with this change.

milianw:
actually no - the previous checkout was just plain wrong in hotspot - thanks @GitMensch

PR #347, opened 2021-11-10
update kpart loading

lievenhey:
someone deprecated all loading methods without adding a new one so we
need to implement our own


milianw:
no, better just wait for a better one to arise and ignore the deprecation warning until then

milianw (changes requested):
let's close this, I'd rather live with the deprecations rather than workarounding it like this

PR #348, opened 2021-11-11
add option to disable column in all tables

### LINKED ISSUE #344: filter for visible costs in tables
When one measures multiple cost metrics, or when we add more deduced cost metrics, the tables become extremely cramped. It would be good if we could add filtering capabilities to all data tables.

A context menu on all tables to allow configuration of visible cost metrics. See existing context menu support in the CostHeaderView, which goes into that direction, but doesn't go far enough yet:

We should sync this across views. I.e. it should be global to some degree. Let's say I'm doing off-cpu profiling, then usually I don't want to see the number of context switches - the deduced off-CPU time is good enough.

We should also show this context menu anywhere in the tables, not just on the header. And then we should add this to the filter menu, to ensure people are aware that some data is not visible.

lievenhey:
this patch adds a shared context menu that disables columns on all
tables.
The context menu is automatically connected if setupContextMenu is
called

closed #344

### DIFF
--- a/src/costcontextmenu.cpp
+++ b/src/costcontextmenu.cpp
@@ -1,0 +1,67 @@
+/*
+  costcontextmenu.cpp
+
+  This file is part of Hotspot, the Qt GUI for performance analysis.
+
+  Copyright (C) 2021 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+  Author: Lieven Hey <lieven.hey@kdab.com>
+
+  Licensees holding valid commercial KDAB Hotspot licenses may use this file in
+  accordance with Hotspot Commercial License Agreement provided with the Software.
+
+  Contact info@kdab.com if any conditions of this licensing are not clear to you.
+
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "costcontextmenu.h"
+
+#include <QHeaderView>
+#include <QMenu>
+#include <QTreeView>
+
+CostContextMenu::CostContextMenu(QObject* parent)
+    : QObject(parent)
+{
+}
+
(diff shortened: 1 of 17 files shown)


lievenhey:
filter all tables showcase

src/costcontextmenu.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+#include "costcontextmenu.h"
+
+#include <QTreeView>
+#include <QHeaderView>
+#include <QMenu>
+
+CostContextMenu *CostContextMenu::instance()
+{
+    static CostContextMenu menu;
```
milianw, line 36: static auto menu = QPointer<CostContextMenu>(new CostContextMenu(QCoreApplication::instance());

to ensure this gets reaped at shutdown

better yet: don't use singletons!

src/costcontextmenu.cpp
```diff
@@ -0,0 +1,92 @@
...
+            } else {
+                m_hiddenColumns.insert(name);
+            }
+
+            emit hiddenColumnsChanged();
+        });
+    }
+}
+
+void CostContextMenu::hideColumns(QTreeView *view)
```
milianw, line 85: why is this now using the QTreeView to hide columns instead of the CostHeaderView's setSectionHidden like before?

milianw (changes requested):

src/costcontextmenu.cpp
lievenhey: because I first added the feature to the view and then connected the header

src/costcontextmenu.cpp
lievenhey: Is there a alternative to the singleton in this case? The only alternative I found was creating the menu in MainWindow and then passing a reference to all Result*Page and then to the views.

milianw: yes exactly, pass things around. I mean hotspot is supposed to support multiple main windows but just testing it now I realize this totally broke when I migrated to KDDW :D will have to fix that.

but yes, please create it centrally and then pass it along, otherwise multiple main windows would be synced which is imo not really desired, is it?

lievenhey: ok, I added it to ResultsPage

src/costcontextmenu.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <QMenu>
+#include <QPointer>
+#include <QTreeView>
+
+CostContextMenu::CostContextMenu(QObject* parent)
+    : QObject(parent)
+{
+}
+
+void CostContextMenu::addToMenu(QTreeView* view, QMenu* menu)
```
milianw, line 41: I don't get the duplication, why do we need this and the one taking the header view? isn't it enough to always simply use the header view? or the view itself?

milianw:
thanks

Issue #340, opened 2021-10-27
unhandled asm line format


GitMensch:
https://github.com/KDAB/hotspot/blob/df39e783a997a7495ba654853720395b3aa0b238/src/models/disassemblyoutput.cpp#L49-54 has an explicit check for that. I'm not sure if this is a bug and something should be done or if this is just an ignorable warning (in which case this info _may_ be added to the message):

```
unhandled asm line format: "/root/.debug/usr/lib64/libc-2.17.so/5861cfac90a2037fb2f649fdaf1b679b2a25f1e7/elf"
unhandled asm line format: "00000000000897c0 <__GI_memset>"
unhandled asm line format: "/root/.debug/usr/lib64/libgmp.so.10.2.0/b8fee69329c044ac323d5e213cc747937d5228db/elf"
unhandled asm line format: "0000000000032060 <__gmpn_sizeinbase+0x70>"
```

lievenhey:
I believe that it is too ample, since we don't want to remove lines like `11d9:  74 0d   je 11e8 <_start@@Base+0xc8>`

PR #349, opened 2021-11-17
skip uninteresting lines in disassembly output

lievenhey:
fixes: #340

### DIFF
--- a/src/models/disassemblyoutput.cpp
+++ b/src/models/disassemblyoutput.cpp
@@ -25,12 +25,13 @@
   along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#include "data.h"
 #include "disassemblyoutput.h"
-#include <QStandardPaths>
+#include "data.h"
 #include <QApplication>
-#include <QProcess>
 #include <QDebug>
+#include <QProcess>
+#include <QRegularExpression>
+#include <QStandardPaths>
 
 static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray &output)
 {
@@ -38,19 +39,35 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
 
     QTextStream stream(output);
     QString asmLine;
+    // detect lines like:
+    // 4f616: 84 c0 test %al,%al
+    static const QRegularExpression disassemblyRegex(QStringLiteral("^    ([0-9a-f]{4,}):\t[0-9a-f ]{2,}"));
     while (stream.readLineInto(&asmLine))
     {
         if (asmLine.isEmpty() || asmLine.startsWith(QLatin1String("Disassembly")))
             continue;
 
-        const auto asmTokens = asmLine.splitRef(QLatin1Char(':'));
-        const auto addrLine = asmTokens.value(0).trimmed();
-
-        bool ok = false;
-        const auto addr = addrLine.toULongLong(&ok, 16);
-        if (!ok) {
-            qWarning() << "unhandled asm line format:" << addrLine;
+        // skip lines like these: 0000000000001265 <main>:
+        const int colonIndex = asmLine.indexOf(QLatin1Char(':'));
+        const int angleBracketIndex = asmLine.indexOf(QLatin1Char('<'));
+        if (angleBracketIndex > 0 && colonIndex > angleBracketIndex) {
             continue;
+        }
+
+        // we don't care about the file name
+        if (asmLine.startsWith(QLatin1Char('/')) && asmLine.contains(QStringLiteral("file format"))) {
+            continue;
+        }
+
+        quint64 addr = 0;
+        const auto match = disassemblyRegex.match(asmLine);
+        if (match.hasMatch()) {
+            bool ok = false;
+            addr = match.capturedRef(1).toULongLong(&ok, 16);
+            if (!ok) {
+                qWarning() << "unhandled asm line format:" << asmLine;
+                continue;
+            }
         }
 
         disassemblyLines.push_back({addr, asmLine});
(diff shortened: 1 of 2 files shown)


src/models/disassemblyoutput.cpp
```diff
@@ -43,15 +43,24 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
         if (asmLine.isEmpty() || asmLine.startsWith(QLatin1String("Disassembly")))
             continue;
 
+        const int colonIndex = asmLine.indexOf(QLatin1Char(':'));
+        const int angleBracketIndex = asmLine.indexOf(QLatin1Char('<'));
+
+        // skip function header line as it is not interesting for us
```
milianw, line 49: can you give an example for such a line?
doesn't this break for symbol names like

Foo::Bar<Asdf>()

src/models/disassemblyoutput.cpp
```diff
@@ -43,15 +43,24 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
...
+        if (angleBracketIndex > 0 && colonIndex > angleBracketIndex) {
+            continue;
+        }
+
+        // we don't care about the file name
+        if (asmLine.startsWith(QLatin1Char('/')) && asmLine.indexOf(QStringLiteral("file format")) > 0) {
+            continue;
+        }
```
milianw, line 57: use `contains()` instead of `indexOf() > 0`

src/models/disassemblyoutput.cpp
```diff
@@ -43,15 +43,24 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
...
+
         const auto asmTokens = asmLine.splitRef(QLatin1Char(':'));
         const auto addrLine = asmTokens.value(0).trimmed();
 
         bool ok = false;
         const auto addr = addrLine.toULongLong(&ok, 16);
-        if (!ok) {
```
milianw, line 51: this should still be included, there may be other unexpected lines in the future after all

milianw (changes requested):

src/models/disassemblyoutput.cpp
lievenhey: the lines look like this `0000000000001090 <main>:` of `0000000000001264 <_ZN4Test4testEv>:`
the names are mangled so it not a problem

src/models/disassemblyoutput.cpp
lievenhey: in that case I would suppose to add a regex like this `^   [0-9a-f]{4,}:       [0-9a-f ]{2,}` to verify that we indeed have a disassembled line here to before trying this since source code lines would break this

milianw: ideally we would extract the source code lines and show them too

src/models/disassemblyoutput.cpp
```diff
@@ -25,34 +25,52 @@
...
+#include <QRegularExpression>
+#include <QStandardPaths>
 
 static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray &output)
 {
     QVector<DisassemblyOutput::DisassemblyLine> disassemblyLines;
 
     QTextStream stream(output);
     QString asmLine;
+    const QRegularExpression disassemblyRegex(QStringLiteral("^   [0-9a-f]{4,}:       [0-9a-f ]{2,}"));
```
milianw, line 42: `static`, and aren't you missing `<>` before the `:`?

lievenhey: this is only to check if the line is an assembly line like this: `   4f616:       84 c0                   test   %al,%al` and not for the function header part.

milianw: still, the regex should be static, and ideally you add what you wrote above as a comment:

```
// detect lines like:
// 4f616: 84 c0 test %al,%al
static const disassemblyRegex = QRegularExpression(QStringLiteral(...));
```

src/models/disassemblyoutput.cpp
```diff
@@ -25,34 +25,52 @@
...
 
+        // we don't care about the file name
+        if (asmLine.startsWith(QLatin1Char('/')) && asmLine.contains(QStringLiteral("file format"))) {
+            continue;
+        }
+
+        quint64 addr = 0;
+        if (disassemblyRegex.match(asmLine).hasMatch()) {
+            const auto asmTokens = asmLine.splitRef(QLatin1Char(':'));
+            const auto addrLine = asmTokens.value(0).trimmed();
```
milianw, line 64: considering you are already doing an expensive regex match above, please at the very least leverage it properly!

add a capture group for the address part and then use `match.capturedRef(1).toULongLong`

milianw (changes requested):

src/models/disassemblyoutput.cpp
```diff
@@ -56,14 +60,15 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
...
+        quint64 addr = 0;
+        const auto match = disassemblyRegex.match(asmLine);
+        if (match.hasMatch()) {
+            bool ok = false;
+            addr = match.capturedRef(1).toULongLong(&ok, 16);
+            if (!ok) {
+                qWarning() << "unhandled asm line format:" << asmLine;
+                continue;
+            }
         }
```
milianw, line 72: and if there's no match, what then? is it fine to add the line with `addr=0` - is that already handled elsewhere correctly, i.e. we don't want to show `0` for all these lines in the view?

src/models/disassemblyoutput.cpp
```diff
@@ -43,6 +43,19 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
         if (asmLine.isEmpty() || asmLine.startsWith(QLatin1String("Disassembly")))
             continue;
 
+        const int colonIndex = asmLine.indexOf(QLatin1Char(':'));
```
milianw, line 50: please give an example for a line that this is supposed to skip

src/models/disassemblyoutput.cpp
lievenhey: I added a check to the disassembly model to skip the cost calculation if addr is 0. This should be fine since normally a process can't access that area.

milianw:
thanks

PR #350, opened 2021-11-17
fix crash in caller callee view

lievenhey:
the caller callee proxy broke to source navigation
this patch fixes that

### DIFF
--- a/src/resultscallercalleepage.cpp
+++ b/src/resultscallercalleepage.cpp
@@ -178,7 +178,6 @@ ResultsCallerCalleePage::toSourceMapLocation(const QString& location, const Data
 
 ResultsCallerCalleePage::SourceMapLocation ResultsCallerCalleePage::toSourceMapLocation(const QModelIndex& index) const
 {
-    Q_ASSERT(qobject_cast<const SourceMapModel*>(index.model()));
     const auto location = index.data(SourceMapModel::LocationRole).toString();
     const auto symbol =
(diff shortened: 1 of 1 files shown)


src/resultscallercalleepage.cpp
```diff
@@ -178,7 +178,8 @@ ResultsCallerCalleePage::toSourceMapLocation(const QString& location, const Data
 
 ResultsCallerCalleePage::SourceMapLocation ResultsCallerCalleePage::toSourceMapLocation(const QModelIndex& index) const
 {
-    Q_ASSERT(qobject_cast<const SourceMapModel*>(index.model()));
+    const auto proxy = qobject_cast<const QSortFilterProxyModel*>(index.model());
```
milianw, line 181: the `proxy` is only used for the assertion, but nowhere else. I'd say we could/should remove the assert completely.

alternatively, make this:

```
Q_ASSERT([]() {
    const auto proxy = q_object_cast<const QAbstractProxyModel*>(index.model());
    return proxy && qobject_cast<const SourceMapModel*>(proxy->sourceModel());
}());
```

milianw (changes requested):

src/resultscallercalleepage.cpp
lievenhey: I tried the lambda but it didn't work (could not capture index) so I removed the assertion

milianw (approved):

milianw:
thanks

PR #352, opened 2021-11-18
add github actions for automatic testing

lievenhey:


lievenhey:
The one that checks for valid build instructions seems to work.

Still need to test the auto build appimage test (that in theory should be able to upload the appimage somewhere)

.github/workflows/build-appimage.yml
```diff
@@ -0,0 +1,24 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: "**"
+ 
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
milianw, line 13: an app image should be using the oldest, not the youngest (i.e. latest) - or is latest here the oldest?

milianw (changes requested):
generally fine with me, but can you try to rebase - is the ubuntu error resolved by your other fix that I just merged now?

.github/workflows/build-appimage.yml
lievenhey: this is just the image that runs the docker container that generates the appimage

milianw: ah, ok

lievenhey:
It should (written before tests are completed) but sadly the appimage generation fails quietly.

lievenhey:
I removed the appimage build script (it didn't work and I don't really want to debug something with a runtime of 8h+)

milianw:
the -actual files should not be added to git, only the -expected.

then, please keep the -expected files as-is and handle it in the test: i.e. if the first comparison fails, do a search&replace on the expected file and then try again. otherwise the tests will now fail on older/other platforms that still produce the data of the previous -expected files

lievenhey:
This isn't a problem. There are now two expected files. One with jumpq and one without. If the disassembly contains a jumpq instruction the test automatically opens the jumpq.expected file instead of the other one.

milianw:
wouldn't it be much nicer if we would have this as a reusable function that does this search&replace automatically? then we wouldn't have to pollute the git history with duplicate test data. and we could easily deploy it to all other data files we may have in the future?

milianw:
lgtm, thank you!

PR #353, opened 2021-11-19
fix compile error on ubuntu

lievenhey:
ubuntu still uses kf5 5.68.0 which does not has the percent signal


milianw (approved):
lgtm, thanks

PR #354, opened 2021-11-22
set kf5auth minimum version to 5.55

lievenhey:
kf5auth 5.55 introduced AuthCore to remove gui stuff from helpers but
the appimage generation still uses 5.33 which does not support this


lievenhey:
KParts is also broken
I think it is easier to just fork kf5-precompiled and build a new version.

milianw:
yes, agreed - or just not use it at all but rely on craft instead

PR #355, opened 2021-11-30
fix compile error on ubuntu

lievenhey:
follow up fix for the existing one since I am to stupid to copy a line from my Ubuntu VM to my host system.


milianw:
conflicts, but otherwise lgtm

milianw:
thanks

Issue #336, opened 2021-10-25
When will we have `annotate`?


fzyzcjy:
Hi thanks for the lib! I wonder when will have the `perf annoate` GUI? It is very useful!

GitMensch:
The code written by @lievenhey referenced here was included by @milianw as part of #349 - is this considered to be fixed now, too?

If yes README.md may need an adjustment again, still saying:

>  But a proper annotation view like perf annotate on the source level, is currently missing

In any case I'm eager to run the new CI build with those changes!

PR #356, opened 2021-12-08
update readme to announce annotate view

lievenhey:
fixes: #336


milianw:
I don't think that we have a proper source annotation just yet, personally. it's getting there, but is still miles away from what I would call "proper" :D

milianw:
have a look at vtune's source annotation to understand what I mean by "proper". also look at perf's annotation view capabilities wrt following jumps etc.

src/models/disassemblymodel.cpp
```diff
@@ -47,6 +61,14 @@ QVariant DisassemblyModel::headerData(int section, Qt::Orientation orientation,
     if (section == DisassemblyColumn)
         return tr("Assembly");
 
+    if (section == LinkedFunctionName) {
+        return QStringLiteral("Linked Function Name");
```
milianw, line 65: tr

src/models/disassemblymodel.cpp
```diff
@@ -47,6 +61,14 @@ QVariant DisassemblyModel::headerData(int section, Qt::Orientation orientation,
...
+    }
+
+    if (section == LinkedFunctionOffset) {
+        return QStringLiteral("Linked Function Offset");
```
milianw, line 69: tr

src/models/disassemblyoutput.cpp
```diff
@@ -22,9 +22,10 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
     QString asmLine;
     // detect lines like:
     // 4f616: 84 c0 test %al,%al
-    static const QRegularExpression disassemblyRegex(QStringLiteral("^    ([0-9a-f]{4,}):\t[0-9a-f ]{2,}"));
+    static const QRegularExpression disassemblyRegex(QStringLiteral("^[ ]+([0-9a-f]{4,}):\t[0-9a-f ]{2,}"));
     while (stream.readLineInto(&asmLine))
     {
+        qDebug() << asmLine;
```
milianw, line 28: remove

src/models/disassemblyoutput.h
```diff
@@ -17,6 +17,8 @@ struct DisassemblyOutput
     {
         quint64 addr = 0;
         QString disassembly;
+        QString linkedFunctionName;
+        int linkedFunctionOffset;
```
milianw, line 21: = 0;

also document what this offset means, it's probably the offset into the `disassembly` string, or?

src/models/disassemblyoutput.cpp
```diff
@@ -49,9 +53,31 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
                 qWarning() << "unhandled asm line format:" << asmLine;
                 continue;
             }
+
+            const auto leftBracketIndex = asmLine.indexOf(QLatin1Char('<'));
```
milianw, line 57: extract this code out into a separate function then call it below, i..e something like

```
struct LinkedFunction
{
    QString name;
    int offset = 0;
};

LinkedFunction extractLinkedFunction(const QString &disassembly);
...
disassemblyLines.push_back({addr, asmLine, extractLinkedFunction(asmLine)});
```

src/models/disassemblyoutput.cpp
```diff
@@ -40,6 +41,9 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
             continue;
         }
 
+        QString linkedFunctionName;
```
milianw, line 44: remove (see below)

src/resultspage.cpp
```diff
@@ -104,10 +104,10 @@ ResultsPage::ResultsPage(PerfParser* parser, QWidget* parent)
     m_summaryPageDock->addDockWidgetAsTab(m_disassemblyDock, KDDockWidgets::InitialVisibilityOption::StartHidden);
     m_disassemblyDock->toggleAction()->setEnabled(false);
     m_summaryPageDock->setAsCurrentTab();
-    if (m_frequencyPage) {
-        m_frequencyDock = dockify(m_frequencyPage, QStringLiteral("frequency"), tr("Fr&equency"), tr("Ctrl+E"));
-        m_summaryPageDock->addDockWidgetAsTab(m_frequencyDock);
-    }
+#if QCustomPlot_FOUND
```
milianw, line 107: why this change?

milianw (changes requested):

src/models/disassemblyoutput.h
lievenhey: it is the offset from the start of the function

src/resultspage.cpp
lievenhey: `frequencypage.h` is only included when `QCustomPlot` is found. If it is not found, the compiler doesn't know that `m_frequencyPage` is a QWidget and fails to compile.
This Change is also in #360

milianw: ok but that has nothing to do with the topic of "update readme to announce annotate view" and should thus not be part of this pull request

Issue #12, opened 2016-12-23
allow launching perf on remote machines


milianw:
Extend the feature in #11  to allow running perf on a remote machine via ssh, then stream the data to localhost

tmcguire:
It would be nice if hotspot can help with the setup here, which might not be easy for novice users. Especially setting up debug symbols, which might only be available on the host's sysroot.

jonhoo:
Presumably another thing that would be needed here would be to supply a symbol file other than the one named in `perf.data`, as the binary may reside at a different location locally.

milianw:
@jonhoo not sure, but I think what you mention is implemented already. Have a look at the command line arguments, cf.:

```
$ hotspot --help
Usage: hotspot [options] [files...]
Linux perf GUI for performance analysis.

Options:
  -h, --help               Displays this help.
  -v, --version            Displays version information.
  --sysroot <path>         Path to sysroot which is used to find libraries.
  --kallsyms <path>        Path to kallsyms file which is used to resolve
                           kernel symbols.
  --debugPaths <paths>     Colon separated list of paths that contain debug
                           information.
  --extraLibPaths <paths>  Colon separated list of extra paths to find
                           libraries.
  --appPath <path>         Path to folder containing the application executable
                           and libraries.
  --arch <path>            Architecture to use for unwinding.

Arguments:
  files                    Optional input files to open on startup, i.e.
                           perf.data files
```

jonhoo:
Ah, yes, you're right. My bad for not digging a bit further.

PR #359, opened 2022-01-18
SSH

lievenhey:
These patches allow hotspot to run `perf record` on remote devices via ssh. The recording data is streamed directly to hotspot without storing it on the device.


GitMensch:
Thank you for this nice addition, it _seems_ that README.md may need an update (or is this actually a draft implementation so "too early for that")?

src/settings.cpp
```diff
@@ -132,3 +132,9 @@ void Settings::setCallgraphColors(const QColor& active, const QColor& inactive)
         emit callgraphChanged();
     }
 }
+
+void Settings::setSshaskpassPath(const QString& sshaskpath)
```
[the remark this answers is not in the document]
milianw: why is this a setting? users should just set the `SSH_ASKPASS` env var, no? I am not aware of any other tool where I'd graphically configure the ssh ask pass binary?

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
                                      callgraphPage->functionColor->color().name());
     });
 }
+
+void SettingsDialog::addSshPage()
+{
+    const auto sshPath = QStandardPaths::findExecutable(QLatin1String("ssh"));
```
milianw, line 303: QStringLiteral

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+    auto item = addPage(page, tr("SSH"));
+    item->setHeader(tr("SSH Settings"));
+    item->setIcon(QIcon::fromTheme(QStringLiteral("preferences-system-windows-behavior")));
+
+    sshPage->setupUi(page);
+
+    auto deviceConfig = config()->group("devices");
+
+    auto saveFunction = [this](KConfigGroup group) {
+        group.writeEntry("hostname", sshPage->ipLineEdit->text());
```
milianw, line 320: rename ipLineEdit to hostnameLineEdit to clarify what it is

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+    sshPage->sshaskpassLineEdit->setText(settings->sshaskpassPath());
+
+    connect(settings, &Settings::sshaskpassChanged, this,
+            [this](const QString& path) { sshPage->sshaskpassLineEdit->setText(path); });
+
+    connect(sshPage->testButton, &QPushButton::pressed, this, [this, devices] {
+        auto ssh = new QProcess(this);
+        ssh->setProgram(QStandardPaths::findExecutable(QLatin1String("ssh")));
+
+        const auto hostname =
```
milianw, line 357: it should be allowed to not specify a user, in which case ssh will use the default internally (either current or whatever is provided in `~/.ssh/.config)`. but I believe connecting to `@localhost` e.g. is going to fail, the `@` prefix must not be added in such a case

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+            QStringLiteral("%1@%2").arg(sshPage->usernameLineEdit->text(), sshPage->ipLineEdit->text());
+        QStringList arguments = {hostname};
+        const auto sshOptions = sshPage->sshOptionsLineEdit->text();
+        if (!sshOptions.isEmpty()) {
+            arguments.append(sshOptions.split(QLatin1Char(' ')));
```
milianw, line 362: this is bound to fail, use `KShell::splitArgs`

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+        }
+
+        sshCopyId->setArguments(arguments);
+        sshCopyId->setProcessEnvironment(sshEnvironment());
+        sshCopyId->start();
```
milianw, line 407: better safe than sorry: connect before start

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+    connect(sshPage->copyKeyButton, &QPushButton::pressed, this, [this, devices] {
+        auto sshCopyId = new QProcess(this);
+        sshCopyId->setProgram(QStandardPaths::findExecutable(QLatin1String("ssh-copy-id")));
+
+        const auto hostname =
+            QStringLiteral("%1@%2").arg(sshPage->usernameLineEdit->text(), sshPage->ipLineEdit->text());
+        QStringList arguments = {hostname};
+        const auto sshOptions = sshPage->sshOptionsLineEdit->text();
+        if (!sshOptions.isEmpty()) {
+            arguments.append(sshOptions.split(QLatin1Char(' ')));
```
milianw, line 402: dito `KShell::splitArgs`

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+        }
+        arguments.append(QLatin1String("perf"));
+
+        ssh->setArguments(arguments);
+        ssh->setProcessEnvironment(sshEnvironment());
+        ssh->start();
```
milianw, line 368: better safe than sorry: connect before start

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+        ssh->setProgram(QStandardPaths::findExecutable(QLatin1String("ssh")));
+
+        const auto hostname =
+            QStringLiteral("%1@%2").arg(sshPage->usernameLineEdit->text(), sshPage->ipLineEdit->text());
+        QStringList arguments = {hostname};
+        const auto sshOptions = sshPage->sshOptionsLineEdit->text();
+        if (!sshOptions.isEmpty()) {
+            arguments.append(sshOptions.split(QLatin1Char(' ')));
+        }
+        arguments.append(QLatin1String("perf"));
```
milianw, line 364: QStringLiteral

src/settingsdialog.cpp
```diff
@@ -290,3 +297,128 @@ void SettingsDialog::addCallgraphPage()
...
+                        devices->updateCurrentConfig();
+                    } else {
+                        showError(tr("Error: %1").arg(QString::fromUtf8(ssh->readAllStandardError())));
+                    }
+                });
+    });
+
+    connect(sshPage->copyKeyButton, &QPushButton::pressed, this, [this, devices] {
+        auto sshCopyId = new QProcess(this);
+        sshCopyId->setProgram(QStandardPaths::findExecutable(QLatin1String("ssh-copy-id")));
```
milianw, line 395: QStringLiteral

src/ssh.cpp
```diff
@@ -0,0 +1,64 @@
...
+    arguments.append(command);
+
+    return arguments;
+}
+
+QProcessEnvironment sshEnvironment()
+{
+    auto env = Util::appImageEnvironment();
+    Settings* settings = Settings::instance();
+    env.insert(QLatin1String("SSH_ASKPASS"), settings->sshaskpassPath());
```
milianw, line 38: QStringLiteral

and what about people like me who actually set this env var already but haven't configured it in hotspot? this is all backwards

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,245 @@
...
+    }
+}
+
+void PerfRecordSSH::sendInput(const QByteArray& input)
+{
+    if (m_recordProcess)
+        m_recordProcess->write(input);
+}
+
+QString PerfRecordSSH::currentUsername()
```
milianw, line 89: this is bogus, for ssh we provide the username externally (or not at all). and we don't allow runtime attaching via ssh anyways (or?). remove this for now

src/ssh.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+QString sshOutput(const QString& deviceName, const QStringList& command)
+{
+    QProcess ssh;
+    ssh.setProgram(QStandardPaths::findExecutable(QLatin1String("ssh")));
+    const auto arguments = assembleSSHArguments(deviceName, command);
+    ssh.setArguments(arguments);
+    ssh.setProcessEnvironment(sshEnvironment());
+    ssh.start();
+    ssh.waitForFinished();
```
milianw, line 50: on a flaky network connection, this will freeze the whole application which is not nice at all. can we find a way to make this async? or at least add timeouts to prevent us from waiting ad infinitum when we drive on a train?

src/ssh.cpp
```diff
@@ -0,0 +1,64 @@
...
+}
+
+int sshExitCode(const QString& deviceName, const QStringList& command)
+{
+    QProcess ssh;
+    ssh.setProgram(QStandardPaths::findExecutable(QLatin1String("ssh")));
+    const auto arguments = assembleSSHArguments(deviceName, command);
+    ssh.setArguments(arguments);
+    ssh.setProcessEnvironment(sshEnvironment());
+    ssh.start();
```
milianw, line 61: dito

src/ssh.cpp
```diff
@@ -0,0 +1,64 @@
...
+    ssh.waitForFinished();
+    return QString::fromUtf8(ssh.readAll());
+}
+
+int sshExitCode(const QString& deviceName, const QStringList& command)
+{
+    QProcess ssh;
+    ssh.setProgram(QStandardPaths::findExecutable(QLatin1String("ssh")));
```
milianw, line 57: dito, also can we share some code here please?

milianw (changes requested):

src/ssh.cpp
lievenhey: I tried making it async, but I didn't find a simple way for this, so I moved the hole thing in another thread (recordpage.cpp:436).

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+bool PerfRecordSSH::canUseAio()
+{
+    // somehow this doesn't work with ssh
+    return false;
+}
+
+bool PerfRecordSSH::canCompress()
+{
+    if (m_deviceName.isEmpty())
```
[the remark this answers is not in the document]
lievenhey: I can run `perf record --aio=0 --compression-level=22` to create a compressed `perf.data` file, so I guess it doesn't relies on it

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,245 @@
...
+bool PerfRecordSSH::canSwitchEvents()
+{
+    if (m_deviceName.isEmpty())
+        return false;
+    return perfRecordHelp(m_deviceName).contains(QLatin1String("--switch-events"));
+}
+
+bool PerfRecordSSH::canUseAio()
+{
+    // somehow this doesn't work with ssh
```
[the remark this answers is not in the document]
lievenhey: perf reports `failed to write perf data, error: Illegal seek` when running with `--aio`.
You can try this with `ssh $USERNAME@localhost "perf record --aio=1 -o - ls" > perf.data`

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,245 @@
...
+    if (m_deviceName.isEmpty())
+        return false;
+    return Zstd_FOUND && perfBuildOptions(m_deviceName).contains(QLatin1String("zstd: [ on  ]"));
+}
+
+bool PerfRecordSSH::isPerfInstalled()
+{
+    if (m_deviceName.isEmpty())
+        return false;
+    return sshExitCode(m_deviceName, {QLatin1String("perf")}) != 127;
```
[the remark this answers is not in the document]
lievenhey: test is part of POSIX but often realized as a buildin. On my system command is realized as a builtin, so I would rather use test

src/settings.cpp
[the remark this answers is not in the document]
lievenhey: It is there, if the user doesn't has set it by default and is too lazy / stupid to set if manually or want a special version

src/settingsdialog.cpp
lievenhey: `const auto sshPath = QStandardPaths::findExecutable(QStringLiteral("ssh"));` or `const auto sshPath = QStringLiteral("ssh"));`

src/ssh.cpp
lievenhey: I added a check to only set it, if `settings->sshAskPassPath()` has a value.

src/ssh.cpp
```diff
@@ -0,0 +1,64 @@
...
+    auto env = Util::appImageEnvironment();
+    Settings* settings = Settings::instance();
+    env.insert(QLatin1String("SSH_ASKPASS"), settings->sshaskpassPath());
+    return env;
+}
+
+QString sshOutput(const QString& deviceName, const QStringList& command)
+{
+    QProcess ssh;
+    ssh.setProgram(QStandardPaths::findExecutable(QLatin1String("ssh")));
```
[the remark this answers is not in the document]
milianw: corerction: This should be `...::findExecutable(QStringLiteral(""))`

src/settings.cpp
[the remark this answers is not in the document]
milianw: ok, but then make sure to not overwrite the env var when it's already set

src/perfrecordssh.cpp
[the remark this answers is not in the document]
milianw: TIL

src/perfrecordssh.cpp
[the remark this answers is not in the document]
milianw: then put that into the comment instead

// perf reports "error: Illegal seek" when trying to use aio and streaming data

src/settingsdialog.cpp
milianw: the former

src/perfrecordssh.cpp
lievenhey: can't do, since recordpage.cpp uses this to detect if off-cpu sampling is possible. But since off-cpu sampling is currently not possible via ssh I return an empty string.

lievenhey:
weird, somehow the ui on the mainwindow got changed, but I didn't touch that file.

GitMensch:
@milianw I'm wondering where this has gone, the requested changes seem to have been done.

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (m_deviceName.isEmpty())
+        return false;
+    return canTrace(QStringLiteral("events/sched/sched_switch"));
+}
+
+QString perfRecordHelp(const QString& hostname)
+{
+    const QString recordHelp = [hostname]() {
+        static QString help =
+            sshOutput(hostname, {QLatin1String("perf"), QLatin1String("record"), QLatin1String("--help")});
```
milianw, line 101: here and below: QStringLiteral

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+        if (help.isEmpty()) {
+            // no man page installed, assume the best
+            help = QLatin1String("--sample-cpu --switch-events");
```
milianw, line 104: dito

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+        return help;
+    }();
+    return recordHelp;
+}
+
+QString perfBuildOptions(const QString& hostname)
+{
+    const QString buildOptionsHelper = [hostname]() {
+        static QString buildOptions =
+            sshOutput(hostname, {QLatin1String("perf"), QLatin1String("version"), QLatin1String("--build-options")});
```
milianw, line 115: dito

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+{
+    if (m_deviceName.isEmpty())
+        return false;
+    return canTrace(QStringLiteral("events/sched/sched_switch"));
+}
+
+QString perfRecordHelp(const QString& hostname)
+{
+    const QString recordHelp = [hostname]() {
+        static QString help =
```
milianw, line 100: the output must not be static for varying hostnames

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+        }
+        return help;
+    }();
+    return recordHelp;
+}
+
+QString perfBuildOptions(const QString& hostname)
+{
+    const QString buildOptionsHelper = [hostname]() {
+        static QString buildOptions =
```
milianw, line 114: dito, must not be static

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    QStringList perfCommand = {QStringLiteral("record"), QStringLiteral("-o"), QStringLiteral("-")};
+    perfCommand += perfOptions;
+    perfCommand += recordOptions;
+    m_outputFile = new QFile(outputPath);
+    m_outputFile->open(QIODevice::WriteOnly);
+
+    m_recordProcess = createSshProcess(m_deviceName, perfCommand);
+
+    emit recordingStarted(QLatin1String("perf"), perfCommand);
```
milianw, line 188: QStringLiteral

src/perfrecordssh.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (!folderInfo.isDir()) {
+        emit recordingFailed(tr("'%1' is not a folder.").arg(folderPath));
+        return;
+    }
+    if (!folderInfo.isWritable()) {
+        emit recordingFailed(tr("Folder '%1' is not writable.").arg(folderPath));
+        return;
+    }
+
+    qRegisterMetaType<QProcess::ExitStatus>("QProcess::ExitStatus");
```
milianw, line 178: what's this for? I don't see any queued or inter-thread signal/slot emission with this type?

that said, assuming it is required: do you really need to specify the string or is it enough to say `qRegisterMetaType<QProcess::ExitStatus>()`?

src/perfrecordssh.h
```diff
@@ -0,0 +1,61 @@
...
+    bool canUseAio() override;
+    bool canCompress() override;
+    bool isPerfInstalled() override;
+
+private:
+    void startRecording(const QStringList& perfOptions, const QString& outputPath, const QStringList& recordOptions,
+                        const QString& workingDirectory);
+
+    std::unique_ptr<QProcess> m_recordProcess = nullptr;
+    QFile* m_outputFile;
```
milianw, line 58: init

src/perfrecordssh.h
```diff
@@ -0,0 +1,61 @@
...
+    bool canSwitchEvents() override;
+    bool canUseAio() override;
+    bool canCompress() override;
+    bool isPerfInstalled() override;
+
+private:
+    void startRecording(const QStringList& perfOptions, const QString& outputPath, const QStringList& recordOptions,
+                        const QString& workingDirectory);
+
+    std::unique_ptr<QProcess> m_recordProcess = nullptr;
```
milianw, line 57: unique_ptr will init itself to this automatically, so no need for that here

src/settingsdialog.cpp
```diff
@@ -293,3 +300,110 @@ void SettingsDialog::addCallgraphPage()
...
+                    }
+                });
+
+        ssh->start();
+    });
+
+    connect(sshPage->copyKeyButton, &QPushButton::pressed, this, [this, devices] {
+        auto sshCopyId = createSshProcess(sshPage->hostnameLineEdit->text(), sshPage->usernameLineEdit->text(),
+                                          sshPage->sshOptionsLineEdit->text(), QStringList({QStringLiteral("perf")}),
+                                          QLatin1String("ssh-copy-id"));
```
milianw, line 390: QStringLiteral

src/settingsdialog.cpp
```diff
@@ -293,3 +300,110 @@ void SettingsDialog::addCallgraphPage()
...
+        const auto sshOptions = group.readEntry("sshoptions");
+
+        sshPage->hostnameLineEdit->setText(hostname);
+        sshPage->usernameLineEdit->setText(username);
+        sshPage->sshOptionsLineEdit->setText(sshOptions);
+    };
+
+    MultiConfigWidget* devices = new MultiConfigWidget(this);
+
+    // TODO: fix tab order
```
milianw, line 340: yes please

src/ssh.h
```diff
@@ -0,0 +1,25 @@
...
+#pragma once
+
+#include <memory>
+
+#include <QString>
+
+class QStringList;
+class QProcess;
+
+// used in the dialog to test connection and copy ssh keys
```
milianw, line 17: can you please expand the documentation? what is the difference between options, command and executable?

src/ssh.h
```diff
@@ -0,0 +1,25 @@
...
+std::unique_ptr<QProcess> createSshProcess(const QString& username, const QString& hostname, const QString& options,
+                                           const QStringList& command,
+                                           const QString& executable = QStringLiteral("ssh"));
+
+// used when recording
+std::unique_ptr<QProcess> createSshProcess(const QString& deviceName, const QStringList& command);
```
milianw, line 23: how does this differ from the above then?

src/perfrecordssh.h
```diff
@@ -0,0 +1,55 @@
...
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include "perfrecord.h"
+
+class QFile;
+
+class PerfRecordSSH : public PerfRecord
```
milianw, line 15: please add some basic documentation on how this is implemented. basically just something that says we overload a bunch of API to disable features and then use an external ssh process to communicate with another machine to remote control the perf session there

src/ssh.h
```diff
@@ -0,0 +1,17 @@
...
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+class QString;
+class QStringList;
+class QProcessEnvironment;
+
+QStringList assembleSSHArguments(const QString& deviceName, const QStringList& command);
```
milianw: please add basic api documentation to all of these

src/ssh.h
```diff
@@ -0,0 +1,17 @@
...
+QProcessEnvironment sshEnvironment();
```
milianw: important to note here: how does this differ from the normal process environment?

milianw (changes requested):

jeff-dagenais:
related to #12

GitMensch:
> related to #12

Actually this "fixes #12" - maybe someone can edit the first post or better directly link it.

GitMensch:
I've tested that with the generated appimage and when doing "test connection and save" I always just get an empty error message:

error

with stderr having

> QProcess: Destroyed while process ("/usr/bin/ssh") is still running.

How to debug this issue? (note: `/usr/bin/ssh -K ruser@rmach` works like a charm)

Side note: the configuration dialog is named "Paths and Architecture Settings" which is getting even more wrong after applying this, I've created #436 to tackle that.

lievenhey:
@GitMensch It should now work (I messed up the order of arguments and ended up calling ssh host@user)

GitMensch:
Rechecked with hotspot-v1.3.0-374-g97c6a8e-x86_64.AppImage - still the same result (empty error + that message in stderr).
Can I suggest to add "command failed: $sshCommandHere" as start of the error message?

GitMensch:
The SSH config now does work (note: if available it would be useful to read in `~/.ssh/config` for the list of hosts).
To get it saved on has to manually click on "copy config".

... but I was not able to record, the "start profiling" was always grey.

src/multiconfigwidget.cpp
```diff
@@ -1,119 +1,211 @@
...
 
-    if (!m_config.isValid())
+    m_configWidget->currentConfigComboBox->clear();
+    auto configGroups = configs();
+    for (const auto& config : configGroups) {
+        m_configWidget->currentConfigComboBox->addItem(config, config);
+    }
+
+    m_configWidget->currentConfigComboBox->setCurrentIndex(0);
+    // m_configWidget->currentConfigComboBox->setValidator
```
milianw, line 127: cleanup

src/multiconfigwidget.cpp
```diff
@@ -1,119 +1,211 @@
...
+
+#include "ui_multiconfigwidget.h"
+
+namespace {
+template<typename Widget, class RestoreFunc>
+void restoreWidget(QWidget* parent, const KConfigGroup& group, RestoreFunc restoreFunc)
+{
+    Q_ASSERT(parent);
+    Q_ASSERT(group.isValid());
+    for (auto* child : parent->findChildren<Widget*>()) {
```
milianw, line 27: this looks super fishy and brittle, please make the restoration logic static and explicit, instead of dynamic like proposed here

this could otherwise break if e.g. we save/restore data within complex widgets that use e.g. a QLineEdit internaly _and_ give it an object name

meaning: this logic here is fine, as long as the list of widgets that should be restored is hardcoded and does not rely on `findChildren`. I.e. change the API such that we pass along the list of widgets, or register them somehow, maybe just pass it along as a `QVector<QWidget*>` to the `setChildWidget` API

src/multiconfigwidget.cpp
```diff
@@ -1,119 +1,211 @@
...
 
 MultiConfigWidget::~MultiConfigWidget() = default;
 
-QString MultiConfigWidget::currentConfig() const
+void MultiConfigWidget::setChildWidget(QWidget* widget)
 {
-    return m_comboBox->currentData().toString();
+    m_child = widget;
+    m_child->setParent(this);
+    m_configWidget->layout->replaceWidget(m_configWidget->placeholder, m_child);
```
milianw, line 110: the placeholder is leaked, or?

> The returned layout item is no longer owned by the layout and should be either deleted or inserted to another layout. The widget from is no longer managed by the layout and may need to be deleted or hidden. The parent of widget from is left unchanged.

src/multiconfigwidget.cpp
```diff
@@ -1,119 +1,211 @@
...
+
+    m_configWidget->currentConfigComboBox->setCurrentIndex(0);
+    // m_configWidget->currentConfigComboBox->setValidator
+
+    if (!configGroups.isEmpty()) {
+        loadConfig(currentConfig());
+    }
+    emit configsChanged();
+}
+void MultiConfigWidget::loadConfig(const QString& name)
```
milianw, line 134: empty line

src/multiconfigwidget.h
```diff
@@ -11,32 +10,44 @@
...
+namespace Ui {
+class MultiConfigWidget;
+}
+
+/** this widget allows fast switching between different configurations
+ * of m_child
+ * saveConfig will automatically save all changes to m_group
+ * use setChildWidget for an unparented one and moveWidgetToChild for a parented one
+ *
+ * ATTENTION: don't connect textChanged, etc to saveCurrentConfig, use returnPressed, ... otherwise saveCurrentConfig
```
milianw, line 22: can't you catch this with a simple bool flag, which you set to true when saving/restoring, and if so, you don't trigger another restore while that is set to true?

milianw (changes requested):
only had time for the first patch for now

src/multiconfigwidget.cpp
lievenhey: would it be fine if the search only goes one level deep? I am not really keen on writing a save/restore function for every form

src/recordpage.cpp
```diff
@@ -192,12 +191,19 @@ RecordPage::RecordPage(QWidget* parent)
...
+            [this](const QString& app) { ui->multiConfig->setConfigGroup(applicationConfig(app)); });
+
+    connect(ui->workingDirectory, qOverload<const QString&>(&KUrlRequester::returnPressed), this,
+            [this](const QString& cwd) {
+                ui->multiConfig->saveCurrentConfig();
+                m_recordHost->setCurrentWorkingDirectory(cwd);
+            });
+
+    connect(ui->applicationParametersBox, &QLineEdit::editingFinished, this, [this] {
+        ui->multiConfig->saveCurrentConfig(); // TODO: set app params
```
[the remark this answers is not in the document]
lievenhey: leaked into this commit

src/models/eventmodelproxy.cpp
```diff
@@ -0,0 +1,20 @@
+/*
+    SPDX-FileCopyrightText: Lieven Hey <lieven.hey@kdab.com>
+    SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#include "eventmodelproxy.h"
+
+EventModelProxy::EventModelProxy(QObject* parent)
```
milianw, line 10: this file shouldn't be part of the patch that touches the multiconfig widget

also, a proxy that never filters is pretty bogus conceptually

src/multiconfigwidget.cpp
```diff
@@ -1,119 +1,203 @@
...
+            if (value.isEmpty()) {
+                continue;
+            }
+            auto index = widget->findText(value);
+            if (index == -1) {
+                index = widget->count();
+                widget->addItem(value);
+            }
+            widget->setCurrentIndex(index);
         }
```
milianw, line 146: else emit a warning that this form widget type isn't supported

src/multiconfigwidget.cpp
```diff
@@ -1,119 +1,203 @@
...
+        if (auto widget = qobject_cast<const QLineEdit*>(formWidget)) {
+            group.writeEntry(name, widget->text());
+        } else if (auto widget = qobject_cast<const KUrlRequester*>(formWidget)) {
+            group.writeEntry(name, widget->text());
+        } else if (auto widget = qobject_cast<const KEditListWidget*>(formWidget)) {
+            auto data = widget->items().join(QLatin1Char(':'));
+            group.writeEntry(name, data);
+        } else if (auto widget = qobject_cast<const QComboBox*>(formWidget)) {
+            group.writeEntry(name, widget->currentText());
+        }
```
milianw, line 175: dito

src/multiconfigwidget.h
```diff
@@ -11,32 +10,43 @@
...
 class MultiConfigWidget : public QWidget
 {
     Q_OBJECT
 public:
     MultiConfigWidget(QWidget* parent = nullptr);
-    ~MultiConfigWidget();
+    ~MultiConfigWidget() override;
 
-    QString currentConfig() const;
+    void setChildWidget(QWidget* widget, const QVector<QWidget*>& formWidgets);
```
milianw, line 29: add basic API documentation to explain that `widget` is basically the content widget, and `formWidgets` are the individual widgets (that must have a unique object name!) which will get saved/loaded automatically

src/settingsdialog.cpp
```diff
@@ -262,3 +266,73 @@ void SettingsDialog::addSourcePathPage()
...
+            path = QStandardPaths::findExecutable(QStringLiteral("ssh-copy-id"));
+        }
+        if (path.isEmpty()) {
+            sshSettingsPage->messageWidget->setText(tr("Could not find ssh-copy-id"));
+            sshSettingsPage->messageWidget->show();
+            return;
+        }
+
+        if (qEnvironmentVariableIsSet("SSH_ASKPASS")) {
+            auto env = QProcessEnvironment::systemEnvironment();
```
milianw, line 300: why do we have to set this, isn't the system env used by default and that will inherit the ssh_askpass value then, no?

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+#include <QProcess>
+#include <QStandardPaths>
+#include <QThread>
+
+#include <KSharedConfig>
+
+#include "settings.h"
+
+// TODO: remove
```
milianw, line 18: ack

src/recordhost.cpp
```diff
@@ -140,12 +140,35 @@ RecordHost::PerfCapabilities fetchLocalPerfCapabilities(const QString& perfPath)
...
+
+    return capabilities;
+}
 }
 
 RecordHost::RecordHost(QObject* parent)
     : QObject(parent)
     , m_checkPerfCapabilitiesJob(this)
     , m_checkPerfInstalledJob(this)
+    , m_remoteDevice(this)
```
milianw, line 171: passing `this` when this is owned on the stack sounds wrong. make this a ptr and forward declare in the header? and only actually instantiate when we are _not_ local?

src/recordhost.cpp
```diff
@@ -262,16 +276,25 @@ void RecordHost::setCurrentWorkingDirectory(const QString& cwd)
             m_cwd = cwd;
             emit currentWorkingDirectoryChanged(cwd);
         }
-        return;
+    } else {
+        if (!m_remoteDevice.checkIfDirectoryExists(cwd)) {
```
milianw, line 281: here and elsewhere: these operations must be async, otherwise we'll freeze the UI on a bad network connection

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+bool RemoteDevice::isConnected() const
+{
+    return QFile(m_tempDir.path() + QStringLiteral("/ssh")).exists();
+}
+
+bool RemoteDevice::checkIfProgramExists(const QString& program) const
+{
+    auto ssh = sshProcess({QStringLiteral("command"), program});
+    ssh->start();
+    ssh->waitForFinished();
```
milianw, line 127: blocking operation called from main thread -> bad

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+    m_connection = sshProcess({});
+
+    connect(m_connection, qOverload<int, QProcess::ExitStatus>(&QProcess::finished), this,
+            [connection = m_connection, path = m_tempDir.path(), this](int exitCode, QProcess::ExitStatus) {
+                if (exitCode != 0) {
+                    emit failedToConnect();
+                } else {
+                    emit disconnected();
+                }
+                qDebug() << exitCode << connection->readAll();
```
milianw: cleanup

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+            });
+
+    m_connection->start();
+}
+
+void RemoteDevice::disconnect()
+{
+    if (m_connection) {
+        if (m_connection->state() == QProcess::ProcessState::Running) {
+            // ssh stops once you clode the write channel
```
milianw, line 109: typo: close

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // 128 -> not found
+    // perf will return 1 and display the help message
+    return exitCode != 128;
+}
+
+bool RemoteDevice::checkIfDirectoryExists(const QString& directory) const
+{
+    auto ssh = sshProcess({QStringLiteral("test"), QStringLiteral("-d"), directory});
+    ssh->start();
+    ssh->waitForFinished();
```
milianw, line 139: dito

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto exitCode = ssh->exitCode();
+    ssh->deleteLater();
+    return exitCode == 0;
+}
+
+bool RemoteDevice::checkIfFileExists(const QString& file) const
+{
+    auto ssh = sshProcess({QStringLiteral("test"), QStringLiteral("-f"), file});
+    ssh->start();
+    ssh->waitForFinished();
```
milianw, line 149: dito

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto exitCode = ssh->exitCode();
+    ssh->deleteLater();
+    return exitCode == 0;
+}
+
+QByteArray RemoteDevice::getProgramOutput(const QStringList& args) const
+{
+    auto ssh = sshProcess(args);
+    ssh->start();
+    ssh->waitForFinished();
```
milianw, line 159: dito

src/remotedevice.cpp
```diff
@@ -0,0 +1,188 @@
...
+    auto output = ssh->readAllStandardOutput();
+    ssh->deleteLater();
+    return output;
+}
+
+QProcess* RemoteDevice::sshProcess(const QStringList& args)
```
milianw, line 165: I don't get it, why do we have a const and a non-const version of this?

milianw (changes requested):
I think we also need some basic test coverage for this, maybe for starters a manual test binary that we can invoke manually to run some tests

then integrate that into the docker images somehow?

src/models/eventmodelproxy.cpp
lievenhey: I think I committed this accidentally

src/settingsdialog.cpp
lievenhey: right, this piece of code is useless

lievenhey: Actually this brakes if the `SSH_ASKPASS` and `SSH_ASKPASS_REQUIRE` are not set. In this case the user is prompted on the command line, which may not be accessable.

milianw: I don't have `_REQUIRE` set in my env and it works. If `SSH_ASKPASS` isn't set, we can set it to `ksshaskpass` (and error out if that isn't available)

src/remotedevice.cpp
milianw: make it so that the `const` version is enough, i.e. don't rely on QObject ownership, just return a `std::unique_ptr<QProcess>`.

PR #360, opened 2022-01-19
rename gui classes in settingsdialog

lievenhey:
some names do not make sense anymore (settingsdialog.ui just contains
the unwind settings) so I changes them to more descriptive names

additional changes:
- removed double /* in settings.cpp
- added #ifdef QCustomPlot_FOUND to resultspage.cpp, otherwise the
  dockify call will fail since the compiler doen't know that
  m_frequencyPage is a QWidget


milianw:
these cleanups/fixes should be separate commits, not one big change that groups a lot of unrelated stuff

PR #364, opened 2022-02-20
Change desktop file name to include domain name

nicolasfella:
This fixes taskbar and window icon on Wayland


nicolasfella:
The implicit desktopfilename is com.kdab.hotspot, renaming the desktop file from hotspot.desktop to com.kdab.hotspot.desktop would be another option to fix this

milianw:
I guess the more verbose .desktop file would be the better choice to make it more explicit or what do you think @nicolasfella ?

milianw:
the change to the perfparser submodule is unrelated (or? what did you change there?)

milianw (changes requested):
perfparser changes are unrelated

milianw:
undid the unrelated perfparser change and pushed it - thanks @nicolasfella

Issue #345, opened 2021-10-31
ignore self cost for trace points


milianw:
For trace points (like sched switches or uprobes) and deduced cost like off-CPU the self cost is never interesting, as it is only non-zero in the function where the trace point is positioned. This leads to useless data in the tables which hides the more interesting inclusive data.

We should detect this situation and then ignore/hide these columns by default. See also #344

PR #367, opened 2022-03-08
add function to automatically hide all tracepoint columns

lievenhey:
these columns are not interesting since they are only non zero in the
function where the tracepoint is positioned

### DIFF
--- a/src/parsers/perf/perfparser.cpp
+++ b/src/parsers/perf/perfparser.cpp
@@ -1161,7 +1161,9 @@ public:
             thread->offCpuTime += switchTime;
 
             if (eventResult.offCpuTimeCostId == -1) {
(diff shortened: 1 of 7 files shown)


lievenhey:
closes #345

lievenhey:
I hope I found all tracepoints, but can you run `sudo perf list | grep Tracepoint | cut -d ":" -f1 | uniq | md5sum` and send me all you tracepoints in case the output is not `22c003cf0b3ade241fadc0d76959c181`

lievenhey:
done. The perfparser provides now a set with all tracepoint derived cost names. This set is used to hide the columns in hideTracepointColumns.

milianw (changes requested):
please hide the _self_ cost, but do keep the _inclusive_ cost as that's useful.

lievenhey:
done

milianw:
thanks, I fixed one minor issue and expanded it a bit.

great work!

PR #368, opened 2022-03-09
Frequencypage improvements

lievenhey:


milianw:
In the last patch that adds the combo box, it should only offer the costs, not the cores for selection. I.e. I would expect the combo box to look exactly like in the flame graph view, and the graph to always show data for all cores

lievenhey:
done

milianw:
ok, lgtm now - thanks!

PR #370, opened 2022-03-16
support custom bottom cost aggregation

### LINKED ISSUE #15: dynamic column aggregation configuration
Similar to WPA, the aggregated cost views should let the user decide how to aggregate the (inclusive) cost data. By default, we should aggregate by symbol name. But it would be great to also support aggregation by:

- thread
- program (when profiling multple programs)
- dso
- cpu

lievenhey:
closes #15


lievenhey:
Example with cpp-threadnames

src/resultsbottomuppage.cpp
```diff
@@ -95,6 +95,33 @@ ResultsBottomUpPage::ResultsBottomUpPage(FilterAndZoomStack* filterStack, PerfPa
...
+            [this](const Data::BottomUpResults& data) { bottomUpByCpu = data; });
+
+    for (const auto& aggregationType :
+         {QLatin1String("Symbol"), QLatin1String("Thread"), QLatin1String("Process"), QLatin1String("CPU")}) {
+        ui->costAggregationComboBox->addItem(aggregationType);
+    }
+
+    connect(ui->costAggregationComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged), this, [this] {
+        const auto currentText = ui->costAggregationComboBox->currentText();
+        if (currentText == QLatin1String("Symbol")) {
```
milianw, line 115: introduce an enum and set that as data and extract it here via currentData, then switch over it instead of doing string comparisons

note e.g. that the strings should be translatable which would then break here

src/resultsbottomuppage.cpp
```diff
@@ -95,6 +95,33 @@ ResultsBottomUpPage::ResultsBottomUpPage(FilterAndZoomStack* filterStack, PerfPa
...
+            [this](const Data::BottomUpResults& data) { bottomUpByThread = data; });
+
+    connect(parser, &PerfParser::bottomUpDataPerProcessAvailable, this,
+            [this](const Data::BottomUpResults& data) { bottomUpByProcess = data; });
+
+    connect(parser, &PerfParser::bottomUpDataPerCpuAvailable, this,
+            [this](const Data::BottomUpResults& data) { bottomUpByCpu = data; });
+
+    for (const auto& aggregationType :
+         {QLatin1String("Symbol"), QLatin1String("Thread"), QLatin1String("Process"), QLatin1String("CPU")}) {
```
milianw, line 109: see below, make it an enum and add a mapping from enum to translated string here

milianw (changes requested):
while this goes in the right direction, I'm sorry to say that this is far from ideal

a) we must not compute _all_ these various groupings right away. for larger datasets this would yield an immense 4x increase in memory consumption - the trees can become quite large. instead, let's reparse the data on demand similar to how we do it when we apply a filter to change the grouping. 

b) additionally, we should remember the user's preferred choice of grouping and restore it on restarts directly

c) then, such features must be used throughout the app. right now, only the bottom-up view has this feature which is bogus. we also need it for the top-down data and the flame graph too.

milianw:
Note that I pushed a little fix to correctly format cpu names. `QLatin1String().arg(int)` does not do what you are expecting, you must use `QLatin1String().arg(QString::number(int))`, otherwise the value int is interpreted as a char directly

lievenhey:
I added a test, but the results are not really usable due to the frequency mode

milianw:
> I added a test, but the results are not really usable due to the frequency mode

What do you mean by that exactly? Why is the test not usable, what is the problem?

src/parsers/perf/perfparser.cpp
```diff
@@ -838,6 +841,24 @@ class PerfParserPrivate : public QObject
             }
         }
 
+        const auto costAggregation = Settings::instance()->costAggregation();
```
milianw, line 844: It's not safe to access the Settings API from a background thread, this is not thread safe. I just fixed the similar issue for the debuginfod urls and pushed it directly. Please fix it similarly here

src/parsers/perf/perfparser.cpp
```diff
@@ -838,6 +841,24 @@ class PerfParserPrivate : public QObject
...
+
+        if (costAggregation == Settings::CostAggregation::ByThread) {
```
milianw, line 846: the code is duplicated below, merge this and say something like `costAggregation == ByThread || costAggregatoin == ByProcess`

src/parsers/perf/perfparser.cpp
```diff
@@ -838,6 +841,24 @@ class PerfParserPrivate : public QObject
...
+            for (auto& thread : bottomUpResult.root.children) {
+                if (tidToName.contains(thread.symbol.symbol)) {
```
milianw, line 848: this is a classic performance antipattern - you do two hash lookups here which is wasteful. instead, use find + iterators:

```
auto it = tidToName.constFind(thread.symbol.symbol);
if (it != tidToName.constEnd()) {
    thread.symbol.prettySymbol = *it;
    thread.symbol.symbol = *it;
}
```

src/parsers/perf/perfparser.cpp
milianw: also, this whole process of giving the top-level entries pretty names here is bogus - that means the topDownResults would have the non-prettied names as leaf nodes, which is undesired. they should also be root nodes, and also get prettified

similarly, we need to prettify or skip these entries when we buildCallerCalleeResult etc. above.

to fix this, apply the pretty names early (see suggestions below). but one way or another the handling for the buildTopDownResult needs to be rethought.

src/parsers/perf/perfparser.cpp
```diff
@@ -1091,7 +1112,25 @@ class PerfParserPrivate : public QObject
...
+            bottomUpResult.addEvent({QString::number(sample.tid)}, type, sampleCost.cost, sample.frames);
+            break;
+        case Settings::CostAggregation::ByProcess:
+            bottomUpResult.addEvent({QString::number(sample.pid)}, type, sampleCost.cost, sample.frames);
+            break;
+        case Settings::CostAggregation::ByCPU:
+            bottomUpResult.addEvent({QLatin1String("CPU %1").arg(QString::number(sample.cpu))}, type, sampleCost.cost,
+                                    sample.frames);
+            break;
+        default:
```
milianw, line 1131: don't use a default, we want to get a compiler warning when we extend the enum

src/parsers/perf/perfparser.cpp
```diff
@@ -1091,7 +1112,25 @@ class PerfParserPrivate : public QObject
...
 
-        bottomUpResult.addEvent(type, sampleCost.cost, sample.frames, frameCallback);
+        const auto costAggregation = Settings::instance()->costAggregation();
+
+        switch (costAggregation) {
+        case Settings::CostAggregation::BySymbol:
+            bottomUpResult.addEvent(type, sampleCost.cost, sample.frames, frameCallback);
+            break;
+        case Settings::CostAggregation::ByThread:
+            bottomUpResult.addEvent({QString::number(sample.tid)}, type, sampleCost.cost, sample.frames);
```
milianw, line 1122: instead of using the raw tid here, map it directly to a thread name via

```
commands.value(sample.pid).value(sample.tid)
```

src/parsers/perf/perfparser.cpp
```diff
@@ -1091,7 +1112,25 @@ class PerfParserPrivate : public QObject
...
+            break;
+        case Settings::CostAggregation::ByProcess:
+            bottomUpResult.addEvent({QString::number(sample.pid)}, type, sampleCost.cost, sample.frames);
```
milianw, line 1125:
```
commands.value(sample.pid).value(sample.pid)
```

src/resultsbottomuppage.cpp
```diff
@@ -95,6 +96,25 @@ ResultsBottomUpPage::ResultsBottomUpPage(FilterAndZoomStack* filterStack, PerfPa
...
+
+    for (const auto& aggregationType :
+         QList<AggregationType>({{QLatin1String("Symbol"), Settings::CostAggregation::BySymbol},
+                                 {QLatin1String("Thread"), Settings::CostAggregation::ByThread},
+                                 {QLatin1String("Process"), Settings::CostAggregation::ByProcess},
+                                 {QLatin1String("CPU"), Settings::CostAggregation::ByCPU}})) {
+        ui->costAggregationComboBox->addItem(aggregationType.name, QVariant::fromValue(aggregationType.aggregation));
+    }
+
+    connect(ui->costAggregationComboBox, QOverload<int>::of(&QComboBox::currentIndexChanged), this, [this] {
```
milianw, line 114: use qOverload instead of QOverload<>::of

src/resultsbottomuppage.cpp
```diff
@@ -95,6 +96,25 @@ ResultsBottomUpPage::ResultsBottomUpPage(FilterAndZoomStack* filterStack, PerfPa
...
         });
+
+    struct AggregationType
+    {
+        QString name;
+        Settings::CostAggregation aggregation;
+    };
+
+    for (const auto& aggregationType :
+         QList<AggregationType>({{QLatin1String("Symbol"), Settings::CostAggregation::BySymbol},
```
milianw, line 107: don't use a QList (required memory allocation), use an initializer list directly

also, use tr() to get translated strings

src/resultsbottomuppage.h
```diff
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include "data.h"
```
milianw, line 11: not needed anymore, undo changes

tests/integrationtests/custom_cost_aggregation_testfiles/by_cpu.txt
```diff
@@ -0,0 +1,9 @@
+
+	CPU 0
+		do_lookup_x
+			_dl_lookup_symbol_x
+				_dl_relocate_object
+					dl_main
+						_dl_sysdep_start
+							_dl_start
+								_start
```
milianw, line 9: where is the check_match branch we see in by_process.txt?!

tests/integrationtests/custom_cost_aggregation_testfiles/by_symbol.txt
```diff
@@ -0,0 +1,14 @@
+
+	_dl_lookup_symbol_x
+		_dl_relocate_object
+			dl_main
+				_dl_sysdep_start
+					_dl_start
+						_start
+	do_lookup_x
```
milianw, line 8: where is check_match?

tests/integrationtests/custom_cost_aggregation_testfiles/by_symbol.txt
```diff
@@ -0,0 +1,14 @@
+
+	_dl_lookup_symbol_x
```
milianw, line 2: where is do_lookup_x?

tests/integrationtests/custom_cost_aggregation_testfiles/by_thread.txt
```diff
@@ -0,0 +1,15 @@
+
+	cpp-threadnames
+		_dl_lookup_symbol_x
+			_dl_relocate_object
+				dl_main
+					_dl_sysdep_start
+						_dl_start
+							_start
+		do_lookup_x
```
milianw, line 9: dito above

also, generally, it would be much better to use a different cost, such as off-CPU time or scheduler switches instead of cycles for this comparison. I would then also expect the other threads to be actually shown here, which currently does not seem to be the case, making me doubt that the aggregation actually works?

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -596,6 +596,58 @@ private slots:
...
+    void testCustomCostAggregation()
+    {
+        QFETCH(Settings::CostAggregation, aggregation);
+        QFETCH(QString, filename);
+
+        const auto actualBinaryFile = QFINDTESTDATA("custom_cost_aggregation_testfiles/" + filename);
+
+        Settings::instance()->setCostAggregation(aggregation);
+
+        QStringList perfOptions = {"--call-graph", "dwarf", "--switch-events"};
```
milianw, line 628: please only record the data once on one of the static binaries we ship in perfparser and then analyze the same perf data with different aggregations to see the differences. add the binary perf.data to git then and don't record it in the test. otherwise we will never produce stable tests that can be run on arbitrary machines with arbitrary libc etc.

furthermore, make sure to record with `--sample-cpu`, otherwise the CPU value will always be 0 for all samples.

milianw (changes requested):

lievenhey:
with the default arguments perf only records two samples

lievenhey:
not really usable if you want to test custom cost allocation when only 6 functions are recorded

src/parsers/perf/perfparser.cpp
lievenhey: does this also apply to `ByProcess`? In that case I can remove the lookup in line 844

lievenhey:
ah, with `--sample-cpu --event sched:sched_switch --switch-events` I get usable data for the test

milianw:
right, that test binary is not meant to generate cycle samples. you should consider using a different one for that purpose, or even create a custom other one if desired.

tests/integrationtests/custom_cost_aggregation_testfiles/by_cpu.txt
```diff
@@ -0,0 +1,143 @@
+
+	CPU 0
+
```
milianw, line 3: what about these empty lines, where do they come from? is that kernel stuff that is missing a proper kallsyms? can you add that too please?

src/parsers/perf/perfparser.cpp
```diff
@@ -820,12 +822,15 @@ class PerfParserPrivate : public QObject
         buildPerLibraryResult();
         buildCallerCalleeResult();
 
+        QHash<QString, QString> tidToName;
```
milianw, line 825: this is unused and can be removed

milianw (changes requested):
I'll adopt and bring this over the finishing line

milianw:
thanks!

PR #371, opened 2022-03-20
CMake: Don't assume KDE_INSTALL_* variables are relative

krop:
ECM only makes sure KDE_INSTALL_FULL_BINDIR and KDE_INSTALL_FULL_LIBEXECDIR
contain absolute paths.


milianw:
thanks, merged

PR #372, opened 2022-03-23
add hint that --debugPaths isn't relative to CWD

### LINKED ISSUE #174: Using relative path in debugPaths doesn't work as expected
If a relative path is supplied in the `--debugPaths` list, it isn't relative to the CWD, but rather to where the file that needs debug symbols is. Quoting [libdwfl.h](https://sourceware.org/git/?p=elfutils.git;a=blob;f=libdwfl/libdwfl.h;h=a0c1d35735c4ce64c5b41b5d91d812a5550f110c;hb=HEAD#l275) (which I think is what's used here):

>   In searches for debuginfo by name, if the remainder of the
>   element is empty, the directory containing the main file is
>   tried; if it's an absolute path name, the absolute directory path
>   (and any subdirectory of that path) containing the main file is
>   taken as a subdirectory of this path; a relative path name is taken
>   as a subdirectory of the directory containing the main file.
>   Hence for /usr/bin/ls, the default string ":.debug:/usr/lib/debug"
>   says to look in /usr/bin, then /usr/bin/.debug, then the path subdirs
>   under /usr/lib/debug, in the order /usr/lib/debug/usr/bin, then
>   /usr/lib/debug/bin, and finally /usr/lib/debug, for the file name in
>   the .gnu_debuglink section (or "ls.debug" if none was found). 

Perhaps every path in debugPaths that is relative and exists should have the absolute path added to the end of debugPaths?

lievenhey:
closes #174

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -110,7 +110,9 @@ int main(int argc, char** argv)
 
     QCommandLineOption debugPaths(
         QLatin1String("debugPaths"),
-        QCoreApplication::translate("main", "Colon separated list of paths that contain debug information."),
+        QCoreApplication::translate("main",
+                                    "Colon separated list of paths that contain debug information. These paths are "
+                                    "relative to the executable and not to the current working directory."),
         QLatin1String("paths"));
     parser.addOption(debugPaths);
 


milianw:
thanks

PR #377, opened 2022-05-04
add assertion to settings::instance

lievenhey:
accessing settings::instance is not thread safe and should only be done
from the main thread


milianw:
thanks!

Issue #235, opened 2020-05-06
Make it possible to view two or more results side by side


obfuscated:
Currently it seems impossible to compare results from two profiling runs.
I think the only way to do this is to start two separate hotspot instances and alt-tab.
It would be a lot better if it is possible to view the results in tabs and switch between them.

This is pretty mandatory (IMO) feature for quick profile-change code-build-profile cycles.

I don't care too much if serializing the results is possible, but it would be a nice feature for more serious session.

I'm at 1.2.0.

milianw:
I agree, see #37  where this is tracked already

obfuscated:
No idea why you're closing this. It is separate and simpler thing. In fact I don't want automatic diffs, they don't work in the real world. I just want better UI.

milianw:
because I misunderstood the report :)

PR #380, opened 2022-05-05
Multiple windows

lievenhey:
let perfparser use `Settings::instance` to access sysroot, ...
removes the broken feature to open multiple windows and replaces it with a working one

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -167,31 +167,53 @@ int main(int argc, char** argv)
 
     const auto settings = Settings::instance();
     applyCliArgs(settings);
-    for (const auto& file : parser.positionalArguments()) {
-        auto window = new MainWindow;
+
+    auto files = parser.positionalArguments();
+    auto arguments = app.arguments();
+    arguments.pop_front(); // pop executable
+
+    for (const auto& file : files) {
+        auto it = std::find(arguments.begin(), arguments.end(), file);
+        if (it != arguments.end()) {
+            arguments.erase(it);
+        }
+    }
+
+    while (files.size() > 1) {
+        // spawn new instances if we have more than one file argument
+        QProcess process;
+        process.setProgram(app.applicationFilePath());
+        const auto file = files.front();
+        files.pop_front();
+        auto args = arguments;
+        args.append(file);
+        process.setArguments(args);
+        if (!process.startDetached()) {
+            qWarning() << "failed to start new hotspot instance";
+        }
+    }
+
+    // we now only have at most one file
+    Q_ASSERT(files.size() <= 1);
+
+    auto window = new MainWindow;
+    if (!files.isEmpty()) {
+        const auto file = files.front();
         QFileInfo info(file);
         if (info.isFile()) {
             window->openFile(file);
         } else if (info.isDir()) {
             window->openFile(file + QStringLiteral("/perf.data"));
         }
-        window->show();
-    }
-
-    // show at least one mainwindow
-    if (parser.positionalArguments().isEmpty()) {
-        auto window = new MainWindow;
-        applyCliArgs(Settings::instance());
-
+    } else {
         // open perf.data in current CWD, if it exists
         // this brings hotspot closer to the behavior of "perf report"
         const auto perfDataFile = QStringLiteral("perf.data");
         if (QFile::exists(perfDataFile)) {
             window->openFile(perfDataFile);
         }
-
-        window->show();
     }
+    window->show();
 
     return app.exec();
 }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -169,28 +169,25 @@ int main(int argc, char** argv)
     applyCliArgs(settings);
 
     auto files = parser.positionalArguments();
-    auto arguments = app.arguments();
-    arguments.pop_front(); // pop executable
-
-    for (const auto& file : files) {
-        auto it = std::find(arguments.begin(), arguments.end(), file);
-        if (it != arguments.end()) {
-            arguments.erase(it);
-        }
-    }
+    const auto originalArguments = app.arguments();
+    // remove leading executable name and trailing positional arguments
+    const auto minimalArguments = originalArguments.mid(1, originalArguments.size() - 1 - files.size());
 
     while (files.size() > 1) {
         // spawn new instances if we have more than one file argument
-        QProcess process;
-        process.setProgram(app.applicationFilePath());
-        const auto file = files.front();
-        files.pop_front();
-        auto args = arguments;
+        const auto file = files.takeLast();
+        auto process = new QProcess(&app);
+        QObject::connect(process, &QProcess::errorOccurred, &app, [=]() {
+            qWarning() << file << process->errorString();
+        });
+        // the event loop locker prevents the main app from quitting while the child processes are still running
+        // we want to keep them all alive and quit them in one go. detaching isn't as nice, as we would not be
+        // able to quit all apps in one go via Ctrl+C then anymore'
+        QObject::connect(process, qOverload<int, QProcess::ExitStatus>(&QProcess::finished), &app,
+                         [lock = std::make_unique<QEventLoopLocker>()]() mutable { lock.reset(); });
+        auto args = minimalArguments;
         args.append(file);
-        process.setArguments(args);
-        if (!process.startDetached()) {
-            qWarning() << "failed to start new hotspot instance";
-        }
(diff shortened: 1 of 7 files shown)


src/main.cpp
```diff
@@ -167,31 +167,50 @@ int main(int argc, char** argv)
...
+    while (files.size() > 1) {
+        // spawn new instances if we have more than file argument
+        QProcess process;
+        process.setProgram(app.applicationFilePath());
+        const auto file = files.front();
+        files.pop_front();
+        auto args = arguments;
+        args.append(file);
+        process.setArguments(args);
+        process.startDetached();
```
milianw, line 191: please check the return type and print an error when this fails

src/main.cpp
```diff
@@ -167,31 +167,50 @@ int main(int argc, char** argv)
...
+    }
+
+    // we now only have at most one file
```
milianw, line 194: assert it

milianw (changes requested):

src/main.cpp
```diff
@@ -167,31 +167,50 @@ int main(int argc, char** argv)
...
     const auto settings = Settings::instance();
     applyCliArgs(settings);
-    for (const auto& file : parser.positionalArguments()) {
-        auto window = new MainWindow;
+
+    auto files = parser.positionalArguments();
+    auto arguments = app.arguments();
+    arguments.pop_front(); // pop executable
+
+    for (const auto& file : files) {
```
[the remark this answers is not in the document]
lievenhey: this is to remove argv[0] (name of the executable)

lievenhey:
closes #235

src/main.cpp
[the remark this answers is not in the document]
milianw: no, that is done via the `pop_front()` above, I still think you can replace this loop via my one liner above

src/mainwindow.cpp
```diff
@@ -164,6 +164,20 @@ MainWindow::MainWindow(QWidget* parent)
...
+    auto openNewWindow = new QAction(QIcon::fromTheme(QStringLiteral("document-open")), tr("Open in new window"), this);
+    openNewWindow->setShortcut(Qt::Key_O | Qt::ControlModifier | Qt::ShiftModifier);
+    connect(openNewWindow, &QAction::triggered, this, [this] {
+        const auto fileName = QFileDialog::getOpenFileName(this, tr("Open File"), QDir::currentPath(),
+                                                           tr("Data Files (perf*.data perf.data.*);;All Files (*)"));
+        QProcess process;
+        process.setProgram(qApp->applicationFilePath());
+        process.setArguments({fileName});
+        if (!process.startDetached()) {
+            qWarning() << "failed to start new hotspot instance";
```
milianw, line 177: << process.errorString()

src/mainwindow.cpp
```diff
@@ -164,6 +164,20 @@ MainWindow::MainWindow(QWidget* parent)
...
     ui->fileMenu->addAction(KStandardAction::open(this, SLOT(onOpenFileButtonClicked()), this));
+
+    auto openNewWindow = new QAction(QIcon::fromTheme(QStringLiteral("document-open")), tr("Open in new window"), this);
+    openNewWindow->setShortcut(Qt::Key_O | Qt::ControlModifier | Qt::ShiftModifier);
+    connect(openNewWindow, &QAction::triggered, this, [this] {
+        const auto fileName = QFileDialog::getOpenFileName(this, tr("Open File"), QDir::currentPath(),
+                                                           tr("Data Files (perf*.data perf.data.*);;All Files (*)"));
+        QProcess process;
+        process.setProgram(qApp->applicationFilePath());
+        process.setArguments({fileName});
```
milianw, line 175: what about other args, i.e. what happens when we do:

```
hotspot --sysroot foo perf.data
```

and then try to open another window, we would lose the sysroot then, no?

milianw (changes requested):

src/main.cpp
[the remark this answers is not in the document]
lievenhey: QStringList doesn't have a resize function so this is not possible.

src/mainwindow.cpp
lievenhey: in theory it should read it from the settings file

lievenhey: now it is read from settings

src/main.cpp
```diff
@@ -147,51 +147,77 @@ int main(int argc, char** argv)
     auto applyCliArgs = [&](Settings* settings) {
         if (parser.isSet(sysroot)) {
             settings->setSysroot(parser.value(sysroot));
+            settings->setLastUsedEnvironment({});
```
milianw, line 150: why? also, please simplify this repetitive code via lambdas:

```
auto applyCliArg = [&](const QCommandLineOption &arg, auto setter) {
    if (parser.isSet(arg)) {
        (settings->*setter)(parser.value(arg));
        // comment to explain why this is done
        settings->setLastUsedEnvironment({});
    }
};
applyCliArg(kallsyms, &Settings::setKallsyms);
...
```

src/main.cpp
```diff
@@ -147,51 +147,77 @@ int main(int argc, char** argv)
...
 
     const auto settings = Settings::instance();
+    settings->loadFromFile();
     applyCliArgs(settings);
-    for (const auto& file : parser.positionalArguments()) {
-        auto window = new MainWindow;
+
+    auto files = parser.positionalArguments();
+    auto arguments = app.arguments();
+    arguments.pop_front(); // pop executable
```
milianw, line 180:
```
const auto originalArguments = app.arguments();
// remove leading executable name and trailing positional arguments
const auto minimalArguments = originalArguments.mid(1, originalArguments.size() - 1 - files.size());
```

src/main.cpp
```diff
@@ -147,51 +147,77 @@ int main(int argc, char** argv)
...
+        // spawn new instances if we have more than one file argument
+        QProcess process;
+        process.setProgram(app.applicationFilePath());
+        const auto file = files.front();
+        files.pop_front();
+        auto args = arguments;
+        args.append(file);
+        process.setArguments(args);
+        if (!process.startDetached()) {
+            qWarning() << "failed to start new hotspot instance";
```
milianw, line 196: qWarning() << "failed to start new hotspot instance for file" << file << process.errorString();

src/main.cpp
```diff
@@ -147,51 +147,77 @@ int main(int argc, char** argv)
...
+
+    while (files.size() > 1) {
+        // spawn new instances if we have more than one file argument
+        QProcess process;
+        process.setProgram(app.applicationFilePath());
+        const auto file = files.front();
+        files.pop_front();
+        auto args = arguments;
+        args.append(file);
+        process.setArguments(args);
```
milianw, line 194: process.setArguments(QStringList(minimalArguments) << file);

src/main.cpp
```diff
@@ -147,51 +147,77 @@ int main(int argc, char** argv)
...
+
+    for (int i = 0, s = files.size(); i < s; i++) {
+        arguments.removeLast();
+    }
+
+    while (files.size() > 1) {
+        // spawn new instances if we have more than one file argument
+        QProcess process;
+        process.setProgram(app.applicationFilePath());
+        const auto file = files.front();
```
milianw, line 190: const auto file = files.takeLast();

milianw (changes requested):

src/main.cpp
lievenhey: otherwise its broken.

in mainwindow 296 we have
```c++
m_lastUsedSettings = m_config->group("PerfPaths").readEntry("lastUsed");
if (!m_lastUsedSettings.isEmpty()) {
    auto currentConfig = m_config->group("PerfPaths").group(m_lastUsedSettings);
    settings->setSysroot(currentConfig.readEntry("sysroot", ""));
    settings->setAppPath(currentConfig.readEntry("appPath", ""));
    settings->setExtraLibPaths(currentConfig.readEntry("extraLibPaths", ""));
    settings->setDebugPaths(currentConfig.readEntry("debugPaths", ""));
    settings->setKallsyms(currentConfig.readEntry("kallsyms", ""));
    settings->setArch(currentConfig.readEntry("arch", ""));
    settings->setObjdump(currentConfig.readEntry("objdump", ""));
}
```
which will overwrite the sysroot, ... if lastUsed is not empty

milianw (approved):

PR #390, opened 2022-05-31
Better disassembly

### LINKED ISSUE #9: annotate source
add an "annotate source" view, similar to what other profilers have:

- source code with cost aggregated by line, show both inclusive/exclusive cost if possible
- assembly code with cost per instruction, again if possible also show inclusive cost for callq & friends

lievenhey:
A rewrite of the disassembly view
closes #9


lievenhey:
grafik

GitMensch:
This is marvelous! I'd really like to see that integrated.
[side note: is there still a CI building AppImage anywhere? old travis is gone and the last CI release is nearly 7 months old]

src/models/disassemblyoutput.cpp
```diff
@@ -65,12 +69,38 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
...
         // we don't care about the file name
         if (asmLine.startsWith(QLatin1Char('/')) && asmLine.contains(QStringLiteral("file format"))) {
             continue;
+        } else if (asmLine.startsWith(QLatin1Char('/'))) {
+            // extract source code line info
+            // these look like this:
+            // - /usr/include/c++/11.2.0/bits/stl_tree.h:2083 (discriminator 1)
+            // - /usr/include/c++/11.2.0/bits/stl_tree.h:3452
+
+            auto lineNumber = asmLine.right(asmLine.length() - colonIndex - 1);
```
milianw, line 88: rightRef

src/models/disassemblyoutput.h
```diff
@@ -24,9 +24,13 @@ struct DisassemblyOutput
         quint64 addr = 0;
         QString disassembly;
         LinkedFunction linkedFunction;
+        int sourceCodeLine;
```
milianw, line 27: = 0;

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+SourceCodeModel::SourceCodeModel(QObject* parent)
+    : QAbstractTableModel(parent)
+    , m_document(new QTextDocument(this))
+#ifdef KF5SyntaxHighlighting_FOUND
+    , m_repository(std::make_unique<KSyntaxHighlighting::Repository>())
+    , m_highlighter(new KSyntaxHighlighting::SyntaxHighlighter(m_document))
+#endif
+{
+    qRegisterMetaType<QTextLine>();
```
milianw, line 30: should be done in a central place, and ideally share code with the other model somehow

note that this doesn't necessarily mean a central base class, composition and extraction into a separate class would also work

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+    m_sourceCode.clear();
+    endResetModel();
+}
+
+void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
+{
+    if (disassemblyOutput.sourceFileName.isEmpty())
+        return;
+
+    QFile file(disassemblyOutput.sourceFileName);
```
milianw, line 49: this needs to take sysroots into account, no?

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,231 @@
...
+    if (role != Qt::DisplayRole || orientation != Qt::Horizontal)
+        return {};
+
+    if (section == SourceCodeColumn)
+        return tr("Source Code");
+
+    if (section == SourceCodeLineNumber)
+        return tr("Source Code Line Number");
+
+    if (section == Highlight) {
```
milianw, line 138: why is "highlight" a section, that should be a custom Role, no?

src/models/sourcecodemodel.h
```diff
@@ -0,0 +1,76 @@
+
```
milianw, line 1: remove empty line

milianw (changes requested):
please share code between the two models, maybe even extract the whole highlighting and color scheme into a shared single class and then reuse that instead of duplicating it.

src/models/codedelegate.cpp
```diff
@@ -0,0 +1,104 @@
...
+        painter->setBrush(backgroundColor(
+            sourceLine, model->data(model->index(index.row(), m_highlightColumn), Qt::DisplayRole).toBool()));
+        painter->drawRect(option.rect);
+    }
+
+    painter->setPen(pen);
+    painter->setBrush(brush);
+
+    if (index.data().canConvert<QTextLine>()) {
+        const auto line = index.data().value<QTextLine>();
```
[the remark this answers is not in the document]
lievenhey: QTestLine seems to be a simple wrapper so it should be fine. It only has the following members:
```c++
int index;
QTextEngine *eng;
```

src/models/disassemblyoutput.cpp
```diff
@@ -65,12 +69,38 @@ static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray
...
+            // main():
+            // after 0000000000001090 <main>:
+            stream.readLine();
             continue;
         }
 
         // we don't care about the file name
         if (asmLine.startsWith(QLatin1Char('/')) && asmLine.contains(QStringLiteral("file format"))) {
             continue;
+        } else if (asmLine.startsWith(QLatin1Char('/'))) {
```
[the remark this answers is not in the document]
lievenhey: if I create relative paths with `-fdebug-prefix-map=$(pwd)=.` they are prefixed with `.`. clang and gcc prefix a relative path with a `.`.

lievenhey:
closes #9

GitMensch:
> closes #9

Note: for GitHub to recognize this please edit the first post and add the information there

src/models/disassemblymodel.cpp
```diff
@@ -6,11 +6,22 @@
...
+#include "highlighter.hpp"
+#include "sourcecodemodel.h"
 
-DisassemblyModel::DisassemblyModel(QObject *parent)
+DisassemblyModel::DisassemblyModel(QObject* parent)
     : QAbstractTableModel(parent)
+    , m_document(new QTextDocument(this))
+    , m_highlighter(new Highlighter(m_document, this))
 {
+    m_document->setDefaultFont(QFont(QStringLiteral("Monospace"), 10));
```
milianw, line 24: please use QFontDatabase::FixedFont here instead, don't hardcode the size

src/models/disassemblymodel.cpp
```diff
@@ -136,3 +173,14 @@ int DisassemblyModel::rowCount(const QModelIndex& parent) const
...
     return parent.isValid() ? 0 : m_data.disassemblyLines.count();
 }
+
+void DisassemblyModel::updateHighlighting(int line)
+{
+    m_highlightLine = line;
+    emit dataChanged(createIndex(0, Columns::DisassemblyColumn), createIndex(rowCount(), Columns::DisassemblyColumn));
+}
+
+int DisassemblyModel::lineForIndex(const QModelIndex& index)
```
milianw, line 183: make this method const

src/models/disassemblymodel.cpp
```diff
@@ -40,6 +51,18 @@ void DisassemblyModel::setDisassembly(const DisassemblyOutput& disassemblyOutput
 {
     beginResetModel();
     m_data = disassemblyOutput;
+
+    QString sourceCode;
```
milianw, line 55: feels ugly to join the lines just to let them be split up again in the document again. afair you can manually insert individual lines too via the QTextCursor API (a bit hidden but would be cool if you could have a look)

something like

```
    QTextCursor cursor(m_document);
    cursor.insertText(line);
```

potentially also `cursor.insertBlock` to split lines, unsure if that's needed

src/models/disassemblyoutput.cpp
```diff
@@ -47,15 +53,18 @@ DisassemblyOutput::LinkedFunction extractLinkedFunction(const QString& disassemb
...
-static QVector<DisassemblyOutput::DisassemblyLine> objdumpParse(const QByteArray &output)
+static ObjectdumpOutput objdumpParse(const QByteArray& output)
 {
     QVector<DisassemblyOutput::DisassemblyLine> disassemblyLines;
 
     QTextStream stream(output);
     QString asmLine;
+    QString sourceFileName;
     // detect lines like:
     // 4f616: 84 c0 test %al,%al
```
milianw, line 64: the comment needs to be updated now, as the below regex has changed?

src/models/highlighter.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
```
milianw, line 1: missing copyright

src/models/highlighter.cpp
```diff
@@ -0,0 +1,64 @@
+#include "highlighter.hpp"
```
milianw, line 1: missing copyright

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,200 @@
...
+void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
+{
+    m_numLines = 0;
+
+    if (disassemblyOutput.sourceFileName.isEmpty())
+        return;
+
+    QFile file(m_sysroot + QDir::separator() + disassemblyOutput.sourceFileName);
+
+    if (!file.open(QIODevice::ReadOnly))
```
milianw, line 43: `| QIODevice::Text` for windows

milianw (changes requested):

src/models/codedelegate.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    if (sourceLine) {
+        painter->setBrush(backgroundColor(sourceLine, index.data(m_highlightRole).toBool()));
+        painter->drawRect(option.rect);
+    }
+
+    painter->setPen(pen);
+    painter->setBrush(brush);
+
+    if (index.data().canConvert<QTextLine>()) {
```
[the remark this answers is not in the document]
lievenhey: this is a c++17 extension, so it is not possible
but I moved the `auto line =...` in front of the if

src/models/disassemblymodel.cpp
lievenhey: did the same for the sourcecodemodel

src/models/disassemblyoutput.cpp
lievenhey: not really, the old regex also validated the hex representation of instructions, but this is not necessary

(note: the change was by accident)

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,148 @@
...
+        return;
+
+    QFile file(m_sysroot + QDir::separator() + disassemblyOutput.sourceFileName);
+
+    if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
+        return;
+
+    beginResetModel();
+
+    const auto lines = QString::fromUtf8(file.readAll()).split(QLatin1Char('\n'));
```
milianw: move down to where it's used

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+        m_validLineNumbers.insert(line.sourceCodeLine);
+    }
+
+    m_sourceCode.clear();
+
+    // the start / end of a function is defined by { }
+    // the following code tries to find ( to include the function signature in the source code
+    if (!lines[minIndex - 1].contains(QLatin1Char('('))) {
+        for (int i = minIndex - 1; i > minIndex - 6; i--) {
```
milianw, line 90: merge the i < 0 check in here, i.e.:

```
for (int i = minIndex - 1; i >= 0 && i > minIndex - 6; --i) {
```

but what does the `-6` do? note how `minIndex--` is also changed in the body below (which is probably wrong?)

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    const auto lines = QString::fromUtf8(file.readAll()).split(QLatin1Char('\n'));
+
+    int maxIndex = 0;
+    int minIndex = INT_MAX;
+
+    m_validLineNumbers.clear();
+
+    for (const auto& line : disassemblyOutput.disassemblyLines) {
+        if (line.sourceCodeLine == 0)
```
milianw, line 65: either add braces here too or remove them below.

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        m_validLineNumbers.insert(line.sourceCodeLine);
+    }
+
+    m_sourceCode.clear();
+
+    // the start / end of a function is defined by { }
+    // the following code tries to find ( to include the function signature in the source code
+    if (!lines[minIndex - 1].contains(QLatin1Char('('))) {
```
milianw: if minIndex == 0 then this underflows

src/models/sourcecodemodel.cpp
```diff
@@ -0,0 +1,148 @@
...
+            if (i < 0)
+                break;
+            if (lines[minIndex - 1].contains(QLatin1Char('(')))
+                break;
+            minIndex--;
+        }
+    }
+
+    for (int i = minIndex - 1; i < maxIndex; i++) {
+        m_sourceCode.push_back(lines[i]);
```
milianw: underflows when minIndex = 0;

tests/modeltests/tst_models.cpp
```diff
@@ -297,7 +297,7 @@ private slots:
         DisassemblyModel model;
         QAbstractItemModelTester tester(&model);
         model.setResults(results);
-        QCOMPARE(model.columnCount(), 3 + results.selfCosts.numTypes()); // disassembly + linked funtion
+        QCOMPARE(model.columnCount(), DisassemblyModel::COLUMN_COUNT + results.selfCosts.numTypes());
```
milianw, line 300: this patch should be squashed into one of the previous ones?

tests/modeltests/vector_static_gcc/vector_static_gcc_v9.1.0.expected.txt
```diff
@@ -1,490 +1,490 @@
-4294544  418790:	f3 0f 1e fa          	endbr64 
-4294548  418794:	55                   	push   %rbp
-4294549  418795:	53                   	push   %rbx
-4294550  418796:	48 83 ec 28          	sub    $0x28,%rsp
-4294554  41879a:	64 48 8b 04 25 28 00 	mov    %fs:0x28,%rax
+4294544  418790:	                                           f3 0f 1e fa          	endbr64
```
milianw, line 1: should this be squashed into the previous patch?

CMakeLists.txt
```diff
@@ -75,6 +75,7 @@ find_package(KF5 COMPONENTS
 
 find_package(KF5Archive 5.85.0 CONFIG)
 find_package(KF5Auth CONFIG)
+find_package(KF5SyntaxHighlighting CONFIG)
```
milianw, line 78: adding the note here: please improve the commit message, at the very least give a summary of how you implement the syntax highlighting

src/models/sourcecodemodel.h
```diff
@@ -53,10 +61,13 @@ public slots:
 private:
     QString m_sysroot;
     QSet<int> m_validLineNumbers;
-    QStringList m_sourceCode;
+    QTextDocument* m_document;
```
milianw, line 64: = nullptr both

tests/modeltests/tst_models.cpp
```diff
@@ -653,6 +653,6 @@ private slots:
     }
 };
 
-QTEST_MAIN(TestModels);
+HOTSPOT_GUITEST_MAIN(TestModels);
```
milianw, line 656: squash into previous patch

milianw (changes requested):

src/models/sourcecodemodel.cpp
lievenhey: I have to rename that variable. Its not really an index, its the line number. Therefore it can't be 0. I will add some assertions to ensure this

src/models/sourcecodemodel.cpp
lievenhey: the -6 means, that hotspot will search up to 6 lines for a `(` to find the function declaration.

In hindsight this is stupid. It would be much better to check if the line before minIndex is empty, of not include it until there is an empty line.
The reason is, DWARF starts a function at the `{` and not at the function declaration, but I would like to include the declaration. The code I wrote is just working for c++ which is kind of stupid.
The new variant should be a bit more robust, since most developers add an empty line between functions to split up the file visually

milianw:
@lievenhey moin, this is looking good and we could merge it as-is as it's an improvement.

But we definitely need to handle inlining better, see https://github.com/KDAB/hotspot/pull/390/commits/72fec27e4ba517ac9a7f9245dc6cdc2ebca7faa2 for some easy work from towards that direction. Could you look into improving this further based on the `cpp-inlining` example? Things that I'd like to see:

- source listings for all files encountered in a function in a combined SourceCodeModel tree model, the file name on the first level and then second level shows source code lines like it does currently
- combined source line costs when code gets inlined from somewhere, this screenshot should hopefully make this clear:

Screenshot_20220828_130753

note how the code is basically all within inlined code, so currently we show 0% cost everywhere which is super misleading

milianw (approved):

GitMensch:
This is really cool! The CI builds don't provide an AppImage, do they? 'd love to test the new annotation and disassembly.

milianw:
Not yet, no - I'll build one manually next week (and do a proper new release too :crossed_fingers:)

GitMensch:
Just rechecking @milianw, does this mean we can expect Hotspot 1.4.0 "soonish"?

PR #392, opened 2022-06-24
Use KIO::CommandLauncherJob to open editors

nicolasfella:
This has several benefits

It makes lauch feedback work properly

It ensures that the lauched application is in the correct cgroup

It allows the called application to raise its window on Wayland and on X11 with focus stealing protection

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -37,6 +37,11 @@
 #include <KStandardAction>
 #include <KColorScheme>
 
+#include <kio_version.h>
+#if KIO_VERSION >= QT_VERSION_CHECK(5, 69, 0)
+#include <KIO/CommandLauncherJob>
+#endif
+
 #include <kddockwidgets/LayoutSaver.h>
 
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -528,9 +530,17 @@ void MainWindow::navigateToCode(const QString& filePath, int lineNumber, int col
             arg.replace(QLatin1String("%c"), QString::number(std::max(1, columnNumber)));
         }
 
-        if (!QProcess::startDetached(command, args)) {
-            m_resultsPage->showError(tr("Failed to launch command: %1 %2").arg(command, args.join(QLatin1Char(' '))));
-        }
+        auto *job = new KIO::CommandLauncherJob(command, args);
+        job->setDesktopName(QString::fromUtf8(ideSettings[ideIdx].desktopEntryName));
```
milianw, line 534: this won't work for custom commands, i.e. the ideIdx == -1 case will crash here

milianw (changes requested):

PR #395, opened 2022-07-20
improve eliding in the flamegraph

lievenhey:
in some cases the flamegraph contains only std::... and not the function
name.
This patch allows the user to change the eliding to ElideLeft so that
you can see the function name

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -13,6 +13,7 @@
 #include <QDebug>
 #include <QDir>
 #include <QFileInfo>
+#include <QFontMetrics>
 #include <QProcessEnvironment>
 #include <QStandardPaths>
 
@@ -31,6 +32,64 @@
 #endif
 
 #include <KParts/ReadOnlyPart>
+
+namespace {
+std::pair<QString, QString> elideArguments(const QString symbolText)
+{
+    QChar quoteChar = QLatin1Char('"');
+    int numQuotes = 0;
+    int depth = 0;
+
+    int endEliding = symbolText.length();
+    int startEliding = 0;
+
+    // this loop tries to to find the opening and closing bracket of the arguments
+    // it does this by counting the number of opening and closing brackets from the back
+    // it also has some basic string checking, so that string arguments like "(" of "\')"
+    // will not be counted
+    for (int i = symbolText.length() - 1; i > 0; i--) {
+        auto c = symbolText[i];
+
+        // detect quotes and skip those
+        if ((c == QLatin1Char('"') || c == QLatin1Char('\'')) && numQuotes == 0) {
+            quoteChar = c;
+            numQuotes++;
+        } else if (c == quoteChar && numQuotes == 1) {
+            if (symbolText[i - 1] == QLatin1Char('\\')) {
+                continue;
+            }
+            numQuotes = 0;
+            continue;
+        }
+        if (numQuotes == 1)
+            continue;
+
+        // mark opening and closing brackets
+        if (c == QLatin1Char(')')) {
+            if (depth == 0) {
+                endEliding = i;
+            }
+            depth++;
+        }
+        if (c == QLatin1Char('(')) {
+            depth--;
+            if (depth == 0) {
+                startEliding = i;
+                break;
+            }
+        }
+    }
+
+    if (endEliding - startEliding == 1) {
+        // don't elide if we have () in symbol
+        return {symbolText, {}};
+    }
+
+    // QLatin1String does not work with …
+    const auto elideSymbol = QStringLiteral("…");
+    return {symbolText.left(startEliding + 1), symbolText.right(symbolText.length() - endEliding)};
+}
+}
 
 QString collapseTemplate(const QString& str, int level)
 {
@@ -58,7 +117,7 @@ QString collapseTemplate(const QString& str, int level)
     };
 
     // QLatin1String does not work with …
-    const auto elideString = QString::fromUtf8("<…");
+    const auto elideString = QStringLiteral("<…");
 
     QString output;
     output.reserve(str.size());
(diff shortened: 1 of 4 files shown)


src/settings.cpp
```diff
@@ -218,7 +218,39 @@ void Settings::loadFromFile()
...
+
+    connect(this, &Settings::flamegraphTextElideModeChanged, [sharedConfig](const Qt::TextElideMode& mode) {
+        sharedConfig->group("Flamegraph").writeEntry("elideMode", elideLookup[mode]);
+    });
+}
+
+void Settings::setFlamegraphTextElideMode(const Qt::TextElideMode& mode)
+{
+    // we only support ElideLeft and ElideRight
+    Q_ASSERT(mode == Qt::ElideLeft || Qt::ElideRight);
```
milianw, line 250: this check is always true because Qt::ElideRight is always true

milianw (changes requested):
instead of making this configurable, I'd say we should make the elision more useful?

 Neither left nor right elision is ideal for absurd symbols like

```
asdf_namespace::foobar<asdf, yxcvyxcv>::blablub(someotherreallylongnames, ...)
```

ideally, we would detect such long names and then iteratively elide them, such that:

a) first we drop the function arguments and the above gets turned into
```
asdf_namespace::foobar<asdf, yxcvyxcv>::blablub(...)
```
b) if that's still too long, we left remove template arguments (we have that already, reuse that logic)
```
asdf_namespace::foobar<...>::blablub(...)
```
c) finally if that's still too long, we left-elide the remainder to produce e.g. (depending on width):
```
...::blablub(...)
```

ideally we also don't use three dots but the utf8 sequence for the ellision, i.e. `…`

tests/modeltests/tst_models.cpp
```diff
@@ -651,6 +651,33 @@ private slots:
...
+        QTest::addRow("elide templates") << 700 << "asdf_namespace::foobar<…>::blablub(…) const";
+        QTest::addRow("elide symbol") << 350 << "…obar<…>::blablub(…) const";
+    }
+
+    void testSymbolEliding()
+    {
+        const QFontMetrics metrics(QFont(QStringLiteral("monospace"), 10));
+
+        const QString testSymbol =
+            QLatin1String("asdf_namespace::foobar<asdf, yxcvyxcv>::blablub(someotherreallylongnames) const");
```
milianw, line 674: QStringLiteral

src/util.cpp
```diff
@@ -32,6 +33,28 @@
 
 #include <KParts/ReadOnlyPart>
 
+namespace {
+QString elideArguments(const QString symbolText)
```
milianw, line 37: const&

src/util.cpp
```diff
@@ -32,6 +33,28 @@
...
+{
+    int startEliding = symbolText.indexOf(QLatin1Char('('));
```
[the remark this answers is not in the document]
milianw: https://invent.kde.org/kdevelop/kdevelop/-/merge_requests/378 contains a bunch of other fancy identifiers that are hard to parse...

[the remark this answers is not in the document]
lievenhey: I added some more logic, so that hotspot is now able to elide those monstrosities

src/util.cpp
```diff
@@ -32,6 +33,66 @@
 
 #include <KParts/ReadOnlyPart>
 
+namespace {
+struct ElidedText
+{
+    QString first;
```
milianw, line 39: `std::pair<QString, QString>`? or use better names

src/util.cpp
```diff
@@ -32,6 +33,28 @@
...
+
+    int endEliding = symbolText.indexOf(QLatin1Char(')'), startEliding);
+
+    if (endEliding - startEliding == 1) {
+        // don't elide if we have () in symbol
+        return symbolText;
+    }
+
+    // QLatin1String does not work with …
+    const auto elideSymbol = QString::fromUtf8("…");
```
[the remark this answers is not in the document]
milianw: this is not resolved

src/util.cpp
```diff
@@ -57,6 +57,9 @@ QString collapseTemplate(const QString& str, int level)
         return false;
     };
 
+    // QLatin1String does not work with …
+    const auto elideString = QString::fromUtf8("<…");
```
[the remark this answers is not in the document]
milianw: not resolved

tests/modeltests/tst_models.cpp
milianw: not done

tests/modeltests/tst_models.cpp
```diff
@@ -651,6 +651,41 @@ private slots:
...
+        QFETCH(QString, elidedSymbol);
+
+        QCOMPARE(Util::elideSymbol(testSymbol, metrics, maxWidth), elidedSymbol);
+    }
+
+    void testSymbolElidingParanthese()
+    {
+        const QFontMetrics metrics(QFont(QStringLiteral("monospace"), 10));
+
+        const QString symbol = QLatin1String("Foo<&bar::operator()>::asdf<XYZ>(blabla<&foo::operator(), ')', '('>)");
```
milianw, line 685: dito

src/util.cpp
```diff
@@ -32,6 +33,66 @@
...
+ElidedText elideArguments(const QString symbolText)
+{
+    QChar quoteChar = QLatin1Char('"');
+    int numQuotes = 0;
+    int depth = 0;
+
+    int endEliding = symbolText.length();
+    int startEliding = 0;
+
+    for (int i = symbolText.length() - 1; i > 0; i--) {
```
milianw, line 52: please add some basic comments that document what this code is doing on a high level

milianw (changes requested):

milianw (approved):
lgtm, thanks

PR #396, opened 2022-07-29
fix startpage layout

lievenhey:
the integration of the debuginfod download progress broke the layout of
the start page
this fix restores the layout


milianw:
thanks

PR #399, opened 2022-09-15
add an improved color scheme for the flamegraph

lievenhey:
the new color scheme selects the color based on the cost in % an very expensive function is colored red, a cheap function is colored yellow

### DIFF
--- a/src/flamegraph.cpp
+++ b/src/flamegraph.cpp
@@ -365,7 +365,16 @@ QBrush brushSystem(const Data::Symbol& symbol)
     return user;
 }
 
-QBrush brush(const Data::Symbol& entry, Settings::ColorScheme scheme)
+QBrush costRatioBrush(quint32 cost, quint32 totalCost)
+{
+    // interpolate between red and yellow
+    // where yellow = 0% and red = 100%
+    float ratio = (1 - static_cast<float>(cost) / totalCost);
+    int hue = ratio * ratio * 60;
+    return QColor::fromHsv(hue, 230, 200, 125);
+}
+
+QBrush brush(const Data::Symbol& entry, Settings::ColorScheme scheme, quint32 cost = 0, quint32 totalCost = 1)
 {
     switch (scheme) {
     case Settings::ColorScheme::Binary:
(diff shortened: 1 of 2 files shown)


lievenhey:
grafik

src/flamegraph.cpp
```diff
@@ -365,7 +365,16 @@ QBrush brushSystem(const Data::Symbol& symbol)
     return user;
 }
 
-QBrush brush(const Data::Symbol& entry, Settings::ColorScheme scheme)
+QBrush improvedBrush(quint32 cost, quint32 totalCost)
```
milianw, line 368: "improved" is very subjective. I don't think it improves anything. please give it a different name, something like `brushByCostRatio` or similar

src/flamegraph.cpp
```diff
@@ -713,6 +724,7 @@ FlameGraph::FlameGraph(QWidget* parent, Qt::WindowFlags flags)
     m_colorSchemeSelector->addItem(tr("Binary"), QVariant::fromValue(Settings::ColorScheme::Binary));
     m_colorSchemeSelector->addItem(tr("Kernel"), QVariant::fromValue(Settings::ColorScheme::Kernel));
     m_colorSchemeSelector->addItem(tr("System"), QVariant::fromValue(Settings::ColorScheme::System));
+    m_colorSchemeSelector->addItem(tr("Improved"), QVariant::fromValue(Settings::ColorScheme::Improved));
```
milianw, line 727: Cost Ratio

src/settings.h
```diff
@@ -23,6 +23,7 @@ class Settings : public QObject
         Binary,
         Kernel,
         System,
+        Improved,
```
milianw, line 26: CostRatio

milianw (changes requested):
I'm fine with adding it, but not under this name. it doesn't improve anything, it's just a different way to colorize it - and note how it essentially duplicates information that we already represent in the flamegraph: the width already encodes the cost ratio, so doing it again with the color... what for?

milianw:
please rebase, the compile errors should be gone

lievenhey:
done

milianw (approved):

PR #402, opened 2022-09-29
Start recording directly if privileges are already elevated

ahmadsamir:
Now you can use sudo to run the commands manually in terminal and use hotspot directly without having to re-enter the root password in the polkit dialog.

### DIFF
--- a/src/perfrecord.cpp
+++ b/src/perfrecord.cpp
@@ -38,6 +38,9 @@
 #endif
 #endif
 
+#include <sys/stat.h>
+#include <fstream>
+
 PerfRecord::PerfRecord(QObject* parent)
     : QObject(parent)
     , m_perfRecordProcess(nullptr)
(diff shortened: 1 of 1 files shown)


src/perfrecord.cpp
```diff
@@ -71,10 +74,47 @@ static QStringList sudoOptions(const QString& sudoBinary)
     return options;
 }
 
+static bool privsAlreadyElevated()
+{
+    std::ifstream ifs;
```
milianw, line 79: add a lambda that reads the value:

```
auto readValue = [](const char *path) -> std::optional<int> {
    std::ifstream ifs;
    ifs.open(path);
    if (!ifs) {
        return std::nullopt;
    }
    int i = 0;
    ifs >> i;
    return i;
};

if (readValue("/proc/sys/kernel/kptr_restrict") != 0 || readValue("/proc/sys/kernel/perf_event_paranoid") != -1) {
    return false;
}
```

src/perfrecord.cpp
```diff
@@ -71,10 +74,47 @@ static QStringList sudoOptions(const QString& sudoBinary)
...
+        isElevated &= i == -1;
+    }
+    if (!isElevated) {
+        return false;
+    }
+
+    const mode_t required = S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH; // 755
+    struct stat buf;
+    auto checkPerms = [required, &buf](const char* path) {
+        return stat(path, &buf) == 0 //
```
milianw, line 105: why that `//` comment? just let clang-format do its job please

src/perfrecord.cpp
```diff
@@ -71,10 +74,47 @@ static QStringList sudoOptions(const QString& sudoBinary)
...
+    isElevated &= std::all_of(paths.begin(), paths.end(), checkPerms);
+
+    return isElevated;
+}
+
 void PerfRecord::startRecording(bool elevatePrivileges, const QStringList& perfOptions, const QString& outputPath,
                                 const QStringList& recordOptions, const QString& workingDirectory)
 {
-    if (elevatePrivileges && geteuid() != 0) {
+    if (elevatePrivileges && !privsAlreadyElevated() && geteuid() != 0) {
```
milianw, line 117: check the geteuid first as it's cheaper

milianw (changes requested):
cool idea, but needs some more work first

src/perfrecord.cpp
ahmadsamir: Isn't it less probable that one would be running a gui as root?

src/perfrecord.cpp
ahmadsamir: OK.

src/perfrecord.cpp
```diff
@@ -71,10 +75,51 @@ static QStringList sudoOptions(const QString& sudoBinary)
     return options;
 }
 
+static bool privsAlreadyElevated()
+{
+    struct ValueMap {
```
milianw, line 80: while it's ok, I do think it's pretty overkill as we only have these two values to check... it's more code to make it this generic over just calling the lambda twice, no?

src/perfrecord.cpp
```diff
@@ -71,10 +75,51 @@ static QStringList sudoOptions(const QString& sudoBinary)
...
+        const char *path = nullptr;
+        int value;
+    };
+    static const std::array<ValueMap, 2> map = {{
+        {"/proc/sys/kernel/kptr_restrict", 0},
+        {"/proc/sys/kernel/perf_event_paranoid", -1},
+    }};
+
+    std::ifstream ifs;
```
milianw, line 89: just move this into the lambda, then you can remove the `ifs.is_open` - or why do you want to reuse it? I don't see any reason for that

milianw (changes requested):

src/perfrecord.cpp
ahmadsamir: So as not to construct two objects of the same type when I can re-use one object.

milianw (approved):
lgtm, thanks

milianw:
manually resolved the clang-format issues and pushed it now, many thanks @ahmadsamir !

PR #407, opened 2022-10-09
Fix segfault in SourceCodeModel::headerData

### LINKED ISSUE #406: disassemble causing a segmentation fault
**Describe the bug**
segmentation fault when you try to disassemble a symbol mapped to a source file that cannot be opened. has to be the first symbol to disassemble.

**To Reproduce**
```
#include <iostream>

int this_simple_function(int x) {
	for (int i = 0; i < 100; ++i) {
		std::cout << "hello" << std::endl;
	}
	return 2;
}

int main() {
	this_simple_function(5);
	return 0;
}
```
```
g++ -g main.cpp
```
```
perf record --call-graph=dwarf ./a.out
```

- open hotspot
- go to flame graph view
- disassemble from right click context menu on e.g. `std::ostream::put(char)`
- segfault

**Expected behavior**
anything but segfault

**Version Info (please complete the following information):**

- Linux Kernel version: 5.19.7-1-MANJARO
- perf version: perf version 5.19.g3d7cb6b04c3f
- hotspot version (appimage? selfcompiled?): self-compiled, 64d2d49a9c25f9c526eb1589d024c13b292ee6dd
- if self-compiled hotspot, what version of elfutils: 0.187

**Additional context**
```cpp
(gdb) bt                                                                                                                                                                                                                                
#0  0x00007ffff50a249c in ?? () from /usr/lib/libc.so.6                                                                                                                                                                                 
#1  0x00007ffff5052958 in raise () from /usr/lib/libc.so.6
#2  0x00007ffff503c53d in abort () from /usr/lib/libc.so.6
#3  0x00007ffff569fede in QMessageLogger::fatal(char const*, ...) const () from /usr/lib/libQt5Core.so.5
#4  0x00007ffff569ff9d in qt_assert_x(char const*, char const*, char const*, int) () from /usr/lib/libQt5Core.so.5
#5  0x0000555555605bdc in QVector<QString>::operator[] (this=0x555555ea8e28, i=0) at /usr/include/qt/QtCore/qvector.h:453
#6  0x0000555555604f46 in Data::Costs::typeName (this=0x555555ea8e28, type=0) at /home/dante/projects/hotspot/src/models/data.h:206
#7  0x0000555555695d8b in SourceCodeModel::headerData (this=0x555555ea8db0, section=2, orientation=Qt::Horizontal, role=0) at /home/dante/projects/hotspot/src/models/sourcecodemodel.cpp:117
#8  0x00007ffff67ef171 in QHeaderView::sectionSizeFromContents(int) const () from /usr/lib/libQt5Widgets.so.5
#9  0x00007ffff67ead1d in QHeaderView::sizeHint() const () from /usr/lib/libQt5Widgets.so.5
#10 0x00007ffff68565ce in QTreeView::updateGeometries() () from /usr/lib/libQt5Widgets.so.5
#11 0x00007ffff67da581 in QAbstractItemView::doItemsLayout() () from /usr/lib/libQt5Widgets.so.5
#12 0x00007ffff684b703 in QTreeView::doItemsLayout() () from /usr/lib/libQt5Widgets.so.5
#13 0x00007ffff6853412 in QTreeView::sizeHintForColumn(int) const () from /usr/lib/libQt5Widgets.so.5
#14 0x00007ffff67efb86 in ?? () from /usr/lib/libQt5Widgets.so.5
#15 0x00007ffff67f0264 in QHeaderView::sectionPosition(int) const () from /usr/lib/libQt5Widgets.so.5
#16 0x00007ffff67f02df in QHeaderView::sectionViewportPosition(int) const () from /usr/lib/libQt5Widgets.so.5
#17 0x00007ffff67f06ba in QHeaderView::resizeSection(int, int) () from /usr/lib/libQt5Widgets.so.5
#18 0x0000555555643ec1 in ResultsDisassemblyPage::setupAsmViewModel (this=0x555555e13a90) at /home/dante/projects/hotspot/src/resultsdisassemblypage.cpp:130
#19 0x00005555556444b8 in ResultsDisassemblyPage::showDisassembly (this=0x555555e13a90, disassemblyOutput=...) at /home/dante/projects/hotspot/src/resultsdisassemblypage.cpp:181
#20 0x00005555556442ef in ResultsDisassemblyPage::showDisassembly (this=0x555555e13a90) at /home/dante/projects/hotspot/src/resultsdisassemblypage.cpp:155
#21 0x0000555555626efd in ResultsPage::onJumpToDisassembly (this=0x555555c3e1d0, symbol=...) at /home/dante/projects/hotspot/src/resultspage.cpp:226
...
```

DanielChabrowski:
resolves #406 

I tracked the issue down to the:
- `SourceCodeModel::setCallerCalleeResults` setting `m_numTypes` to `1`
- file fails to open in `SourceCodeModel::setDisassembly` and we return early
- since we return early we never call `m_costs.initializeCostsFrom(m_callerCalleeResults.selfCosts);`
- therefore `m_costs.m_typeNames` is empty
- call to `return m_costs.typeName(section - COLUMN_COUNT);` segfaults

### DIFF
--- a/src/models/sourcecodemodel.cpp
+++ b/src/models/sourcecodemodel.cpp
@@ -9,6 +9,7 @@
 
 #include <QDir>
 #include <QFile>
+#include <QScopeGuard>
 #include <QTextBlock>
 #include <QTextDocument>
 
@@ -31,10 +32,14 @@ void SourceCodeModel::clear()
     endResetModel();
 }
 
-void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
+void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
+                                     const Data::CallerCalleeResults& results)
 {
-    m_numLines = 0;
+    beginResetModel();
+    auto guard = qScopeGuard([this]() { endResetModel(); });
+
     m_numTypes = 0;
+    m_numLines = 0;
 
     if (disassemblyOutput.mainSourceFileName.isEmpty())
         return;
@@ -44,16 +49,13 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
     if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
         return;
 
-    beginResetModel();
-
     int maxLineNumber = 0;
     int minLineNumber = INT_MAX;
 
     m_validLineNumbers.clear();
 
-    auto entry = m_callerCalleeResults.entry(disassemblyOutput.symbol);
     m_costs = {};
-    m_costs.initializeCostsFrom(m_callerCalleeResults.selfCosts);
+    m_costs.initializeCostsFrom(results.selfCosts);
 
     m_numTypes = m_costs.numTypes();
 
@@ -76,13 +78,15 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
             minLineNumber = line.sourceCodeLine;
         }
 
-        auto it = entry.offsetMap.find(line.addr);
-        if (it != entry.offsetMap.end()) {
-            const auto& locationCost = it.value();
-            const auto& costLine = locationCost.selfCost;
-            const auto totalCost = m_callerCalleeResults.selfCosts.totalCosts();
+        const auto entry = results.entries.find(disassemblyOutput.symbol);
+        if (entry != results.entries.end()) {
+            const auto it = entry->offsetMap.find(line.addr);
+            if (it != entry->offsetMap.end()) {
+                const auto& locationCost = it.value();
+                const auto& costLine = locationCost.selfCost;
 
-            m_costs.add(line.sourceCodeLine, costLine);
+                m_costs.add(line.sourceCodeLine, costLine);
+            }
         }
 
         m_validLineNumbers.insert(line.sourceCodeLine);
@@ -99,8 +103,6 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
     cursor.select(QTextCursor::SelectionType::LineUnderCursor);
     cursor.removeSelectedText();
     cursor.insertText(disassemblyOutput.symbol.prettySymbol);
-
-    endResetModel();
 }
 
 QVariant SourceCodeModel::headerData(int section, Qt::Orientation orientation, int role) const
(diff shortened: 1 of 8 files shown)


src/models/sourcecodemodel.cpp
```diff
@@ -34,6 +34,7 @@ void SourceCodeModel::clear()
 void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput)
 {
     m_numLines = 0;
```
milianw, line 36: this was probably wrong before, but we need to `beginResetModel` early and ensure that we always call `endResetModel` then too. I suggest we rewrite this by leveraging `qScopeGuard`:

```
beginResetModel();
auto guard = qScopeGuard([this]() { endResetModel(); });

m_numLines = 0;
m_numTypes = 0;
...
```

src/models/sourcecodemodel.cpp
```diff
@@ -196,6 +199,5 @@ void SourceCodeModel::setCallerCalleeResults(const Data::CallerCalleeResults& re
 {
     beginResetModel();
```
milianw, line 200: I guess this doesn't actually influence the model at all, right? so we should actually remove the `{begin,end}ResetModel` here now - only the calls to `SourceCodeModel::setDisassembly` have an effect. I'd actually go as far as merging the two methods into one, to ensure that one cannot set one without the other.

milianw (changes requested):
great, thanks for the investigation! can we get a unit test too please?

src/models/sourcecodemodel.cpp
DanielChabrowski: Done.

src/models/sourcecodemodel.cpp
DanielChabrowski: Merged the two methods for both `SourceCodeModel` and `DisassemblyModel`. In case of `SourceCodeModel` we no longer need to copy `Data::CallerCalleeResults` after replacing
`auto entry = m_callerCalleeResults.entry(disassemblyOutput.symbol);`
with
`const auto entry = results.entries.find(disassemblyOutput.symbol);`.

The change here is that `entry` creates an empty symbol if it's not found and in this case we would quickly discard it since `entry->offsetMap` would be empty.

DanielChabrowski:
> great, thanks for the investigation! can we get a unit test too please?
I did write one, or more like copy one from DisassemblyModel test, for a slightly different case - for when the fileName is empty. We get the same early return nonetheless.

I can't say I'm comfortable enough with QTest and the data formats used here to provide more though.

On a side note, I couldn't actually see the text in disassembly tab until I made the changes related to the pen and the brush. My assumption here is that we weren't restoring QPen to previous state since we weren't making a copy before changes.

milianw:
Thank you @DanielChabrowski, I have landed your changes here now. Regarding the rendering issues, please open a new bug report and pull request once you have figured out the reason for that - sounds odd! What kind of color scheme and platform do you run hotspot on?

PR #408, opened 2022-10-18
fix crash in SourceCodeModel::headerData

lievenhey:
m_costs was not initialized if no source code was found this caused a crash in headerData when m_costs.typeNames[0] was accessed

fixes: #406


lievenhey:
ah, this is already fixed in #407

milianw:
let's take the other pull request then, as it is more thorough

PR #409, opened 2022-10-20
Improve CMake code for finding kgraphviewer

dfaure-kdab:
* Look for the library (part) not for the executable
* Add information about it for FeatureSummary
* While at it, describe better the QCustomPlot dependency


milianw:
thanks!

PR #412, opened 2022-11-08
Fix cost aggregation

### LINKED ISSUE #403: incorrect aggregation base in bottom down view when excluding threads
**Describe the bug**
Before the recent 596f5e3759, the aggregate by thread didn't work at all in my testing. Now it does create separate columns, using the thread name, awesome! However, as soon as you exclude one thread in the bottom event view, it seems like the top function (likely where the execution was at the moment of perf record sampling) actually shows up in the bottom as the base!

So basically when selecting 
Suppose I have a `main()` which calls `myFunc1()`, which calls `myFunc2()`, which calls `myFunc3()`, I get stacks that look like this:

- myFunc2()
- myFunc1()
- main()
- __libc_start_call_main
- __SI__libc_start_main_alias_1_0
- _start
- myFunc3()

**To Reproduce**

1. Perf record a multi-threaded program
2. select "Aggregate cost by: threads"
3. right click on a thread in the bottom event view, then "Exclude Thread #12345"4. 
4. Inspect the call stack

**Expected behavior**
The thread name should should still appear at the bottom, and the right function (e.g. `myFunc3()`) should still be at the top.

**Screenshots**
image
image

**Version Info (please complete the following information):**

- Linux Kernel version: in docker desktop on macOS, Linux eb1328965548 5.10.124-linuxkit 
- perf version: linux-tools-generic: Installed: 5.19.0.15.15
- hotspot version: self compiled 07825e6
- elfutils version: self compiled, 6cb01b4ef3cdd4f83d21bbcec72fa66712ac14e6

lievenhey:
fixes #403

### DIFF
--- a/src/parsers/perf/perfparser.cpp
+++ b/src/parsers/perf/perfparser.cpp
@@ -552,6 +552,32 @@ void addCallerCalleeEvent(const Data::Symbol& symbol, const Data::Location& loca
     }
 }
 
+template<typename FrameCallback>
+void addBottomUpResult(Data::BottomUpResults* bottomUpResult, Settings::CostAggregation costAggregation,
+                       const QHash<qint32, QHash<qint32, QString>>& commands, int type, quint64 cost, qint32 pid,
+                       qint32 tid, quint32 cpu, const QVector<qint32>& frames, const FrameCallback& frameCallback)
+{
+    switch (costAggregation) {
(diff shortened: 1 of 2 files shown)


lievenhey:
I am not sure where to place this code (currently it is a duplicate from PerfParserPrivate::addToBottomUpResult).

PR #413, opened 2022-11-08
Fix ki18n initialisation

CarlSchwan:
Otherwise we get the following error mesage: kf.i18n: KLocalizedString: Using an empty domain, fix the code. msgid:

@albert-astals-cid-kdab ;)


milianw:
thanks @CarlSchwan !

PR #415, opened 2022-11-11
don't reparse perf.data when changing cost aggregation

### LINKED ISSUE #387: don't reparse perf.data when changing cost aggregation

lievenhey:
closes #387

### DIFF
--- a/src/parsers/perf/perfparser.cpp
+++ b/src/parsers/perf/perfparser.cpp
@@ -1393,6 +1393,7 @@ PerfParser::PerfParser(QObject* parent)
     : QObject(parent)
     , m_isParsing(false)
     , m_stopRequested(false)
+    , m_costAggregationChanged(false)
 {
     qRegisterMetaType<Data::Summary>();
     qRegisterMetaType<Data::BottomUp>();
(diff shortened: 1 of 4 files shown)


src/parsers/perf/perfparser.cpp
```diff
@@ -1674,7 +1676,7 @@ void PerfParser::filterResults(const Data::FilterAction& filter)
         const bool excludeByBinary = !filter.excludeBinaries.isEmpty();
         const bool filterByStack = includeBySymbol || excludeBySymbol || includeByBinary || excludeByBinary;
 
-        if (!filter.isValid()) {
+        if (!filter.isValid() && !m_costAggregationChanged) {
```
milianw, line 1679: this is not threadsafe, you need to copy it into the lambda capture instead of accessing the bool directly here

src/parsers/perf/perfparser.cpp
```diff
@@ -1862,6 +1864,8 @@ void PerfParser::filterResults(const Data::FilterAction& filter)
         emit tracepointDataAvailable(tracepointResults);
         emit eventsAvailable(events);
         emit parsingFinished();
+
+        m_costAggregationChanged = false;
```
milianw, line 1868: dito, this can race

I suggest you instead add a signal with the last-used cost aggregation, then store that in a member and instead of the bool flag you then compare if the last-used cost aggregation is different to what is now requested

milianw (changes requested):

milianw (approved):

PR #420, opened 2022-11-14
output full objdump command on empty output

GitMensch:
currently we get a warning / note with no details; adding those allows to execute it outside of hotspot and check for possible issues with that invocation

fair warning: this change is untested (other than CI passing); and if there's a testcase for the message it needs to be adjusted

### DIFF
--- a/src/models/disassemblyoutput.cpp
+++ b/src/models/disassemblyoutput.cpp
@@ -175,7 +175,8 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
     }
 
     if (output.isEmpty()) {
-        disassemblyOutput.errorMessage += QApplication::tr("Empty output of command %1").arg(objdump);
+        disassemblyOutput.errorMessage +=
+            QApplication::tr("Empty output of command %1 %2").arg(objdump, arguments.join(QLatin1Char(' ')));
     }
 
     const auto objdumpOutput = objdumpParse(output);


GitMensch:
now verified via CI built AppImage (that's quite comfortable btw) that this does neither fail building nor a test, but does show the full command line which can also be copy+pasted (that would have speed up the analysis for #422 - and is likely useful to debug any issues there may be with the call in any case [even better would be a tooltip over the message with the complete output, but as this can be seen when executing it via command line I don't think that's necessary]).

src/models/disassemblyoutput.cpp
```diff
@@ -176,6 +176,7 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
 
     if (output.isEmpty()) {
         disassemblyOutput.errorMessage += QApplication::tr("Empty output of command %1").arg(objdump);
```
milianw, line 178: this is now a string puzzle which makes localization hard/impossible for some languages

instead, write it like so:

```
disassemblyOutput.errorMessage += QApplication::tr("Empty output of command %1 %2").arg(objdump, arguments.join(QLatin1Char(' ')));
```

milianw (changes requested):

src/models/disassemblyoutput.cpp
GitMensch: Thanks for the review - that makes much more sense!

milianw (approved):
thanks, lgtm

PR #421, opened 2022-11-14
use messageType warning for disassembly issues

GitMensch:
making it identical to resultspageui where it was added in 002d0e1b325b738dfbe48c65c9f97f0106ee8b81

again: untested (just verified that the result is valid XML)

Note: the widget in resultspageui is animated on show, which isn't the case for resultsdisassemblypage (at least with x-forwarding the animation stutters, so I like the one without more); whatever "the way it should be done" is - it is likely a good idea to use the same approach for both widgets.

### DIFF
--- a/src/resultsdisassemblypage.ui
+++ b/src/resultsdisassemblypage.ui
@@ -24,7 +24,11 @@
     <number>0</number>
    </property>
    <item>
-    <widget class="KMessageWidget" name="errorMessage"/>
+    <widget class="KMessageWidget" name="errorMessage">
+     <property name="messageType">
+      <enum>KMessageWidget::Warning</enum>
+     </property>
+    </widget>
    </item>
    <item>
     <widget class="KSqueezedTextLabel" name="filenameLabel">


src/resultsdisassemblypage.ui
```diff
@@ -24,7 +24,14 @@
     <number>0</number>
    </property>
    <item>
-    <widget class="KMessageWidget" name="errorMessage"/>
+    <widget class="KMessageWidget" name="errorMessage">
+     <property name="wordWrap">
+      <bool>false</bool>
```
milianw, line 29: I would think disabling wordwrap could lead to odd sized windows for long error message. can you please enable it?

milianw (changes requested):
there's a typo in your commit message (user -> use)

furthermore, enabling animations would be easy - that should be just another properly flag, no?

finally: if the other message widget also doesn't disables wordwrap, please (in a follow-up patch) enable it for both.

src/resultsdisassemblypage.ui
GitMensch: So should I just drop the wordwrap here or explicit enable it or explicit enable it in both (it is explicit disabled in the other place)?

milianw: in a second patch, explicitly enable it in both

GitMensch: If you want that in a separate patch instead of a separate commit: sure, will do that (also would do the second commit here)

GitMensch: > furthermore, enabling animations would be easy - that should be just another properly flag, no?

The other parts do it via emitting an event that does this via code https://github.com/KDAB/hotspot/commit/002d0e1b325b738dfbe48c65c9f97f0106ee8b81#diff-bb8df26e05a17485c12aade282ff1bcfe991825bf350fbf75697310a6cc35a25R265-R270 - something I don't feel comfortable to do or adjust currently

milianw: a patch and a commit is the same, I don't understand how they could be separate? I did not mean a separate merge request like you did not :)

regarding the animation: ok, ignore that for now then.

milianw (approved):

Issue #440, opened 2022-12-02
command line option or - preferable - a separate tool to convert perf.data to .perfparser


GitMensch:
**Is your feature request related to a problem? Please describe.**
Recorded perf data is not unlikely to be "out of date" when the application/os changes.
I was therefore happy to see the following entry in the 1.4 release notes:
> you can export and import data from hotspot in a custom format that is efficient to load and self-contained, meaning exported data can be read on any other machine with hotspot and does not require access to the original client application code for unwinding anymore

When opening Hotspot and opening a trace, I do see  "Save As" ... `*.perfparser`, which I _assume_ to be this new format.
Using "File Open" does not contain a dropdown entry for `*.perfparser`, but I guess this would be the correct way to open it.

As I'd like to "directly convert" after recording a command line option for that would be very useful.

**Describe the solution you'd like**
I'd _like_ a solution that works even without any X initialization, enabling conversion even in headless / CI systems. I _guess_ that would mean that the main hotspot binary cannot be used for that, but I don't know.

**Describe alternatives you've considered**
Open by hand, save by hand; but that cannot be done automated.

**Additional context**
I'd guess that for the Disassembly view the "original client application" binaries and code must be still available, no? Are there other places where this would be necessary? Could I copy those over to a second machine and just reference them in the second instance?

GitMensch:
It is in code-wise, can you please add some docs on this, too?

milianw:
What is needed outside the `--help` output?

GitMensch:
Find a good place in the README to put the output of `--help` to would be fine.

milianw:
> Find a good place in the README to put the output of `--help` to would be fine.

patches welcome

GitMensch:
OK, patches will come :-)

What do you think of splitting out README.md into HACKING.md (the build instructions "from source" and "debugging Appimage", linked from the main README)?

milianw:
yes sounds like a good idea to me

PR #443, opened 2022-12-03
perfdata added to file open list

GitMensch:
along with a group that includes it and moving out to a separate function again

fixing the side issue mentioned in #440

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -177,8 +177,7 @@ MainWindow::MainWindow(QWidget* parent)
     auto openNewWindow = new QAction(QIcon::fromTheme(QStringLiteral("document-open")), tr("Open in new window"), this);
     openNewWindow->setShortcut(Qt::Key_O | Qt::ControlModifier | Qt::ShiftModifier);
     connect(openNewWindow, &QAction::triggered, this, [this] {
-        const auto fileName = QFileDialog::getOpenFileName(this, tr("Open File"), QDir::currentPath(),
-                                                           tr("Data Files (perf*.data perf.data.*);;All Files (*)"));
+        const auto fileName = queryOpenDataFile();
         if (!fileName.isEmpty())
             openInNewWindow(fileName);
     });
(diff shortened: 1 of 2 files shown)


src/mainwindow.cpp
```diff
@@ -279,10 +278,18 @@ void MainWindow::closeEvent(QCloseEvent* event)
     QMainWindow::closeEvent(event);
 }
 
+QString MainWindow::getOpenDataFileName()
```
milianw, line 281: please remove the `get` prefix

src/mainwindow.cpp
```diff
@@ -279,10 +278,18 @@ void MainWindow::closeEvent(QCloseEvent* event)
...
+{
+    const QString filter = tr("Hotspot data Files (perf*.data perf.data.* *.perfparser);;"
```
milianw, line 283: I'd prefer just `auto filter = ...`

src/mainwindow.cpp
milianw: and make this method const

milianw (changes requested):
please reword your commit message, it's pretty confusing. it would be better to clearly state that you are adding the `*.perfdata` pattern to the file open filter list

src/mainwindow.cpp
```diff
@@ -279,10 +278,18 @@ void MainWindow::closeEvent(QCloseEvent* event)
     QMainWindow::closeEvent(event);
 }
 
+const QString MainWindow::openDataFileName()
```
milianw, line 281: const the method, not the return type

and I think it would be better to rename this function to `queryOpenDataFile`

milianw (changes requested):

src/mainwindow.cpp
GitMensch: renaming sounds good - how do you want the signature to be - and how is this function different to other utility functions in this source file?

milianw:
```
QString MainWindow::queryOpenDataFile() const
```

I don't quite get what you mean by that question? are you saying there are utility functions that can be const which are not marked as const? fix that then please

milianw (approved):
code lgtm, let's wait for the CI and then this can be merged

milianw:
Argh sorry, it cannot be const b/c we need to pass `this` as a non-const `QWidget *` parent... sorry please remove the `const` again!

.gitpod.yml
```diff
@@ -0,0 +1,20 @@
+image: workspace-c
```
milianw, line 1: unrelated file?

milianw (changes requested):

.gitpod.yml
GitMensch: totally

milianw (approved):

PR #444, opened 2022-12-03
switch to nondeprecated function

GitMensch:
KWindowSystem::activeWindow() to KX11Extras::activeWindow()

### DIFF
--- a/src/perfrecord.cpp
+++ b/src/perfrecord.cpp
@@ -20,7 +20,7 @@
 #include <unistd.h>
 
 #include <KUser>
-#include <KWindowSystem>
+#include <kx11extras.h>
 
 #include "util.h"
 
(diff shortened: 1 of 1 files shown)


GitMensch:
I've just followed the warnings and then the docs here and tested the build via GH actions as I don't have the necessary setup locally.
I guess for KX11extras there is another cmake dependency to be added, too?

milianw:
as indicated by the failures, this has to be made optional while keeping backwards compatibility with older distros

GitMensch:
I'm not sure what the KDE version numbering says and what the minimal supported version is.
If "optional" I _guess_ that means an optional module via cmake that adds a `define` and then use a conditional compile for the include and the use?

GitMensch:
Can you please specify how this would be handled?

X11Extras is mandatory for KDDockWidgets and that is mandatory for Hotspot now. It looks to me that either the docking must be optional to support older distros or X11Extras should just be added to the module list (which I'll do for this PR later, now that there's an environment where I can test the compilation outside of the CI).

milianw:
We should detect the kde frameworks version and, if it's new enough, include the new header and otherwise stick to the old code and approach

GitMensch:
Maybe I don't understand or wasn't clear:

* currently a developer that builds Hotspots needs to build KDDockWidgets before
* a developer that builds KDDockWidgets _must_ have X11Extras "dev" installed

Therefore the question was "is there any reason not to add the X11Extras dependency" in Hotspot?
I _seems_ to me that making it optional would only be reasonable when at lease one of the following is true:
* X11Extras is optional "someday" in KDDockWidgets (I could create an issue for that, if you want me to)
* KDDockWidgets is optional "someday" in Hotspot

No?

milianw:
I don't follow - what has KDDockWidgets to do with this? That one depends on *Q*X11Extras but here we are talking about *K*X11Extras?

Issue #339, opened 2021-10-27
minor glitch in Disassembly view


GitMensch:
![contained &](https://user-images.githubusercontent.com/6699539/139075047-28f6168b-b08c-4b87-a6bd-473b8319a5aa.png)

the title has a contained `&`.

Hm, this isn't shown in the Dec2020 screenshot https://raw.githubusercontent.com/KDAB/hotspot/master/screenshots/dockwidgets.png is that a new issue or only happening on this machine?

milianw:
no input since many months, closing

GitMensch:
Wherever it is _recognized_ the acceleration key works fine (menu, "within the view", ...) but not for separate "windows"

![sample](https://user-images.githubusercontent.com/6699539/205068302-d62e5284-a345-45c7-8c3f-508d01df57fe.png)

lievenhey:
I can confirm this issue on 5.26.3 with wayland
happens after moving a dockwidget in its own window
![grafik](https://user-images.githubusercontent.com/82457690/205939595-56912e24-2318-43a7-8e8d-77b65783cf4d.png)

GitMensch:
So that's actually a QT issue? [note: the setting above does not change anything on the gitpod or local builds using x11 forwarding]

iamsergio:
I think we reached the conclusion this was a Qt bug (not with all styles), with the implementation of drawing QStyle::CE_DockWidgetTitle

GitMensch:
Thanks for the note - is there a workaround?

iamsergio:
Bug should be reported to Qt. Here's a minimal repro without dock widgets

```
#include <QtWidgets>

// These work:
// ./qtwidgets_basic -style Breeze
// ./qtwidgets_basic -style Oxygen

// This one doesn't:
// ./qtwidgets_basic -style Fusion

class MyWidget : public QWidget {
public:
  void paintEvent(QPaintEvent *ev) override {
    QPainter p(this);

    QStyleOptionDockWidget titleOpt;
    titleOpt.initFrom(this);
    titleOpt.title = "&My Title";

    style()->drawControl(QStyle::CE_DockWidgetTitle, &titleOpt, &p, this);
  }
};

int main(int argc, char **argv) {
  QApplication app(argc, argv);

  qDebug() << QStyleFactory::keys();

  MyWidget window;

  window.show();

  return app.exec();
}
```

GitMensch:
[...]

PR #445, opened 2022-12-03
initial gitpod configuration

GitMensch:
* setup and building, including dependency KDDockWidgets
* x11 config allowing to preview the GUI in the browser

This enables easy direct work on Hotspot if you only have the browser around and removes the necessary setup for that.
Just start (wait a bit until that all is built) and work.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -18,6 +18,7 @@
   - [Arch Linux](#arch-linux)
   - [OpenSUSE](#opensuse)
   - [Building Hotspot itself](#building-hotspot-itself)
+- [Online](#online)
 - [Debugging the AppImage](#debugging-the-appimage)
   - [AppImage-Debugging using manual symbol loading](#appimage-debugging-using-manual-symbol-loading)
   - [AppImage-Debugging using debuginfod](#appimage-debugging-using-debuginfod)
@@ -124,6 +125,19 @@ make
 ```
 
 If you need help building this project for your platform, [contact us for help](https://www.kdab.com/about/contact/).
+
+## Online
+
+Instead of installing all dependencies and building Hotspot locally you may do so online with the following link:
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io#https://github.com/GitMensch/hotspot)
+This environment uses a clean, pre-configured Ubuntu 22.04 LTS environment. Just open, possibly
+have a look at the multiple building tasks, wait a bit until building is finished which is done
+as soon as a hotspot window opens in the preview and you can start coding / debugging hotspot.
+
+Notes:
+
+- This Gitpod setup is maintained by @GitMensch and not used by @KDAB directly.
+- Building may take a while as we don't use pre-builts or a special docker container in this environment yet.
 
 ## Debugging the AppImage
 
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ rustc_demangle.build.tar.bz2
 d-demangle.tar.gz
 *.actual
 *.orig
+*.cache
 scripts/output
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,124 @@
+# X11 default option to be able to test the result in the browser
+image: gitpod/workspace-full-vnc
+ports:
+  - port: 5900
+    onOpen: ignore
+  - port: 6080
+    onOpen: open-preview
+
+# build tasks, run in order but waiting on sync-points
+# last one is seen first
+tasks:
+  - name: setup coding environment on Ubuntu 22.04
+    before: |
+      sudo apt-get update
+      sudo apt-get -y upgrade
+      sudo apt-get install -y qtbase5-dev qtbase5-private-dev libqt5svg5-dev libqt5x11extras5-dev \
+          cmake libdwarf-dev mesa-common-dev libqcustomplot-dev \
+          libboost-iostreams-dev libboost-program-options-dev libboost-system-dev libboost-filesystem-dev \
+          build-essential curl git wget autotools-dev autoconf libtool liblzma-dev libz-dev gettext cmake \
+          libzstd-dev ninja-build libdw-dev libelf-dev extra-cmake-modules \
+          libkf5archive-dev libkf5threadweaver-dev libkf5i18n-dev libkf5configwidgets-dev \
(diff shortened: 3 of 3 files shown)


GitMensch:
This actually worked out quite good. Apart from the initial config I've added a second commit with a README adjustment, allowing you to "click" on it and have a look what it does. This second commit can be amended or dropped completely.

@milianw what do you think?

milianw:
can you explain when one would want to use that? can one open local files in the hotspot that is opened there? or is this meant to be used for development by people who don't want to have a local dev setup?

generally, I don't know anything about gitpod and have never used nor heard about it yet. please give a better motivation in your commit message.

also, we do have a pretty full featured docker image already, would it help to re-use that for gitpod? the problem would be where to host the image, as the images on ghcr.io cannot be accessed without a github token.

GitMensch:
Sure.

The idea is mostly to have an "instant" running dev environment without any local setup, allowing to directly start on PRs (or even just for tinkering).
Having it online at hand is also very useful when you only have a browser, no problem to do that from a tablet.

As you can ssh/scp to that machine you could also open recorded perfdata files (note: the README says something about the exported format to be unstable, not sure if that still is the case), that's a nice idea I did not have before.

The existing docker container could be used, too - this would make it "more instant" (and if it would be usable without the token could be easily users by others in Github Actions in general, the fork I've done does run and fail everything because of the missing token). But that is a general separate issue.

GitMensch:
Does the link in the updated README work for you? Maybe you can tinker around a bit (keeping in mind that the initial setup takes a bit because of updating the system to QT 5.15 and building KDDockWidgets up front.... (I'm creating the Docker issue right now...)

milianw:
I don't get it, how would I compile the code in the browser then?

Generally, as this is not something I'm endorsing myself, please state that clearly in the README blurb. Also better document how one would use that to work on the project, or link to the necessary documentation. As-is it's not clear to me at all how I would be using this.

GitMensch:
> I don't get it, how would I compile the code in the browser then?

maybe just have a look? What is happening when you click on

[Open in Gitpod](https://gitpod.io#https://github.com/GitMensch/hotspot)

In short:
* a docker container is loaded for you which has a web-x11 configured, that will start hotspot once it is build and has an automated checkout of the repo from the URL 
* this container also features a web-coding via "vscode in the browser" where you actually interact with that container and allows you to use any command via its GUI or via the terminal window
* it then builds the software (see the recipe in the yml file, should be similar enough scripts+definitions found in this repo)
* starts hotspot (you see that in the web preview, but can also move that out to a separate browser window)

milianw:
I did try that out, but the last two points didn't happen for me. I saw a VS code editor and an XFCE desktop. I tried to build but the "build and run" said that the cmake tools are not installed. I just tried it again and could reproduce that:

Screenshot_20221205_111833

GitMensch:
Thanks for trying it out - Interesting.

Note: _very likely_ the complete installation was not finished (this should get much faster with #446)

Please have a look at the 4 tasks (those in the lower right corner) that start when the workspace starts they wait for each other and I think the last one was not finished, as soon a it is (and it waits for all other tasks as necessary) hotspot should open (and then you may open the xfce in a separate window).

milianw:
Where do you see the 4 tasks, can you please highlight the area in the screenshot? I cannot see anything indicating an ongoing task

GitMensch:
Sure. To ensure I have a "clean start" for that I've went to https://gitpod.io/workspaces and clicked on "delete" of the existing workspace, then opened it new (from URL or from workspace overview, doesn't matter).
As I've configured the gitpod branch in the repo as the default one, it is started, and as you've recognized, builds something and starts the X11 session.

The tasks (active and finished) ar in the lower right corner
grafik
waiting a bit (you can click on the tasks to see what they perform) leads then to hotspot being build and as last task that waits for the others, starts it - this was the exact moment where this happened:

grafik

As you may see now, the three above are a bit "greyed out" - because they ar finished, what you also see is that hotspot was started.

... at least I hope you can reproduce it.

And hey, that is X11 again so the issue #339 can be seen there, too.

GitMensch:
Rebased and rechecked, open for comments.

The main "open" issue was: can we include the online/gitpod note and "link" in HACKING or should the last commit be adjusted/removed?

HACKING.md
```diff
@@ -24,6 +25,14 @@
 
 <!-- tocstop -->
 
+## Online
```
milianw, line 28: I'm not OK with putting this _before_ the "building hotspot" section. I'm also lacking some kind of motivation - when would one use this setup? To actually work on hotspot and write patches without a local dev environment? Or just to run hotspot "in the browser"? Because building isn't needed thanks to appimages either?

GitMensch: The answer is "to actually work on hotspot in the browser, writing patches in a clean environment, not needing a complete local dev environment", I've adjusted that and moved the Online part one below the building.

GitMensch:
Rebased to current master, updated all dependencies and updated to Ubuntu Jammy.

GitMensch:
@milianw I've moved the doc note. If you're fine with the current state I'd adjust it to this repo, so you can pull; if not please drop a note which changes you would like to see.

HACKING.md
```diff
@@ -125,6 +126,15 @@ make
 
 If you need help building this project for your platform, [contact us for help](https://www.kdab.com/about/contact/).
 
+## Online
+
+Instead of installing all dependencies and building Hotspot locally you may do so online with the following link:
```
milianw, line 131: please add a sentence like

Note: This Gitpod setup is maintained externally by @GitMensch   and not part of the official KDAB hotspot repository.

milianw (changes requested):
I think this can (should?) all be one commit as the individual pieces don't really add any value on their own?

that said - for me only the little remark in the documentation is needed before this can land. As I won't be using this myself, I want to make it clear that I won't maintain this myself. Can we maybe instead just link a GitPod.md file in your fork of hotspot on github, where you explain this setup and then people can report issues there?

GitMensch:
Squashing the commits would be possible, sure.

Linking to an external repo seems bad, as there is no control over that (and personally I try to clean up the repos under my username on GH, too).

Still adding a sentence like "Note: This Gitpod setup is maintained by @GitMensch and not used by @KDAB itself." to that paragraph would be possible.

What do you think?

milianw:
thanks

GitMensch:
Thanks for merging it in, this makes at least my PRs a lot easier :-)
I'll watch out for improvements, the biggest one may be integration of #446.

Issue #425, opened 2022-11-14
Disassembly window: add "find" to source code view


GitMensch:
**Is your feature request related to a problem? Please describe.**
I'd like to navigate fast in the source code, adding an outline view would be overkill, but ...

**Describe the solution you'd like**
... adding a find/search would be very useful.

**Describe alternatives you've considered**
opening the file in an editor as in #424, then looking for the right line, then manually scrolling to that line in the source code view.

Issue #418, opened 2022-11-14
FR: add option to set source paths


GitMensch:
Double clicking on a location in the callee/caller pane that has relative source file references I get the "failed to find file for location".

* recording: one locally before opening hotspot with `perf record --sample-cpu -z --aio --call-graph dwarf`
* original sources under: /home/me/project/lib
* original build under /home/me/project/build/lib - using relative files starting with ../../lib 
* use of installed binaries under /opt/install/project/lib
* everything works without any setup in GDB
* hotspot fails to resolve the source

I _think_ this likely can be worked around using the configuration, but as my tinkering with that did not work out asking for help.

milianw:
See the other report: please help us reproduce this issue by providing a project we can compile that triggers this behavior

GitMensch:
The other issue is about handling stored debug info correct.
This issue is about the option to configure paths. How would I configure Hotspot in the following scenario:

~~~
pushd tests/test-clients
gcc $(PWD)/cpp-stdin/main.cpp
popd
mv tests/test-clients/cpp-stdin ./cpp-stdin 
perf record --sample-cpu -z --aio --call-graph dwarf ./tests/test-clients/a.out
~~~

* source file not on the original place
* hotspot can't find that file
* user wants to configure lookup paths

milianw:
Ah I see, yes we should probably add some kind of path mapping or at least a list of source paths to query.

GitMensch:
Note: GDB comes with both `directories` and `substitute-path` settings; but I _guess_ the first one would work for most cases where perf is used.

GitMensch:
kind of related to #450 - which is about binary data as both are easily stored for inspecting perf data later / on a different machine

PR #448, opened 2022-12-05
make syntax highlighting customizeable

### LINKED ISSUE #434: FR Source Code window: configurable source highlighting
In some "rarer" cases the highlighter chooses the wrong syntax, it _seems_ that this is all filename (extension) based now:

https://github.com/KDAB/hotspot/blob/fd2e8fda9df8281a515ff2621ca455c8402f6e31/src/models/sourcecodemodel.cpp#L64-L69

An option to change the language the highlighter uses, ideally also away to see what highlighter is used.
I'd suggest to add a footer line to the source window which has a label "syntax" and a drop down to choose between the available syntax (also showing which syntax is used). For performance reasons the list should only be created when the Disassembly View is opened the first time (or was closed).

**Additional context**
This is most likely to happen with preparsed files (bison/flex/embedded sql preprocessors/...) - and also with "uncommon" file extensions.

lievenhey:
fixes #434

### DIFF
--- a/src/models/disassemblymodel.cpp
+++ b/src/models/disassemblymodel.cpp
@@ -16,13 +16,12 @@
 #include "highlighter.hpp"
 #include "sourcecodemodel.h"
 
-DisassemblyModel::DisassemblyModel(QObject* parent)
+DisassemblyModel::DisassemblyModel(KSyntaxHighlighting::Repository* repository, QObject* parent)
     : QAbstractTableModel(parent)
     , m_document(new QTextDocument(this))
-    , m_highlighter(new Highlighter(m_document, this))
+    , m_highlighter(new Highlighter(m_document, repository, this))
 {
     m_document->setUndoRedoEnabled(false);
-    m_highlighter->setDefinitionForName(QStringLiteral("GNU Assembler"));
 }
 
 DisassemblyModel::~DisassemblyModel() = default;
--- a/src/models/disassemblymodel.h
+++ b/src/models/disassemblymodel.h
@@ -18,11 +18,16 @@
 class QTextDocument;
 class Highlighter;
 
+namespace KSyntaxHighlighting {
+class Definition;
+class Repository;
+}
+
 class DisassemblyModel : public QAbstractTableModel
 {
     Q_OBJECT
 public:
-    explicit DisassemblyModel(QObject* parent = nullptr);
+    explicit DisassemblyModel(KSyntaxHighlighting::Repository* repository, QObject* parent = nullptr);
     ~DisassemblyModel();
 
     void setDisassembly(const DisassemblyOutput& disassemblyOutput, const Data::CallerCalleeResults& results);
@@ -38,6 +43,11 @@ public:
 
     Data::FileLine fileLineForIndex(const QModelIndex& index) const;
     QModelIndex indexForFileLine(const Data::FileLine& line) const;
+
+    Highlighter* highlighter() const
+    {
+        return m_highlighter;
+    }
 
     enum Columns
     {
--- a/src/models/highlighter.cpp
+++ b/src/models/highlighter.cpp
@@ -19,13 +19,16 @@
 #include <KSyntaxHighlighting/Theme>
 #endif
 
-Highlighter::Highlighter(QTextDocument* document, QObject* parent)
+Highlighter::Highlighter(QTextDocument* document, KSyntaxHighlighting::Repository* repository, QObject* parent)
     : QObject(parent)
 #if KF5SyntaxHighlighting_FOUND
-    , m_repository(std::make_unique<KSyntaxHighlighting::Repository>())
     , m_highlighter(new KSyntaxHighlighting::SyntaxHighlighter(document))
+    , m_repository(repository)
 #endif
 {
+#if !KF5SyntaxHighlighting_FOUND
+    Q_UNUSED(repository);
+#endif
     document->setDefaultFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
 
     // if qApp is used, UBSAN complains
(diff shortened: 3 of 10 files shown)


src/models/disassemblymodel.h
```diff
@@ -18,11 +18,16 @@
...
+class Definition;
+class Repository;
+}
+
 class DisassemblyModel : public QAbstractTableModel
 {
     Q_OBJECT
 public:
-    explicit DisassemblyModel(QObject* parent = nullptr);
+    explicit DisassemblyModel(KSyntaxHighlighting::Repository* repository = nullptr, QObject* parent = nullptr);
```
milianw, line 30: the repo shouldn't be optional, should it?

src/models/disassemblymodel.cpp
```diff
@@ -16,13 +16,14 @@
...
-DisassemblyModel::DisassemblyModel(QObject* parent)
+DisassemblyModel::DisassemblyModel(KSyntaxHighlighting::Repository* repository, QObject* parent)
     : QAbstractTableModel(parent)
     , m_document(new QTextDocument(this))
-    , m_highlighter(new Highlighter(m_document, this))
+    , m_highlighter(new Highlighter(m_document, repository, this))
 {
     m_document->setUndoRedoEnabled(false);
-    m_highlighter->setDefinitionForName(QStringLiteral("GNU Assembler"));
+    // TODO
```
milianw, line 25: yes todo :P

src/resultsdisassemblypage.cpp
```diff
@@ -123,6 +136,39 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(QWidget* parent)
...
+        auto m = new QStringListModel(this);
+        m->setStringList(schemes);
+
+        auto completer = new QCompleter(this);
+        completer->setModel(m);
+        completer->setCaseSensitivity(Qt::CaseInsensitive);
+        completer->setCompletionMode(QCompleter::PopupCompletion);
+        box->setCompleter(completer);
+
+        connect(completer, qOverload<const QModelIndex&>(&QCompleter::activated), this,
```
milianw, line 157: why is this using a completer? what happens when you select something with the mouse graphically?

why is the combo box even editable, i.e. why the completer? what happens if one enters garbage?

src/resultsdisassemblypage.cpp
```diff
@@ -186,6 +232,16 @@ void ResultsDisassemblyPage::showDisassembly(const DisassemblyOutput& disassembl
     m_disassemblyModel->clear();
     m_sourceCodeModel->clear();
 
+#if KF5SyntaxHighlighting_FOUND
+    auto sourceCodeDefinition = m_repository->definitionForFileName(disassemblyOutput.mainSourceFileName);
+
+    ui->assemblyComboBox->setCurrentText(QStringLiteral("GNU Assembler"));
```
milianw, line 238: this is the wrong place to do that

src/resultsdisassemblypage.cpp
lievenhey: > why is this using a completer?

To suggest highlighting definitions. Since there are multiple ways for some languages (C++, CPP, CXX, cpp, ...) I use a complete to suggest an installed definition.

> what happens when you select something with the mouse graphically?

it works (`QComboBox::activated` triggers)

> why is the combo box even editable, i.e. why the completer?

Its the easiest way to add a search function. I didn't want a list, since there 350 different definitions

> what happens if one enters garbage?

Syntax Highlighting gets disabled

src/resultsdisassemblypage.cpp
```diff
@@ -123,6 +136,39 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(QWidget* parent)
...
+        completer->setCaseSensitivity(Qt::CaseInsensitive);
+        completer->setCompletionMode(QCompleter::PopupCompletion);
+        box->setCompleter(completer);
+
+        connect(completer, qOverload<const QModelIndex&>(&QCompleter::activated), this,
+                [this, model](const QModelIndex& entry) {
+                    model->setSyntaxHighlightDefinition(m_repository->definitionForName(entry.data().toString()));
+                });
+        connect(box, qOverload<int>(&QComboBox::activated), this, [this, model, box]() {
+            model->setSyntaxHighlightDefinition(m_repository->definitionForName(box->currentText()));
```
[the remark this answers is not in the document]
lievenhey: when a file is opened, it will be set to file extension + ASM
the code above only triggers if the user sets the syntax highlighting to something custom

src/resultsdisassemblypage.cpp
lievenhey: why? this needs to be set if a file gets opened, in case the user changed something

milianw: the problem is that you have to set it twice, manually. that's wrong. I think here it should only be needed to set the definition once on the respective model which should then emit a signal which lets the ui adapt to select the corresponding item

src/resultsdisassemblypage.cpp
milianw: put differently, do we need both? isn't `QComboBox::activated` getting emitted when we select an item from the completer? do we really need both? if so, share the common code in a lambda and call it from both slots to make it clear that these do the same thing

we also need the back-sync code to select the current syntax highlighting mode from the model in the ui

src/models/highlighter.cpp
```diff
@@ -19,13 +19,16 @@
...
     : QObject(parent)
 #if KF5SyntaxHighlighting_FOUND
-    , m_repository(std::make_unique<KSyntaxHighlighting::Repository>())
     , m_highlighter(new KSyntaxHighlighting::SyntaxHighlighter(document))
+    , m_repository(repository)
 #endif
 {
+#if KF5SyntaxHighlighting_FOUND == 0
+    Q_UNUSED(repository);
+#endif
```
milianw, line 31:
```
#else
    Q_ASSERT(repository);
```

milianw: otherwise the code below would crash where `m_repository` is accessed without checks

I think this might actually lead to crashes in the unit tests when `updateColorTheme` is called below?

milianw (changes requested):

src/resultsdisassemblypage.cpp
lievenhey: `QCompleter::activated` is not needed (qt docs weren't clear)

GitMensch:
Thanks!

src/models/highlighter.hpp
```diff
@@ -17,18 +17,21 @@ class QTextDocument;
...
 {
     Q_OBJECT
 public:
-    Highlighter(QTextDocument* document, QObject* parent = nullptr);
+    Highlighter(QTextDocument* document, KSyntaxHighlighting::Repository* repository, QObject* parent = nullptr);
     ~Highlighter();
 
-    void setDefinitionForFilename(const QString& filename);
-    void setDefinitionForName(const QString& name);
+    void setDefinition(const KSyntaxHighlighting::Definition& definition);
```
milianw, line 31: add getter for the current definition too

src/models/sourcecodemodel.h
```diff
@@ -59,6 +69,7 @@ class SourceCodeModel : public QAbstractTableModel
 public slots:
     void updateHighlighting(int line);
     void setSysroot(const QString& sysroot);
+    void setSyntaxHighlightDefinition(const KSyntaxHighlighting::Definition& definition);
```
milianw, line 72: dito, remove

milianw (changes requested):

milianw (approved):

GitMensch:
Thank you for all those useful changes to the new view! The only feature-wise thing open is #425 (I may be able to try adding that using `QTextDocument.find` [the hardest part for me would be the popup? to add, allowing the user to entering his search and the `FindFlags`]) and the likely more complicated #418 (something I don't feel comfortable tinkering with).

PR #457, opened 2023-01-05
improvements for the disassembler

### LINKED ISSUE #450: Disassembly: objdump call does not honor debugPaths  / extraLibPaths
**Describe the bug**
The disassembly feature seems to call objdump always with the recorded path, especially when you take an exported .perfparser file and the matching shared objects to a different machine it can be cumbersome to recreate the original structure - and won't be useful if the shared object is a different version than the one recorded earlier

**To Reproduce**

1. record perf data which include libsome.so
2. convert perf data to perfparser data
3. move libsome.so around
4. start hotspot with perfparser file - everything is still fine
5. use disassembly option, see an error message of objdump not finding the file

Note: the "move" of debugging data is possibly less relevant than "recorded, converted, source and binaries saved" and then opening the perfdata later finding the debugging data changed in the meanwhile. Using debugPaths  / extraLibPaths would fix both.

**Expected behavior**
* searching the file first, then passing its path to objdump
* search with the given so-name in debugPaths first, then in extraLibPaths 
* only if it isn't found in those check the original one
* if the file cannot be resolved: give a clear message of that missing file (instead of "objdump error")

lievenhey:
Now the disassembler looks at symbol.actualPath and checks if that file exists (usualy located in .debug)
If the file does not exists it searches in debugPaths and then in extraLibPaths
If none is found it tries symbol.path (original lib file which may not include debug informations)

fixes #450


src/resultsdisassemblypage.ui
```diff
@@ -34,13 +34,76 @@
...
    </item>
    <item>
-    <widget class="KSqueezedTextLabel" name="filenameLabel">
-     <property name="text">
-      <string>KSqueezedTextLabel</string>
-     </property>
-     <property name="textElideMode">
-      <enum>Qt::ElideRight</enum>
-     </property>
+    <widget class="QWidget" name="widget_4" native="true">
```
GitMensch, line 37: can you please adjust the search bar to share its horizontal layout with the customSourceCodeHighlighting widget? This would "save" a UI line and would then also be reasonable to be always visible.

GitMensch: I agree that this isn't better; but isn't there a horizontal "container" layout that can take two widgets side by side?

lievenhey: There is QHorizontalLayout but it still looks weird

grafik

lievenhey: by using three widgets it does no longer align with the two code views

src/models/disassemblyoutput.cpp
```diff
@@ -67,6 +70,61 @@ DisassemblyOutput::LinkedFunction extractLinkedFunction(const QString& disassemb
     }
     return function;
 }
+
+QString findInSubdirRecursive(const QString& path, const QString& filename)
+{
+    // find filename in path
+    // some distros (Ubuntu) use subdirs to store their debug files
+    QString filepath = path + QDir::separator() + filename;
```
milianw, line 78: this method can be substantially simplified by leveraging QDirIterator

src/models/disassemblyoutput.cpp
```diff
@@ -67,6 +70,61 @@ DisassemblyOutput::LinkedFunction extractLinkedFunction(const QString& disassemb
...
+
+    auto result = findBinary(settings->debugPaths().split(QLatin1Char(':')), symbol.binary);
+    if (!result.isEmpty())
+        return result;
+
+    result = findBinary(settings->extraLibPaths().split(QLatin1Char(':')), symbol.binary);
+    if (!result.isEmpty())
+        return result;
+
+    // disassemble the binary if no debug fiel was found
```
milianw, line 121: typo: file

src/models/disassemblyoutput.cpp
```diff
@@ -188,7 +246,12 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
     else
         qCInfo(disassemblyoutput) << "objdump binary does not support `--visualize-jumps`:" << processPath;
 
-    arguments.append(symbol.actualPath);
+    auto binary = findBinaryForSymbol(symbol);
+    if (binary.isEmpty()) {
+        disassemblyOutput.errorMessage += QApplication::tr("Could not find %1").arg(symbol.binary);
```
milianw, line 251: use different QObject (something related to disassembly) as tr context

src/resultsdisassemblypage.cpp
```diff
@@ -135,12 +136,41 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
                              [functionName](const Data::Symbol& symbol) { return symbol.symbol == functionName; });
 
             if (symbol != m_callerCalleeResults.entries.keyEnd()) {
-                setSymbol(*symbol);
-                showDisassembly();
+                m_symbolStack.push_back(*symbol);
+                m_stackIndex++;
+                emit stackChanged();
+            } else {
+                ui->symbolNotFound->setText(tr("%1 is not known to hotspot").arg(functionName));
```
milianw, line 144: simplify the message, just "unknown symbol %1"

src/resultsdisassemblypage.cpp
```diff
@@ -135,12 +136,41 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+                ui->symbolNotFound->show();
             }
         }
     });
 
+    connect(ui->stackBackButton, &QPushButton::pressed, this, [this] {
+        m_stackIndex--;
+        if (m_stackIndex < 0)
+            m_stackIndex = 0;
```
milianw, line 153: shouldn't this wrap around to the last item then?

src/resultsdisassemblypage.cpp
```diff
@@ -135,12 +136,41 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+
+        emit stackChanged();
+    });
+
+    connect(ui->stackNextButton, &QPushButton::pressed, this, [this] {
+        m_stackIndex++;
+        if (m_stackIndex >= m_symbolStack.size())
+            m_stackIndex = m_symbolStack.size() - 1;
```
milianw, line 161: shouldn't this wrap around to 0 then?

src/resultsdisassemblypage.cpp
```diff
@@ -215,7 +245,7 @@ void ResultsDisassemblyPage::setupAsmViewModel()
 void ResultsDisassemblyPage::showDisassembly()
 {
     // Show empty tab when selected symbol is not valid
-    if (m_curSymbol.symbol.isEmpty()) {
+    if (m_symbolStack[m_stackIndex].symbol.isEmpty()) {
```
milianw, line 280: also check m_symbolStack.isEmpty() as this would otherwise crash

milianw: then introduce a local and reuse it below

```
const auto &curSymbol = m_symbolStack[m_stackIndex];
```

src/resultsdisassemblypage.cpp
```diff
@@ -231,7 +261,9 @@ void ResultsDisassemblyPage::showDisassembly()
         return isArm ? QStringLiteral("arm-linux-gnueabi-objdump") : QStringLiteral("objdump");
     };
 
-    showDisassembly(DisassemblyOutput::disassemble(objdump(), m_arch, m_curSymbol));
+    ui->symbolNotFound->hide();
+
+    showDisassembly(DisassemblyOutput::disassemble(objdump(), m_arch, m_symbolStack[m_stackIndex]));
```
milianw, line 303: see above

src/resultsdisassemblypage.cpp
```diff
@@ -245,11 +277,11 @@ void ResultsDisassemblyPage::showDisassembly(const DisassemblyOutput& disassembl
     m_disassemblyModel->highlighter()->setDefinition(m_repository->definitionForName(QStringLiteral("GNU Assembler")));
 #endif
 
-    const auto& entry = m_callerCalleeResults.entry(m_curSymbol);
+    const auto& entry = m_callerCalleeResults.entry(m_symbolStack[m_stackIndex]);
```
milianw, line 317: see above

src/resultsdisassemblypage.cpp
```diff
@@ -245,11 +277,11 @@ void ResultsDisassemblyPage::showDisassembly(const DisassemblyOutput& disassembl
...
 
     ui->filenameLabel->setText(disassemblyOutput.mainSourceFileName);
     // don't set tooltip on symbolLabel, as that will be called internally and then get overwritten
-    setToolTip(Util::formatTooltip(entry.id, m_curSymbol, m_callerCalleeResults.selfCosts,
+    setToolTip(Util::formatTooltip(entry.id, m_symbolStack[m_stackIndex], m_callerCalleeResults.selfCosts,
```
milianw, line 321: see above

src/resultsdisassemblypage.cpp
```diff
@@ -171,6 +172,37 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
 
+    auto searchShortcut = new QShortcut(QKeySequence::Find, this);
+    connect(searchShortcut, &QShortcut::activated, this, [this] {
+        ui->searchWidget->setVisible(true);
+        ui->searchEdit->setFocus();
+    });
+
+    connect(ui->closeSearch, &QPushButton::pressed, this, [this] { ui->searchWidget->setVisible(false); });
+
+    connect(ui->nextSearchResult, &QPushButton::pressed, this,
```
milianw, line 183: can this be used purely from the keyboard? F3 / Shift F3?

src/resultsdisassemblypage.cpp
```diff
@@ -171,6 +172,37 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
         showDisassembly();
     });
 
+    auto searchShortcut = new QShortcut(QKeySequence::Find, this);
+    connect(searchShortcut, &QShortcut::activated, this, [this] {
+        ui->searchWidget->setVisible(true);
+        ui->searchEdit->setFocus();
+    });
+
+    connect(ui->closeSearch, &QPushButton::pressed, this, [this] { ui->searchWidget->setVisible(false); });
```
milianw, line 181: can this be closed from the keyboard with Escape?

src/resultsdisassemblypage.cpp
```diff
@@ -171,6 +172,37 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+        ui->sourceCodeView->scrollTo(m_searchResultIndex, QAbstractItemView::ScrollHint::PositionAtCenter);
+        ui->searchResultWarning->hide();
+    });
+
+    // search if enter is pressed
+    connect(ui->searchEdit, &QLineEdit::returnPressed, ui->nextSearchResult, &QPushButton::pressed);
+
+    // I don't know how to show an error so I use a warning widget
+    // kate, ... set the backgroundcolor to red but I don't know how they do this
+    // google say ->setStylesheet("background: red");
```
milianw, line 200: no, stylesheets are bad as our KDAB blog should have thought you! we can set a palette with a negative foreground text color, KColorScheme even has API for that afair

src/resultsdisassemblypage.h
```diff
@@ -83,6 +83,8 @@ class ResultsDisassemblyPage : public QWidget
     CodeDelegate* m_disassemblyDelegate;
     CodeDelegate* m_sourceCodeDelegate;
 
+    QModelIndex m_searchResultIndex = {};
```
milianw, line 86: remove `= {};` - unneeded

milianw (changes requested):
great stuff, but overall some changes are required

GitMensch:
Note: If possible it would be nice to have the Dissassembly Window be split, so the sizes of "source" vs. "assembly" can be adjusted. The later often has < 60 characters (at least with C, is that different with C++?), while the code is much wider. As it is now there is often a lot of empty space on the right side while the left has the code elided  - and the user seems to not able to adjust these sizes.

GitMensch:
Thanks for 2ac1771349e90af5f1b6c53fc824dc1cd0fd883d, sadly I can't get to to work with the resulting appimage (but that also has no search, so I guess it has build "something", not this PR...

GitMensch:
Nevermind, I've forgot the `chmod +x` and used an old version. The sliding with the new version works perfectly!
For the search there seems to be something missing in the UI (at least the menu, possibly?) for the search; I _think_ a magnifying glass in the bottom bar which contains the syntax highlighting selection would be ideal.
Using [CTRL]+[F], then up/down buttons work fine (a "case-sensitive" check mark would be nice, though).

GitMensch:
@milianw: Any chance to re-review?

lievenhey:
there is now a magnifying glass and a checkbox for a case sensitive search

GitMensch:
The adjusted search works good. The only missing thing seems to be some visual feedback, maybe just select the line with the result?

README.md
```diff
@@ -314,6 +315,17 @@ hotspot currently only shows the name of the tracepoints in the timeline.
...
 ![hotspot tracepoints](screenshots/tracepoints.png?raw=true "hotspot tracepoints")
 
+### Disassembler
+
+![hotspot disassembler page](screenshots/disassembler.png?raw=true "hotspot disassembler page")
+
+Hotspot includes an disassembler, which can show you the cost per instruction.
+The disassembler uses colors to indicate which assembly lines correspond to
+which source code line.
+For easier navigation, you can simply clock on a line and the other view will jump to it.
```
GitMensch, line 325: typo: click, additional

README.md
```diff
@@ -314,6 +315,17 @@ hotspot currently only shows the name of the tracepoints in the timeline.
...
+You can follow function calls with a double click.
+In the sourcecode view you can press ctrl+f to open a search window.
```
GitMensch, line 327: or use the magnifier icon  (possibly update the png fie)

GitMensch (commented):
Other than the "visual feedback" on the search and the minor README notes, I see no open points from a user-perspective.

src/resultsdisassemblypage.cpp
```diff
@@ -214,8 +244,13 @@ void ResultsDisassemblyPage::setupAsmViewModel()
 
 void ResultsDisassemblyPage::showDisassembly()
 {
+    if (m_symbolStack.isEmpty())
+        return;
+
+    auto& curSymbol = m_symbolStack[m_stackIndex];
```
milianw, line 285: `const auto &`, the symbol shouldn't get modified, or?

src/resultsdisassemblypage.cpp
```diff
@@ -231,25 +266,31 @@ void ResultsDisassemblyPage::showDisassembly()
...
 }
 
 void ResultsDisassemblyPage::showDisassembly(const DisassemblyOutput& disassemblyOutput)
 {
     m_disassemblyModel->clear();
     m_sourceCodeModel->clear();
 
+    // this function is only called if m_symbolStack is non empty (see above)
+    Q_ASSERT(!m_symbolStack.isEmpty());
+    auto& curSymbol = m_symbolStack[m_stackIndex];
```
milianw, line 320: dito

src/models/sourcecodemodel.cpp
```diff
@@ -230,3 +230,46 @@ void SourceCodeModel::setSysroot(const QString& sysroot)
...
+
+        if (searchLine(lineIndex, searchString, caseSensitive)) {
+            emit searchResultFound(index(lineIndex, SourceCodeColumn));
+            return;
+        }
+    }
+    emit noSearchResult();
+}
+
+void SourceCodeModel::findBackwards(const QString& search, const QModelIndex& start, bool caseSensitive)
```
milianw, line 250: this is basically the same code as above, the only difference is the direction to go from the current line. please extract common code out and share code

src/models/sourcecodemodel.cpp
```diff
@@ -230,3 +230,46 @@ void SourceCodeModel::setSysroot(const QString& sysroot)
...
+    emit noSearchResult();
+}
+
+bool SourceCodeModel::searchLine(int index, const QString& search, bool caseSensitive)
+{
+    auto block = m_document->findBlockByLineNumber(index + m_startLine);
+    if (!block.isValid())
+        return false;
+
+    auto stringToSearch = caseSensitive ? block.text() : block.text().toLower();
```
milianw, line 272: instead of toLower here, pass Qt::CaseSensitivity to QString::contains

src/resultsdisassemblypage.cpp
```diff
@@ -13,6 +13,7 @@
 #include <QDirIterator>
 #include <QFile>
 #include <QFileDialog>
+#include <QKeyEvent>
```
milianw, line 16: because I cannot comment on the commit message:

please get into the habit of writing better commit messages! a one liner is only acceptable for really trivial typo fixes or similar. a non-trivial patch like this here (and the others) definitely deserves a real commit messages. give a summary of what the patch does (e.g. what kind of keyboard navigation? what keys? for what widget?) also give an explanation for some design choices, i.e. why do you use an even filter instead of `QShortcut` or similar?

better yet, use `KStandardAction`: https://api.kde.org/legacy/pykde-4.7-api/kdeui/KStandardAction.html#findNext https://api.kde.org/legacy/pykde-4.7-api/kdeui/KStandardAction.html#findPrev

src/resultsdisassemblypage.ui
```diff
@@ -113,34 +113,31 @@
...
-        <layout class="QVBoxLayout" name="verticalLayout_4">
-         <property name="spacing">
-          <number>0</number>
-         </property>
-         <property name="leftMargin">
-          <number>0</number>
-         </property>
-         <property name="topMargin">
-          <number>0</number>
+    <widget class="QSplitter" name="widget">
```
milianw, line 116: again, better commit messages! by reading the diff, it seems you achieve the result by adding a splitter, right? say so in the commit message!

milianw (changes requested):
please write better commit messages

GitMensch:
Just wondering: would it be more useful to split this PR into multiple parts? It solves multiple issues an it _may_ be faster (or only more work, not sure) to get  _some_ implemented/fixed if the number of changes that need to be reviewed are smaller?

lievenhey:
I removed the search stuff. This should now only contain reviewed code.
I will add the search again in a new PR.

GitMensch:
Thank you very much!

GitMensch:
If it isn't possible to re-review the reviewed code / merge that into master - could you please rebase the branch so the resulting appimage has both?

lievenhey:
done

GitMensch:
Thank you, the resulting AppImage works fine.

Would it be reasonable to tackle #418 in this PR, too? That issue is about adding settings and use them in the disassembler view.

GitMensch:
Ah, so there's only a missing GUI setting for `--extraLibPaths` in 6684819a (we do have some for `--debugPaths` already)?

milianw:
Thank you for your work @lievenhey as always, and sorry for taking so long with this one!

GitMensch:
> `6684819a` should already fix that. That commit allows the disassembler to search `--debugPaths` and `--extraLibPaths`.

That seems to handle only binaries, from `-help`:

>   --extraLibPaths <paths>  Colon separated list of extra paths to find libraries.

PR #458, opened 2023-01-10
use non-deprecated function KX11Extras::activeWindow

lievenhey:
since kf5 version 101 KWindowSystem::activeWindow is deprecated and
KX11Extras::activeWindow should be used

### DIFF
--- a/src/perfrecord.cpp
+++ b/src/perfrecord.cpp
@@ -20,7 +20,13 @@
 #include <unistd.h>
 
 #include <KUser>
+
+#include <kwindowsystem_version.h>
+#if KWINDOWSYSTEM_VERSION >= QT_VERSION_CHECK(5, 101, 0)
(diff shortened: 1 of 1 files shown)


src/perfrecord.cpp
```diff
@@ -63,9 +69,14 @@ static QStringList sudoOptions(const QString& sudoBinary)
 {
     QStringList options;
     if (sudoBinary.endsWith(QLatin1String("/kdesudo")) || sudoBinary.endsWith(QLatin1String("/kdesu"))) {
+#if KWINDOWSYSTEM_VERSION >= QT_VERSION_CHECK(5, 101, 0)
+        auto activeWindowId = QString::number(KX11Extras::activeWindow());
```
milianw, line 73: remove the QString::number from the two branches, only do that below in the `options.append()` line

PR #459, opened 2023-01-11
extend appimage debugging notes

GitMensch:


HACKING.md
```diff
@@ -125,18 +127,94 @@ make
...
+#15 0x000000000043a929 in main (argc=<optimized out>, argv=<optimized out>) at /__w/hotspot/hotspot/src/main.cpp:69
+(gdb) continue
+
+```
+
+or with an extracted image:
+
+```bash
+$ ./hotspot-v1.4.0-43-g02fd82c-x86_64.AppImage --appimage-extract
+$ PATH=./squashfs-root/usr/bin:$PATH LD_LIBRARY_PATH=./squashfs-root/usr/lib64:/usr/local/lib64:/usr/lib64:./squashfs-root/usr/lib:$LD_LIBRARY_PATH gdb -q --exec=hotspot --symbols hotspot-debuginfo-v1.4.0-43-g02fd82c/usr/bin/hotspot -ex run
```
milianw, line 203: maybe we should add some kind of `--appimage-debug gdb` or similar to our `AppRun` which would prepend the debugger before starting hotspot?

or maybe better: extract the code that sets up the env from the `AppRun`, such that we can do:

```
source appimage.env
gdb -q --exec=hotspot --symbols hotspot-debuginfo-v1.4.0-43-g02fd82c/usr/bin/hotspot -ex run
```

directly?

milianw (approved):
lgtm already, do you want to make it even better or should I merge as-is already?

GitMensch:
Thanks for the review - no further changes planned.

HACKING.md
GitMensch: appimage.env would be marvelous! Even better would be if appimage can be executed, too: `/path/to/appimage.env hotspot` (or, using gdb that way).

`--appimage-debug Name-Of-Debugger-Here [additional debugger arguments] [-- additional arguments passed to hotspot]` would be nice, too; allowing things like `/hotspot-v1.4.0-43-g02fd82c-x86_64.AppImage --appimage-debug /opt/path/to/newer/gdb/bin/gdb -x additional_printers.gdb --args -- some.perf.data`

Should we create one (or two) issue(s) for those good ideas?

GitMensch: Should we create one (or two) issue(s) for those good ideas of yours?

milianw: yes please do that

Issue #466, opened 2023-02-01
Caller / Callee view: Location is sorted  alphabetic


GitMensch:
**Describe the bug**
Sorting the Location (lower right part of the view) leads to lines 1,10,11,12,13,14,15,16,17,18,19,100,101,102,...

**To Reproduce**

1. Go to the view
2. order by Location
3. See error - if you have a source recorded with lines like that

**Expected behavior**
Order the source file part alphabetic, the line part numeric (1-9,10-99,...)

**Version Info (please complete the following information):**
- hotspot version: /hotspot-v1.4.1-44-g46244df-x86_64.AppImage

PR #468, opened 2023-02-28
fix sorting order of location in caller callee view

lievenhey:
sort by line not by alphabetic string
fixes: #466

### DIFF
--- a/src/models/callercalleeproxy.cpp
+++ b/src/models/callercalleeproxy.cpp
@@ -8,6 +8,7 @@
 
 #include "callercalleeproxy.h"
 
+#include "callercalleemodel.h"
 #include "data.h"
 
 namespace {
@@ -31,4 +32,21 @@ bool match(const QSortFilterProxyModel* proxy, const Data::FileLine& fileLine)
 
     return matchImpl(needle, fileLine.file);
 }
+}
+
+SourceMapProxy::SourceMapProxy(QObject* parent)
+    : CallerCalleeProxy<SourceMapModel>(parent)
+{
+}
+
+SourceMapProxy::~SourceMapProxy() = default;
+
+bool SourceMapProxy::lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const
+{
+    if (source_left.column() == source_right.column() && source_left.column() == SourceMapModel::Location) {
+        const auto left = source_left.data(SourceMapModel::SortRole).value<Data::FileLine>();
+        const auto right = source_right.data(SourceMapModel::SortRole).value<Data::FileLine>();
+        return left < right;
+    }
+    return QSortFilterProxyModel::lessThan(source_left, source_right);
 }
--- a/src/models/callercalleeproxy.h
+++ b/src/models/callercalleeproxy.h
@@ -20,6 +20,8 @@ bool match(const QSortFilterProxyModel* proxy, const Data::Symbol& symbol);
 bool match(const QSortFilterProxyModel* proxy, const Data::FileLine& fileLine);
 }
 
+class SourceMapModel;
+
 template<typename Model>
 class CallerCalleeProxy : public QSortFilterProxyModel
 {
@@ -43,6 +45,15 @@ protected:
 
         return CallerCalleeProxyDetail::match(this, key);
     }
+};
 
-private:
+class SourceMapProxy : public CallerCalleeProxy<SourceMapModel>
+{
+    Q_OBJECT
+public:
+    SourceMapProxy(QObject* parent = nullptr);
+    ~SourceMapProxy();
+
+protected:
+    bool lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const override;
 };
--- a/src/resultscallercalleepage.cpp
+++ b/src/resultscallercalleepage.cpp
@@ -36,11 +36,22 @@
 #endif
 
 namespace {
+QSortFilterProxyModel* createProxy(SourceMapModel* model)
+{
+    return new SourceMapProxy(model);
+}
+
+template<typename Model>
+QSortFilterProxyModel* createProxy(Model* model)
+{
+    return new CallerCalleeProxy<Model>(model);
+}
+
 template<typename Model>
 Model* setupModelAndProxyForView(QTreeView* view, CostContextMenu* contextMenu)
 {
     auto model = new Model(view);
-    auto proxy = new CallerCalleeProxy<Model>(model);
+    auto proxy = createProxy(model);
     proxy->setSourceModel(model);
     proxy->setSortRole(Model::SortRole);
     view->setModel(proxy);
--- a/src/models/callercalleemodel.h
+++ b/src/models/callercalleemodel.h
@@ -265,7 +265,7 @@ public:
     {
         if (role == SortRole) {
             if (column == Location) {
-                return fileLine.toString();
+                return QVariant::fromValue(fileLine);
             }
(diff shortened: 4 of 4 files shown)


milianw:
huh? how does this fix it? you now only sort by the line which is totally bogus? 

instead, we should use a QCollator in the proxy model for sorting and enable numeric mode on that

lievenhey:
I didn't know that sorting by filename is required. In that case we can leverage `FileLine::operator<`

src/models/callercalleeproxy.h
```diff
@@ -8,13 +8,9 @@
 
 #pragma once
 
+#include "data.h"
```
milianw, line 11: move to .cpp

src/models/callercalleeproxy.h
```diff
@@ -8,13 +8,9 @@
...
 #include <QSortFilterProxyModel>
 
-namespace Data {
```
milianw, line 13: keep this as-is

src/models/callercalleeproxy.h
```diff
@@ -44,5 +40,16 @@ class CallerCalleeProxy : public QSortFilterProxyModel
         return CallerCalleeProxyDetail::match(this, key);
     }
 
+    bool lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const override
```
milianw, line 43: de-inline

milianw (changes requested):
please don't inline the code, as it leads to code bloat

src/models/callercalleeproxy.h
```diff
@@ -44,5 +40,16 @@ class CallerCalleeProxy : public QSortFilterProxyModel
...
+    {
+        QVariant left = source_left.data(sortRole());
```
milianw, line 45: I would change the logic: check the column first like is done in the model itself and then only if that's matching request the variants and assert that `canConvert` is true

src/models/callercalleeproxy.h
lievenhey: sadly this is not possible, since Qt does not offer a compare function for QVariant (they added one in Qt 6.2)

milianw: I'm sorry, but you are doing it wrong™ 

the source model does not matter here, the only thing that matters is the role which should be Model::SortRole and could be hard coded in the ctor here already.

as such, we can expect (and assert!) that the values we get here are actually locations for the location column! Meaning, rewrite this as:

```
Q_ASSERT(sortRole() == Model::SortRole);
if (source_left.column() == source_right.column() && source_left.column() == Model::Location)
{
    return ...Detail::lessThanLocation(source_left, source_right, Model::SortRole);
}
```

milianw (changes requested):

src/models/callercalleeproxy.h
lievenhey: The source model matters, `ResultsCallerCalleePage` create a `CallerCalleeProxy` with a `CallerModel`, a `CalleeModel` and a `SourceMapModel`. Only the `SourceMapModel` has the `Location` column.

I should create a new proxy only for the `SourceMapModel` to handle this.

milianw: or use a `std::optional` or similar to pass the location column, and only when that's given use this feature.

src/models/callercalleeproxy.h
```diff
@@ -46,3 +50,24 @@ class CallerCalleeProxy : public QSortFilterProxyModel
 
 private:
 };
+
+template<typename Model>
+class SourceMapProxy : public CallerCalleeProxy<Model>
```
milianw, line 55: this shouldn't be a template anymore, but instead hardcode Model == SourceMapModel

src/models/callercalleeproxy.h
```diff
@@ -46,3 +50,24 @@ class CallerCalleeProxy : public QSortFilterProxyModel
...
+{
+public:
+    SourceMapProxy(QObject* parent = nullptr)
```
milianw, line 58: then this can be de-inlined, also add a non-inline dtor please

src/models/callercalleeproxy.h
```diff
@@ -46,3 +50,24 @@ class CallerCalleeProxy : public QSortFilterProxyModel
...
+        : CallerCalleeProxy<Model>(parent)
+    {
+    }
+
+protected:
+    bool lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const override
```
milianw, line 64: dito, this can then be de-inlined as it won't be a template anymore.

src/models/callercalleeproxy.h
```diff
@@ -20,6 +20,10 @@ bool match(const QSortFilterProxyModel* proxy, const Data::Symbol& symbol);
 bool match(const QSortFilterProxyModel* proxy, const Data::FileLine& fileLine);
 }
 
+namespace SourceMapProxyDetail {
```
milianw, line 23: see below: no need for this anymore, just de-inline the method directly

milianw (changes requested):

GitMensch:
Rechecked: I can confirm that the resulting appimage fixe the original issue just fine, sorting first the filename (alphanumeric) then the line numbers (numeric).

Thank you!

src/models/callercalleeproxy.h
```diff
@@ -46,3 +48,16 @@ class CallerCalleeProxy : public QSortFilterProxyModel
...
+{
+    Q_OBJECT
+public:
+    SourceMapProxy(QObject* parent = nullptr);
+    ~SourceMapProxy();
+
+protected:
+    bool lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const override;
+
+private:
```
milianw, line 62: remove this and the line above

src/models/callercalleeproxy.cpp
```diff
@@ -32,3 +33,21 @@ bool match(const QSortFilterProxyModel* proxy, const Data::FileLine& fileLine)
...
+SourceMapProxy::SourceMapProxy(QObject* parent)
+    : CallerCalleeProxy<SourceMapModel>(parent)
+{
+}
+
+SourceMapProxy::~SourceMapProxy() { }
+
+bool SourceMapProxy::lessThan(const QModelIndex& source_left, const QModelIndex& source_right) const
+{
+
```
milianw, line 46: remove this empty line

src/models/callercalleeproxy.cpp
```diff
@@ -32,3 +33,21 @@ bool match(const QSortFilterProxyModel* proxy, const Data::FileLine& fileLine)
     return matchImpl(needle, fileLine.file);
 }
 }
+
+SourceMapProxy::SourceMapProxy(QObject* parent)
+    : CallerCalleeProxy<SourceMapModel>(parent)
+{
+}
+
+SourceMapProxy::~SourceMapProxy() { }
```
milianw, line 42: = default;

src/resultscallercalleepage.cpp
```diff
@@ -36,11 +36,21 @@
 #endif
 
 namespace {
+QSortFilterProxyModel* createProxy(SourceMapModel* model)
+{
+    return new SourceMapProxy(model);
+}
+template<typename Model>
```
milianw, line 43: add empty line above

milianw (changes requested):
really minor nits, otherwise lgtm now

milianw (approved):
lgtm now, thanks

PR #470, opened 2023-03-10
add search function for source code was: cache source code in vector for easier access in search function

lievenhey:
I updated the SourceCodeModel to only include the function that is disassembled. Until now the whole file was included which wasted a lot of performance since KSyntaxHighlighter was run on the whole file.

### DIFF
--- a/src/models/sourcecodemodel.cpp
+++ b/src/models/sourcecodemodel.cpp
@@ -13,7 +13,10 @@
 #include <QTextBlock>
 #include <QTextDocument>
 
+#include <algorithm>
+
 #include "highlighter.hpp"
+#include "search.h"
 
 SourceCodeModel::SourceCodeModel(KSyntaxHighlighting::Repository* repository, QObject* parent)
     : QAbstractTableModel(parent)
@@ -243,4 +246,20 @@ QModelIndex SourceCodeModel::indexForFileLine(const Data::FileLine& fileLine) co
 void SourceCodeModel::setSysroot(const QString& sysroot)
 {
     m_sysroot = sysroot;
+}
+
+void SourceCodeModel::find(const QString& search, Direction direction, int offset)
+{
+    auto searchFunc = [&search](const SourceCodeLine& line) {
+        return line.text.indexOf(search, 0, Qt::CaseInsensitive) != -1;
+    };
+
+    int resultIndex = ::search(
+        m_sourceCodeLines, searchFunc, [this] { emit resultFound({}); }, direction, offset);
+
+    if (resultIndex >= 0) {
+        emit resultFound(createIndex(resultIndex + 1, SourceCodeColumn));
+    } else {
+        emit resultFound({});
+    }
 }
--- a/src/models/sourcecodemodel.cpp
+++ b/src/models/sourcecodemodel.cpp
@@ -51,6 +51,14 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
     if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
         return;
 
+    QString sourceCode = QString::fromUtf8(file.readAll());
+
+    m_sourceCodeLines.clear();
+    m_document->clear();
+
+    m_document->setPlainText(sourceCode);
+    m_document->setTextWidth(m_document->idealWidth());
+
     int maxLineNumber = 0;
     int minLineNumber = INT_MAX;
 
@@ -60,13 +68,6 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
     m_inclusiveCosts.initializeCostsFrom(results.inclusiveCosts);
 
     m_mainSourceFileName = disassemblyOutput.mainSourceFileName;
-
-    const auto sourceCode = QString::fromUtf8(file.readAll());
-
-    m_document->clear();
-
-    m_document->setPlainText(sourceCode);
-    m_document->setTextWidth(m_document->idealWidth());
 
     for (const auto& line : disassemblyOutput.disassemblyLines) {
         if (line.fileLine.line == 0 || line.fileLine.file != disassemblyOutput.mainSourceFileName) {
@@ -101,9 +102,18 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
     Q_ASSERT(minLineNumber < maxLineNumber);
 
     m_prettySymbol = disassemblyOutput.symbol.prettySymbol;
-    m_startLine = minLineNumber - 2;
-    m_lineOffset = minLineNumber - 1;
-    m_numLines = maxLineNumber - m_startLine;
+    m_startLine = minLineNumber - 1; // convert to index
+    m_numLines = maxLineNumber - minLineNumber + 1; // include minLineNumber
+
+    m_sourceCodeLines.reserve(m_numLines);
+
+    for (int i = m_startLine; i < m_startLine + m_numLines; i++) {
+        auto block = m_document->findBlockByLineNumber(i);
+        if (!block.isValid())
+            continue;
+
+        m_sourceCodeLines.push_back({block.text(), block.layout()->lineAt(0)});
+    }
 }
 
 QVariant SourceCodeModel::headerData(int section, Qt::Orientation orientation, int role) const
@@ -142,7 +152,7 @@ QVariant SourceCodeModel::data(const QModelIndex& index, int role) const
         return {};
     }
 
-    const auto fileLine = Data::FileLine(m_mainSourceFileName, index.row() + m_lineOffset);
+    const auto fileLine = Data::FileLine(m_mainSourceFileName, index.row() + m_startLine);
     if (role == FileLineRole) {
         return QVariant::fromValue(fileLine);
     } else if (role == Qt::ToolTipRole) {
@@ -154,19 +164,17 @@ QVariant SourceCodeModel::data(const QModelIndex& index, int role) const
             if (index.row() == 0) {
                 return m_prettySymbol;
             }
-            const auto block = m_document->findBlockByLineNumber(index.row() + m_startLine);
-            if (!block.isValid())
-                return {};
+
             if (role == SyntaxHighlightRole)
-                return QVariant::fromValue(block.layout()->lineAt(0));
-            return block.text();
+                return QVariant::fromValue(m_sourceCodeLines[index.row() - 1].line);
+            return m_sourceCodeLines[index.row() - 1].text;
         }
 
         if (index.column() == SourceCodeLineNumber) {
             return fileLine.line;
         }
 
-        auto cost = [role, id = index.row() + m_lineOffset](int type, const Data::Costs& costs) -> QVariant {
+        auto cost = [role, id = index.row() + m_startLine](int type, const Data::Costs& costs) -> QVariant {
             const auto cost = costs.cost(type, id);
             const auto totalCost = costs.totalCost(type);
             if (role == CostRole) {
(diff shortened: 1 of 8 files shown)


milianw:
I thought that was by design, to ensure the syntax highlighting is correct? how does this behave when we have code like

```
int
/* some comment */
MY_MACRO
actualFunction(
   int a,
   int b
)
throw(...)
{
}
```

lievenhey:
The debuginfos mark the line with `{` as the start of the function. So unless you manage to break KSyntaxHighlighting in one line we are save.

I thought about adding some logic to detect the start of the declaration, but there are too many edge cases to correctly detect this. The logic also would only work for C / C++ and would likely break if some other languages are used.
That is the reason why line 0 contains the symbol name.

milianw:
> The debuginfos mark the line with `{` as the start of the function. So unless you manage to break KSyntaxHighlighting in one line we are save.

This style is pretty common, and would then break, or?

```
longTypeName foo(int bar,
                                   double foo) {

}
```

> I thought about adding some logic to detect the start of the declaration, but there are too many edge cases to correctly detect this. The logic also would only work for C / C++ and would likely break if some other languages are used. That is the reason why line 0 contains the symbol name.

Exactly, which is why I actually like the old approach of just highlighting the full file. it's not that slow, or?

GitMensch:
I commonly have to wait around 5 seconds (small programs) to 30 seconds (big ones) - not tested so far with huge ones.
But in my case those are actually nearly all in one (generated) big function....

I personally don't see this as a big issue (other then #468 and much more #457 which do solve big issues I nearly always stumble over when working with hotspot).

I'd be most concerned what this change means to inline functions and functions that heavily use macros.
@lievenhey Would you mind sharing "before" and "after" results of your added test as image?

GitMensch:
Testing with current appImage yields in console flooded with "HSV parameters out of range" when hovering over the source code.

lievenhey:
> This style is pretty common, and would then break, or?
> 
> ```
> longTypeName foo(int bar,
>                  double foo) {
> 
> }
> ```

that doesn't brake syntax highlighting. It also resists
```
longTypeName foo(int bar, double foo) /*
*/ {

}
```

lievenhey:
> > Until now the whole file was included which wasted a lot of performance since KSyntaxHighlighter was run on the whole file.

>an issue?

I didn't really profile this but it seemed that highlighting a file with 1000+ lines to show a function with 10 lines. But now I feel stupid for spending time on something that has a negligent impact on performance.

Does this patch help in your case?

GitMensch:
> Does this patch help in your case?

Rechecked: With both CI builds 49 (current from this PR) and 50 (current from master) it takes ~27 seconds to open the disassembly on a "bad" sample with visible text. With master I've then seen ~1 second then the black text gets colored, with this PR it was nearly instant after the black-only text - but the ~27 seconds before coloring were identical.

As noted: I think the issue is xcb from "cloud" to local xserver.

GitMensch:
> Exactly, which is why I actually like the old approach of just highlighting the full file. it's not that slow, or?

I think the "slowness" is not that bad for the highlighting of the complete file.
But I do think the currently "everything highlighted, but often only minor parts - related to the current function - are assigned with costs" is sometimes confusing. Not sure if we can show them for the complete file / part that is highlighted.

src/models/sourcecodemodel.cpp
```diff
@@ -101,9 +102,16 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
     Q_ASSERT(minLineNumber < maxLineNumber);
 
     m_prettySymbol = disassemblyOutput.symbol.prettySymbol;
-    m_startLine = minLineNumber - 2;
-    m_lineOffset = minLineNumber - 1;
-    m_numLines = maxLineNumber - m_startLine;
+    m_startLine = minLineNumber - 1; // convert to index
+    m_numLines = maxLineNumber - minLineNumber + 1; // include minLineNumber
+
+    for (int i = m_startLine; i < m_startLine + m_numLines; i++) {
```
milianw, line 111: m_sourceCodeLines.reserve(m_numLines);

src/models/sourcecodemodel.cpp
```diff
@@ -242,3 +244,25 @@ void SourceCodeModel::setSysroot(const QString& sysroot)
 {
     m_sysroot = sysroot;
 }
+
+void SourceCodeModel::find(const QString& search, Direction direction, int offset)
+{
+    auto lines = m_sourceCodeLines;
+
+    std::rotate(lines.begin(), lines.begin() + offset, lines.end());
```
milianw, line 252: please document: why this rotate? I don't get why that's needed as it requires a re-allocation of `lines` (it detaches when you call `.begin()` on the non-const container)

generally, isn't it enough to just do a std::find_if on the subrange given by `offset`? and the direction would just decide whether we use forward or reverse iterators?

milianw (changes requested):
two minor nits, otherwise lgtm

milianw:
uhm did you push the wrong branch to this review? the discussion we had here doesn't match the code I just reviewed, or?

lievenhey:
I removed some stuff since it didn't work as expected.

milianw:
no, we should search in the focused widget via the correct shortcut context, and let the user use Tab to switch focus as needed

GitMensch:
> we should search in the focused widget via the correct shortcut context, and let the user use Tab to switch focus as needed

agreed in principle - but can we then please:
* focus to source code widget when entering the tab
* provide a fast way to switch between source code, disassembly and timeline widget

Background: the tab key switches between every control which includes, additional to those three:
* fast navigate
* syntax highlighting (2 times)
* new: search (2 times)
* timeline search
* timeline event source

BTW: CTRL+F doesn't position to search in timeline currently

src/models/sourcecodemodel.cpp
```diff
@@ -51,6 +51,14 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
     if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
         return;
 
+    QString sourceCode = QString::fromUtf8(file.readAll());
```
milianw, line 56: commenting here b/c github doesn't allow me to comment on the commit message: you don't call reverse/rotate anymore. just say that it makes searching easier and be done with it

src/models/sourcecodemodel.cpp
```diff
@@ -106,6 +108,8 @@ void SourceCodeModel::setDisassembly(const DisassemblyOutput& disassemblyOutput,
     m_startLine = minLineNumber - 1; // convert to index
     m_numLines = maxLineNumber - minLineNumber + 1; // include minLineNumber
 
+    m_sourceCodeLines.reserve(m_numLines);
```
milianw, line 111: this hunk belongs into the first patch of this series

src/models/sourcecodemodel.cpp
```diff
@@ -242,3 +246,39 @@ void SourceCodeModel::setSysroot(const QString& sysroot)
...
+{
+    auto searchFunc = [&search](const SourceCodeLine& line) {
+        return line.text.indexOf(search, 0, Qt::CaseInsensitive) != -1;
+    };
+
+    auto start = direction == Direction::Forward
+        ? (m_sourceCodeLines.begin() + offset)
+        : (m_sourceCodeLines.end() - (m_numLines - offset - 1)); // 1 one due to offset of the reverse iterator
+
+    auto it = direction == Direction::Forward
```
milianw, line 260: I believe this code here could be greatly clarified by hoisting it into a function that takes the range as iterators and returns a result index (or -1 if nothing is found).

then you call that with either forward or backward iterators. my hunch is that the code would be much more readable then since you'd only have to take care of the direction when calling the function, but not inside it.

anyhow, leave it as-is for now, I'll try to clean it up myself then

src/resultsdisassemblypage.cpp
```diff
@@ -171,6 +173,55 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+    connect(m_sourceCodeModel, &SourceCodeModel::resultFound, this, [this, colorScheme](const QModelIndex& index) {
+        auto palette = ui->searchEdit->palette();
+        m_currentSearchIndex = index;
+        palette.setBrush(QPalette::Text, colorScheme.foreground());
+        ui->searchEdit->setPalette(palette);
+        ui->sourceCodeView->setCurrentIndex(index);
+    });
+
+    connect(m_sourceCodeModel, &SourceCodeModel::noResultFound, this, [this, colorScheme]() {
+        auto palette = ui->searchEdit->palette();
```
milianw, line 219: DRY: can this share code with the super similar slot above? something like a setCurrentSearchIndex which you then call from here with an invalid index?

also: why is m_currentSearchIndex not set here?

src/resultsdisassemblypage.cpp
```diff
@@ -171,6 +173,55 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+        palette.setColor(QPalette::Text, Qt::red);
```
milianw, line 220: nack, don't hardcode colors. use KColorScheme and query it for a negative color

src/resultsdisassemblypage.cpp
```diff
@@ -310,17 +361,18 @@ void ResultsDisassemblyPage::showDisassembly(const DisassemblyOutput& disassembl
 
     ResultsUtil::hideEmptyColumns(m_callerCalleeResults.selfCosts, ui->assemblyView, DisassemblyModel::COLUMN_COUNT);
 
-    ResultsUtil::hideEmptyColumns(m_callerCalleeResults.selfCosts, ui->sourceCodeView,
-                                  SourceCodeModel::COLUMN_COUNT);
+    ResultsUtil::hideEmptyColumns(m_callerCalleeResults.selfCosts, ui->sourceCodeView, SourceCodeModel::COLUMN_COUNT);
```
milianw, line 364: unrelated?

src/resultsdisassemblypage.h
```diff
@@ -70,6 +70,9 @@ class ResultsDisassemblyPage : public QWidget
     // Model
     DisassemblyModel* m_disassemblyModel;
     SourceCodeModel* m_sourceCodeModel;
+    QModelIndex m_currentSearchIndex;
+    // Current chosen function symbol
+    Data::Symbol m_curSymbol;
```
milianw, line 75: unused?

milianw (changes requested):

milianw:
> > we should search in the focused widget via the correct shortcut context, and let the user use Tab to switch focus as needed
> 
> agreed in principle - but can we then please:
> 
>     * focus to source code widget when entering the tab
> 
>     * provide a fast way to switch between source code, disassembly and timeline widget
> 
> 
> Background: the tab key switches between every control which includes, additional to those three:
> 
>     * fast navigate
> 
>     * syntax highlighting (2 times)
> 
>     * new: search (2 times)
> 
>     * timeline search
> 
>     * timeline event source
> 
> 
> BTW: CTRL+F doesn't position to search in timeline currently

I think what you are looking for here, besides proper tab order (which is probably not a given, have to double check): ALT-modifiers to trigger quick focus changes. That way, you could press `ALT+...` to focus e.g. the assembly then `CTRL+F` to search in there. and then `ALT+...` to focus the source code followed by another `CTRL+F` to search in that one then.

src/resultsdisassemblypage.cpp
lievenhey: yes

src/models/sourcecodemodel.h
```diff
@@ -23,6 +23,12 @@ class Repository;
 class Definition;
 }
 
+struct SourceCodeLine
```
milianw, line 26: add `Q_DECLARE_TYPEINFO(SourceCodeLine, Q_MOVABLE_TYPE)`

milianw (approved):
would be cool if you can add the missing typeinfo and then submit it

Issue #476, opened 2023-03-24
postpone display initialization to skip it when in console only mode


GitMensch:
**Is your feature request related to a problem? Please describe.**
Using hotspot in CI commonly means there is no (virtual) display available.

**Describe the solution you'd like**
I'd like to be able to run hotspot's `--exportTo` via CI without hacky setups. Ideally the QT display init could be postponed after the command line options are processed - and in the case of `--exportTo` or `--version` be not done at all.

**Describe alternatives you've considered**
Use recorded perf.data instead - but that's bigger and needs more post-processing on the client.

**Additional context**
~~~
$> ./hotspot-v1.4.1-46-g3b42145-x86_64.AppImage  --exportTo perfparser.PROG  perf.data.PROG
qt.qpa.xcb: could not connect to display
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb.
~~~

PR #481, opened 2023-04-12
allow hotspot to run without gui if --exportTo is supplied

lievenhey:
This allows automatic conversion from perf.data files to perfparser files without the need of an display. For example in an CI

closes: #476

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -18,6 +18,7 @@
 #include "hotspot-config.h"
 #include "mainwindow.h"
 #include "models/data.h"
+#include "parsers/perf/perfparser.h"
 #include "settings.h"
 #include "util.h"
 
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -203,32 +217,35 @@ int main(int argc, char** argv)
...
                            .arg(file, url.toDisplayString(QUrl::PrettyDecoded | QUrl::PreferLocalFile))
                     << Qt::endl;
-                app.exit(0);
+                app->exit(0);
             });
             auto destination = QUrl::fromUserInput(parser.value(exportTo), QDir::currentPath(), QUrl::AssumeLocalFile);
-            window->saveAs(file, destination);
-            return app.exec();
-        } else {
+            perfParser.startParseFile(file);
```
milianw, line 231: ...here this could do

```
auto perfParser = new PerfParser(app.data());
perfParser->startParseFile(file);
perfParser.exportResults(destination);
```

that said: is the order of the two commands even safe? what if parsing takes longer? what if it is finished already?

src/mainwindow.cpp
```diff
@@ -94,10 +94,10 @@ int firstAvailableIde()
 }
 }
 
-MainWindow::MainWindow(QWidget* parent)
+MainWindow::MainWindow(PerfParser* parser, QWidget* parent)
```
milianw, line 97: undo these changes again

milianw (changes requested):

milianw (approved):

milianw:
thanks!

PR #484, opened 2023-05-04
New elevated privilege system

zeno-endemann-kdab:
This is currently only very minimally tested


src/elevateprivilegeshelper.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include <QObject>
+#include <QString>
+#include <QStringList>
+
+#include <unistd.h>
+
+class QSocketNotifier;
+
+class InitiallyStoppedProcess
```
milianw, line 18: could you do me a favor and add basic api documentation to this class and its methods? I guess run can only be executed after reset?

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+    int wstatus;
+    waitpid(m_pid, &wstatus, WUNTRACED);
+    if (!WIFSTOPPED(wstatus)) {
+        m_pid = -1;
+        return false;
+    }
+
+    // continue
+
+    ::kill(m_pid, SIGCONT);
```
milianw, line 78: here and below: please add error handling

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+        args.emplace_back(opt.toLocal8Bit());
+    const auto wd = workingDirectory.toLocal8Bit();
+
+    std::vector<char*> argsArray(args.size() + 1);
+    for (size_t i = 0; i < args.size(); ++i)
+        argsArray[i] = args[i].data();
+    argsArray.back() = nullptr;
+
+    // fork
+    m_pid = fork();
```
milianw, line 41: forking means we COW the whole address space which can be really slow when that is large. that would e.g. be the case when you start hotspot, analyze some large file and then rerun the record from there.

is there a way this could use QProcess such that we can leverage vfork/posix_spawn when available? or is there maybe something like that in boost which we could use to get this behavior in a more efficient manner?

see also: https://codereview.qt-project.org/c/qt/qtbase/+/417829

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <vector>
+
+InitiallyStoppedProcess::~InitiallyStoppedProcess()
+{
+    kill();
+}
+
+bool InitiallyStoppedProcess::reset(const QString& exePath, const QStringList& exeOptions,
```
milianw, line 21: this could be renamed to prepareExecAndWait or similar

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+        qFatal("Failed to exec %s", argsArray[0]);
+    } else if (m_pid < 0) {
+        qCritical("Failed to fork (?)");
+        return false;
+    }
+
+    return true;
+}
+
+bool InitiallyStoppedProcess::run()
```
milianw, line 62: this could be renamed to continueExec

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+    if (m_pid <= 0)
+        return false;
+
+    // wait for child to be stopped
+
+    int wstatus;
+    waitpid(m_pid, &wstatus, WUNTRACED);
```
milianw, line 70: error handling?

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+PerfControlFifoWrapper::~PerfControlFifoWrapper()
+{
+    close();
+}
+
+bool PerfControlFifoWrapper::open()
+{
+    close();
+
+    QString fifoParentPath = QStandardPaths::writableLocation(QStandardPaths::RuntimeLocation);
```
milianw, line 106: auto

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+    if (fifoParentPath.isEmpty())
+        fifoParentPath = QStandardPaths::writableLocation(QStandardPaths::TempLocation);
```
milianw, line 108: why this fallback, shouldn't the above be sufficient?

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    const auto fifoBasePath =
+        QStringLiteral("%1/hotspot-%2-%3-perf")
```
milianw, line 111: QLatin1String(...).arg()

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+            .arg(fifoParentPath, QString::number(getpid()), QUuid::createUuid().toString().mid(1, 6));
```
milianw, line 112: what is this short uuid thingy, is that generating a 5 digit random string? can't you do that directly?

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+    m_ctlFifoPath = fifoBasePath + QStringLiteral("-control.fifo");
```
milianw, line 113: here and in the line below: QLatin1String

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+    m_ackFifoPath = fifoBasePath + QStringLiteral("-ack.fifo");
+
+    if (mkfifo(m_ctlFifoPath.toLocal8Bit().data(), 0600) != 0) {
+        qCritical() << "Cannot create fifo" << m_ctlFifoPath;
```
milianw, line 117: here and elsewhere: please use categorized logging

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+    return true;
+}
+
+void PerfControlFifoWrapper::start()
+{
+    if (m_ctlFifoFd < 0)
+        return;
+
+    m_ackReady = new QSocketNotifier(m_ackFifoFd, QSocketNotifier::Read);
```
milianw, line 144: is this leaked? don't you need to set a parent?

should it be asserted/verified that the m_ackReady wasn't created yet?

src/elevateprivilegeshelper.h
```diff
@@ -0,0 +1,76 @@
...
+    bool reset(const QString& exePath, const QStringList& exeOptions, const QString& workingDirectory);
+    bool run();
+    void terminate();
+    void kill();
+
+private:
+    pid_t m_pid = -1;
+};
+
+class PerfControlFifoWrapper : public QObject
```
milianw, line 41: please put these classes into separate files, they are not tightly coupled after all

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+    if (fifoParentPath.isEmpty())
+        fifoParentPath = QStandardPaths::writableLocation(QStandardPaths::TempLocation);
+
+    const auto fifoBasePath =
+        QStringLiteral("%1/hotspot-%2-%3-perf")
+            .arg(fifoParentPath, QString::number(getpid()), QUuid::createUuid().toString().mid(1, 6));
+    m_ctlFifoPath = fifoBasePath + QStringLiteral("-control.fifo");
+    m_ackFifoPath = fifoBasePath + QStringLiteral("-ack.fifo");
+
+    if (mkfifo(m_ctlFifoPath.toLocal8Bit().data(), 0600) != 0) {
```
milianw, line 116: DRY: can you introduce a reusable function that does the mkfifo + open and then call that once for the ack and once for the ctl fifo?

also note that you are converting the QString to QByteArray twice for each fifo

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+    m_ackFifoFd = ::open(m_ackFifoPath.toLocal8Bit().data(), O_RDONLY | O_NONBLOCK);
+    if (m_ackFifoFd < 0) {
+        qCritical() << "Cannot open fifo" << m_ackFifoPath;
+        return false;
+    }
+
+    return true;
+}
+
+void PerfControlFifoWrapper::start()
```
milianw, line 139: for clarity I'd rename this to `requestStart` as it just triggers an async request

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+    if (m_ctlFifoFd < 0)
```
milianw, line 141: emit a warning?

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+        read(m_ackFifoFd, buf, sizeof(buf));
+        emit started();
+        m_ackReady->disconnect(this);
+    });
+
+    const char start_cmd[] = "enable\n";
+    write(m_ctlFifoFd, start_cmd, sizeof(start_cmd) - 1);
+}
+
+void PerfControlFifoWrapper::stop()
```
milianw, line 156: dito, requestStop

src/elevateprivilegeshelper.cpp
```diff
@@ -0,0 +1,186 @@
...
+{
+    if (m_ctlFifoFd < 0)
+        return;
```
milianw, line 159: emit warning

milianw (changes requested):
great thanks, here's just some code review. I'll test the functionality in the next days

src/perfrecord.cpp
```diff
@@ -247,26 +157,58 @@ void PerfRecord::startRecording(const QStringList& perfOptions, const QString& o
...
+        }
+        options +=
+            {QStringLiteral("--control"),
+             QStringLiteral("fifo:%1,%2").arg(m_perfControlFifo.controlFifoPath(), m_perfControlFifo.ackFifoPath())};
+
+        { // pre-create output
+            // elevated perf will obviously create a root-owned output by default, but testing revealed that
+            // perf will write into a pre-existing file if it is empty (without changing ownership)
+            const QString bakPath(outputPath + QStringLiteral(".old"));
+            QFile::remove(bakPath);
```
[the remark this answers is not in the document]
zeno-endemann-kdab: From the Qt docs: If a file with the name newName already exists, rename() returns false (i.e., QFile will not overwrite it).
I don't really think error handling is necessary for this, it is fine if any of this fails.

src/elevateprivilegeshelper.cpp
zeno-endemann-kdab: sure this could be improved. Don't know what's available in boost for that. vfork seems a bit dangerous to me, for posix_spawn I suppose one would need to move this initially-stopped-exec functionality in a separate helper binary or something.

src/elevateprivilegeshelper.cpp
zeno-endemann-kdab: QStandardPaths documentation states that RuntimeLocation can be empty while TempLocation cannot, that's why I added the fallback

src/elevateprivilegeshelper.cpp
lievenhey: we can only use vfork if we create a new binary, since vfork only allows for `__exit()` and `exec*` calls. I think `clone` with `CLONE_WM | CLONE_VFORK` might be the easiest way to solve this.

src/elevateprivilegeshelper.cpp
lievenhey: I replaced it by `QString::number(QRandomGenerator::generate() % 1000000, 16)`

src/elevateprivilegeshelper.cpp
zeno-endemann-kdab: From my understanding, VFORK suspends the calling process until exec is called, which we do not want here at all. I don't think VFORK is usable for this. posix_spawn and a helper process should work though.

lievenhey:
lets continue the discussion in the update pr #493

PR #486, opened 2023-05-17
Set source path

lievenhey:
this patch add allows the user to add a custom source path which will
overwrite the source path provided in the binary
this makes the disassembler, ... work if the source code directory was
moved after creating the binary

closes: #418

### DIFF
--- a/src/models/disassemblyoutput.cpp
+++ b/src/models/disassemblyoutput.cpp
@@ -121,6 +121,32 @@ QString findBinaryForSymbol(const QStringList& debugPaths, const QStringList& ex
 }
 }
 
+// not in an anonymous namespace so we can test this
+QString findSourceCodeFile(const QString& originalPath, const QStringList& sourceCodePaths, const QString& sysroot)
+{
+    if (QFile::exists(originalPath)) {
+        return originalPath;
+    }
+
+    QString sysrootPath = sysroot + QDir::separator() + originalPath;
+    if (QFile::exists(sysrootPath)) {
+        return sysrootPath;
+    }
+
+    for (const auto& sourcePath : sourceCodePaths) {
+        for (auto it = originalPath.begin(); it != originalPath.end();
+             it = std::find(++it, originalPath.end(), QDir::separator())) {
+            QString path = sourcePath + QDir::separator() + QString(it, std::distance(it, originalPath.end()));
+            QFileInfo info(path);
+            if (info.exists()) {
+                return info.canonicalFilePath();
+            }
+        }
+    }
+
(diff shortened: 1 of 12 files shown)


src/models/disassemblyoutput.cpp
```diff
@@ -119,6 +119,52 @@ QString findBinaryForSymbol(const QStringList& debugPaths, const QStringList& ex
 
     return {};
 }
+
+int longestCommonLengthFromEnd(const QString& a, const QString& b)
+{
+    const auto lengthA = a.length();
+    const auto lengthB = b.length();
+    int shortestLength = std::min(lengthA, lengthB);
```
milianw, line 127: `const auto`

...but:

I believe this whole thing is basically an application of std::mismatch, no? see

https://en.cppreference.com/w/cpp/algorithm/mismatch

src/models/disassemblyoutput.cpp
```diff
@@ -119,6 +119,52 @@ QString findBinaryForSymbol(const QStringList& debugPaths, const QStringList& ex
...
+    QString candidate;
+
+    QFileInfo info(originalPath);
+
+    for (const auto& path : sourceCodePaths) {
+        // we want the longest matching path since some projects are organized like this:
+        // - liba/lib.c
+        // - libb/lib.c
+
+        QDirIterator it(path, {info.fileName()}, QDir::NoFilter, QDirIterator::Subdirectories);
```
milianw, line 156: if someone adds `/` to the source code paths then this will become _really_ slow as we'll be basically looping the whole file system for every source code file we are looking up?

instead, we should:

- first lookup the originalPath directly
- then fallback to the originalPath in the sysroot
- then loop over the source paths in this manner:

given a path `/foo/bar/file.c` and a source path `$SP` you'd start lookup:

```
$SP/foo/bar/file.c
$SP/bar/file.c
$SP/file.c
```

i.e. pop the dirs from the front of the source path until only the filename remains and attach that to the source dir.

I think ideally the loop should also be such that we first lookup the full paths in all source dirs, then pop the dir from front of the source path and try again all source dirs. that way the most specific path in any of the source dirs wins, which sounds correct to me?

milianw (dismissed):

src/models/disassemblyoutput.cpp
lievenhey: I was sure something like this was in the standard, but I could find anything. Seems like I used the wrong search words

milianw (changes requested):

milianw (approved):
lgtm, feel free to push once the CI has tested it

Issue #483, opened 2023-04-27
KDDockWidgets warning during startup


GitMensch:
When starting hotspot (current AppImage) I get 

 ~~~
 kf.coreaddons: "Could not find plugin kgraphviewerpart"
void KDDockWidgets::Config::setFlags(KDDockWidgets::Config::Flags) Only use this function at startup before creating any DockWidget or MainWindow ; These are already created:  ("results") ("summary", "bottomUp", "topDown", "flameGraph", "callerCallee", "disassembly", "frequency", "timeLine") QVector()
~~~

I can't remember to have seen the second message before, so creating this issue...

PR #487, opened 2023-05-17
fix kddockwidgets warning

lievenhey:
fixes: #483


milianw:
please run clang-format to fix the format and then add your patch here

GitMensch:
@milianw looks like the requested formatting was done

PR #493, opened 2023-06-30
New elevate system

lievenhey:


src/initiallystoppedprocess.cpp
```diff
@@ -0,0 +1,115 @@
...
+        args.emplace_back(opt.toLocal8Bit());
+    const auto wd = workingDirectory.toLocal8Bit();
+
+    std::vector<char*> argsArray(args.size() + 1);
+    for (size_t i = 0; i < args.size(); ++i)
+        argsArray[i] = args[i].data();
+    argsArray.back() = nullptr;
+
+    // fork
+    m_pid = fork();
```
lievenhey, line 59: @milianw 
> forking means we COW the whole address space which can be really slow when that is large. that would e.g. be the case when you start hotspot, analyze some large file and then rerun the record from there.
>
> is there a way this could use QProcess such that we can leverage vfork/posix_spawn when available? or is > there maybe something like that in boost which we could use to get this behavior in a more efficient manner?
> 
> see also: https://codereview.qt-project.org/c/qt/qtbase/+/417829

@zeno-endemann-kdab 
> sure this could be improved. Don't know what's available in boost for that. vfork seems a bit dangerous to me, for posix_spawn I suppose one would need to move this initially-stopped-exec functionality in a separate helper binary or something.

@lievenhey 
> we can only use vfork if we create a new binary, since vfork only allows for __exit() and exec* calls. I think clone with CLONE_WM | CLONE_VFORK might be the easiest way to solve this.

@zeno-endemann-kdab 
> From my understanding, VFORK suspends the calling process until exec is called, which we do not want here at all. I don't think VFORK is usable for this. posix_spawn and a helper process should work though.

milianw: I think vfork is fine, why shouldn't we be halting the process until exec is called? that should be done quickly after all.

zeno-endemann-kdab: Maybe I misunderstand something, but we fork here to immediately stop the child before exec-ing so that the parent process has time to attach perf. Making the parent wait for the child to exec would obviously deadlock then.

lievenhey: from the docs:
> vfork() differs from [fork(2)](https://man7.org/linux/man-pages/man2/fork.2.html) in that the calling thread is
       suspended until the child terminates (either normally, by calling
       [_exit(2)](https://man7.org/linux/man-pages/man2/_exit.2.html), or abnormally, after delivery of a fatal signal), or it
       makes a call to [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html).

`SIGSTOP` is not a fatal signal

zeno-endemann-kdab: One can of course create a helper binary ("exec_stopped" or something) that contains the sigstop-then-exec logic, and that one could be vfork'ed safely, if that is what you meant.

milianw: indeed, my bad. anyhow, let's ignore this for now and I'll try to look into alternatives to `fork` in the future

src/initiallystoppedprocess.cpp
```diff
@@ -0,0 +1,115 @@
...
+    if (m_pid == 0) { // inside child process
+        // change working dir
+        if (!wd.isEmpty() && chdir(wd.data()) != 0)
+            qFatal("Failed to change working directory to %s", wd.data());
+
+        // stop self
+        raise(SIGSTOP);
+
+        // exec
+        execvp(argsArray[0], argsArray.data());
```
milianw, line 70: error handling, and pipe in `std::strerror(errno)`

src/initiallystoppedprocess.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <sys/wait.h>
+#include <vector>
+
+namespace {
+Q_LOGGING_CATEGORY(initiallystoppedprocess, "hotspot.initiallystoppedprocess")
+
+void sendSignal(pid_t pid, int signal)
+{
+    if (!kill(pid, signal)) {
+        qCCritical(initiallystoppedprocess, "Failed to send %u to %u", signal, pid);
```
milianw, line 29: pipe in `std::strerror(errno)`

src/perfcontrolfifowrapper.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <QObject>
+#include <QSocketNotifier>
+#include <QString>
+
+#include <memory>
+
+class QSocketNotifier;
+
+class PerfControlFifoWrapper : public QObject
```
milianw, line 18: please add a short api documentation blurb:

```
/**
 * Wrapper for the control and ack FIFOs for perf record.
 * 
 * For more information, refer to `man perf record` and search for `--control=fifo:`.
 */
```

src/perfrecord.cpp
```diff
@@ -28,32 +28,36 @@
...
     , m_perfRecordProcess(nullptr)
-    , m_elevatePrivilegesProcess(nullptr)
     , m_outputPath()
     , m_userTerminated(false)
 {
+    connect(&m_perfControlFifo, &PerfControlFifoWrapper::started, this,
+            [this]() { m_targetProcessForPrivilegedPerf.continueStoppedProcess(); });
+
+    connect(&m_perfControlFifo, &PerfControlFifoWrapper::noFIFO, this, [this] {
+        emit recordingFailed(QStringLiteral("Fifo missing")); /*TODO: better message*/
```
milianw, line 59: remove the TODO, and replace this with a more generic error like

```
Failed to start process, broken control FIFO
```

milianw (changes requested):
some small nits, mostly around error handling, otherwise this lgtm and can go in then

milianw (approved):

PR #494, opened 2023-06-30
Disassembler inlining

lievenhey:
this adds a simple detection for inlined code to elide it from the assembly view.

grafik

I am not sure if this works correctly so it would be nice if someone has a minimal example to test this.
This currently brakes syntax highlighting and code navigation

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,7 @@ remove_definitions(-DQT_NO_SIGNALS_SLOTS_KEYWORDS)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 add_subdirectory(models)
+add_subdirectory(disassembler)
 
 ecm_install_icons(
     ICONS
@@ -89,10 +90,7 @@ if(QCustomPlot_FOUND)
     set(HOTSPOT_SRCS ${HOTSPOT_SRCS} frequencypage.cpp)
 endif()
 
-add_executable(
-    hotspot
-    ${HOTSPOT_SRCS}
-)
+add_executable(hotspot ${HOTSPOT_SRCS})
 
 target_link_libraries(
     hotspot
@@ -113,6 +111,7 @@ target_link_libraries(
     KF5::IconThemes
     KF5::Parts
     models
+    disassembler
     PrefixTickLabels
     KDAB::kddockwidgets
 )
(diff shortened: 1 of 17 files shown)


lievenhey:
TODO: detect inline same file

lievenhey:
The inline detection is now working.

Output:
grafik

example.c:
```c
#include <stdio.h>

inline int square(int i) {
        return i * i;
}

int main() {
        for (int i = 0; i < 10000; i++) {
                int sum = 0;
                for (int j = 0; j < i; j++) {
                        sum += square(j);
                }
                printf("Sum: %d\n", sum);
        }
}
```

GitMensch:
If it does what I think it does then: yay. Can you please share what the previous build outputs to be able to compare the results?

lievenhey:
before:
image
after:
image
(click to expand)

GitMensch:
Any idea why this [fails on ArchLinux with clang](https://github.com/KDAB/hotspot/actions/runs/5658820452/job/15330979400?pr=494)?

lievenhey:
no, seems to only happen on the ci

lievenhey:
found it
`objdump ...` on non clazy target:
```asm
00000000000012a0 <main>:
main():
/home/lieven/KDAB/hotspot/tests/test-clients/cpp-recursion/main.cpp:19
```
`objdump ...` on clazy target:
```asm
00000000000012f0 <main>:
std::basic_ios<char, std::char_traits<char> >::widen(char) const:
/opt/hotspot/tests/test-clients/cpp-recursion/main.cpp:19
```
I used the the first instance of `something():` as function name. Since on clazy it is `std::basic_ios<char, std::char_traits<char> >::widen(char) const` it thinks `main` is an inlined function

time to rewrite it using libdwarf

GitMensch:
related to #495

GitMensch:
Would that need a doc update (screenshots, maybe)?
Is libdwarf a new dependency to be added?

GitMensch:
so libwarf was added which was thought to fix the CI, but it is still failing :-(

GitMensch:
I'd like to give this a test - can you rebase that, please?

GitMensch:
I'd like to give this a test - can you rebase that, please?

Apart from the actual inline view it will be interesting to see how the parsing with libdwarf effects the necessary time to open the disassembly view (while the time currently needed seems to be related to (the use of) Qt #505)

GitMensch:
This will still take some time. We want to move the logic to the `perfparser` since it has all the logic for finding binaries etc.

_Originally posted by @lievenhey in https://github.com/KDAB/hotspot/issues/518#issuecomment-1759596746_

That sounds very reasonable. If I understand that correctly, this means that perfparser may use libdwarf to do this and more, and then perfparser will be used both for handling the actual "perf.data" (as now) and additionally do "something special" for the disassembly, but not replacing objdump, right?

lievenhey:
Disassembling will still be handled by objdump. We just use libdwarf for additional information.

lievenhey:
Well for one it doesn't work that good and I hadn't much time to work on this.
We recently discussed using a real disassembly library so this will soon be out of date.
The reason is, that we are currently try to parse the human readable output of `objdump`, but that doesn't work so well with more complex information.

lievenhey:
closing in favour of a rewrite /close

GitMensch:
*holding my breath* Is someone actually working on that rewrite or is this PRs approach "just considered obsolete"?

lievenhey:
Yes, it can already detect inlined coded and interleaves it with the other code.
grafik

PR #497, opened 2023-07-07
add option to set custom source paths

lievenhey:
this patch add allows the user to add a custom source path which will overwrite the source path provided in the binary
this makes the disassembler, ... work if the source code directory was moved after creating the binary

closes: #418

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,8 @@ Options:
                            libraries.
   --appPath <path>         Path to folder containing the application executable
                            and libraries.
+  --sourcePaths <paths>    Colon separated list of extra paths to the source
+                           code.
   --arch <path>            Architecture to use for unwinding.
   --exportTo <path>        Path to .perfparser output file to which the input
                            data should be exported. A single input file has to
@@ -325,6 +327,9 @@ which source code line.
 For easier navigation, you can simply click on a line and the other view will jump to it.
 You can follow function calls with a double click.
 In the sourcecode view you can press ctrl+f or press the search icon to open a search window.
+
+If you have the sources in different directory, you can use `--sourcePaths` or the settings to
+select tell the disassembler to search there for the source code.
 
 ## Known Issues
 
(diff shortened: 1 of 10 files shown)


GitMensch:
Tested with the CI result:

* opened recording
* used Disassembly on a function - after a while that was shown, but not source, the source reference pointed to a directory not existing any more
* gone to settings, added the new source path (select, then add, see it in the list) - Disassembly view was not refreshed
* used Disassembly on a different function - the source reference still points to the same not existing directory, not the configured one where the source exists
* used Disassembly on a different function which has its source folder unchanged, works
* restarted HotSpot, sourcePath was not restored from last session, stopped it
* started the AppImage `./hotspot-v1.4.1-74-ge66d066-x86_64.AppImage --sourcePaths=/my/source`, got an error
     >hotspot: Unknown option 'sourcePaths'.
  
  So sadly: not successful.

lievenhey:
thanks for testing. I fixed the `--sourcePaths` option and added reloading if `sourcePaths` changes

GitMensch:
>  fixed the `--sourcePaths` option and added reloading if `sourcePaths` changes

Thanks, this works nearly as expected now.
It _seems_ that while the entries that were included in the GUI are not shown in the configuration after a reload of the application, they are still active (which is not necessarily bad, I actually prefer _those_ to be saved) - maybe they are needed to be included when the setting page is opened?

Not sure how you'd show "persistent" and "per session" entries there (I'd personally mostly need "per session" or "per configuration file", because different projects have the same binary name but the source is in different directories) .

Note that in any case this PR missed to add the option (now also in `--help`) to https://github.com/KDAB/hotspot/blob/master/README.md#command-line-options

milianw:
@GitMensch  I think the per-session vs. user settings is a general issue we need to tackle somehow - can you please open a ticket to track that?

the rest of this pull request is now mergeable I believe

milianw (approved):

PR #508, opened 2023-09-21
Tracepoint

lievenhey:
Show tracepoints in TimeLineWidget


lievenhey:
image

src/models/eventmodel.cpp
```diff
@@ -79,13 +81,26 @@ int EventModel::rowCount(const QModelIndex& parent) const
...
     case Tag::Overview:
-        return (parent.row() == 0) ? m_data.cpus.size() : m_processes.size();
+        switch (parent.row()) {
+        case 0:
+            return m_data.cpus.size();
+        case 1:
+            return m_processes.size();
+        case 2:
+            return m_data.tracepoints.size();
+        default:
```
milianw, line 102: remove the default, turn the assert into a `Q_UNREACHABLE()` below the switch

src/models/eventmodel.cpp
```diff
@@ -25,7 +25,9 @@ enum class Tag : quint8
     Overview = 2,
     Cpus = 3,
     Processes = 4,
-    Threads = 5
+    Threads = 5,
+    Tracepoints = 6,
+    LAST_TAG,
```
milianw, line 33: I dislike how that pollutes all switch statements. please instead introduce just below the Tag enum:

```
constexpr auto LAST_TAG = Tag::Tracepoints;
```

src/models/eventmodel.cpp
```diff
@@ -79,13 +81,26 @@ int EventModel::rowCount(const QModelIndex& parent) const
...
     case Tag::Cpus:
     case Tag::Threads:
+    case Tag::Tracepoints:
+        return 0;
         break;
     case Tag::Processes:
         return m_processes.value(parent.row()).threads.size();
     case Tag::Overview:
-        return (parent.row() == 0) ? m_data.cpus.size() : m_processes.size();
+        switch (parent.row()) {
```
milianw, line 93: ideally, we also introduce an enum here, i.e.:

```
switch (static_cast<OverViewRow>(parent.row())
```

src/models/eventmodel.cpp
```diff
@@ -137,17 +152,28 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
...
     auto tag = dataTag(index);
 
+    Q_ASSERT(static_cast<int>(tag) < static_cast<int>(Tag::LAST_TAG));
+
     if (tag == Tag::Invalid || tag == Tag::Root) {
         return {};
     } else if (tag == Tag::Overview) {
         if (role == Qt::DisplayRole) {
-            return index.row() == 0 ? tr("CPUs") : tr("Processes");
+            switch (index.row()) {
```
milianw, line 166: see above, enum

src/models/eventmodel.cpp
```diff
@@ -137,17 +152,28 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
 
     auto tag = dataTag(index);
 
+    Q_ASSERT(static_cast<int>(tag) < static_cast<int>(Tag::LAST_TAG));
```
milianw, line 160: <= once my constexpr LAST_TAG is used

src/models/eventmodel.cpp
```diff
@@ -39,7 +41,7 @@ quintptr combineDataTag(Tag tag, quintptr data)
 Tag dataTag(quintptr internalId)
 {
     auto ret = (internalId << DATATAG_UNSHIFT) >> DATATAG_UNSHIFT;
-    if (ret > static_cast<quintptr>(Tag::Threads))
+    if (ret > static_cast<quintptr>(Tag::LAST_TAG))
```
milianw, line 46: this is wrong in your code, will get correct with my code (your patch as-is would need `>=`)

just saying, keep this as-is - just change `LAST_TAG` as indicated above

src/models/eventmodel.cpp
```diff
@@ -79,13 +81,26 @@ int EventModel::rowCount(const QModelIndex& parent) const
...
+        case 1:
+            return m_processes.size();
+        case 2:
+            return m_data.tracepoints.size();
+        default:
+            Q_ASSERT(false);
+        }
     case Tag::Root:
-        return 2;
+        return m_data.tracepoints.isEmpty() ? 2 : 3;
```
milianw: ideally these magic numbers also get turned into constexpr or enums

src/models/eventmodel.cpp
```diff
@@ -137,17 +152,28 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
...
+            switch (index.row()) {
+            case 0:
+                return tr("CPUs");
+            case 1:
+                return tr("Processes");
+            case 2:
+                return tr("Tracepoints");
+            }
         } else if (role == Qt::ToolTipRole) {
             if (index.row() == 0) {
```
milianw, line 177: switch here too like above

src/models/eventmodel.cpp
```diff
@@ -204,47 +230,123 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
 
     const Data::ThreadEvents* thread = nullptr;
     const Data::CpuEvents* cpu = nullptr;
+    const Data::TracepointEvents* tracepoint = nullptr;
 
     if (tag == Tag::Cpus) {
         cpu = &m_data.cpus[index.row()];
-    } else {
-        Q_ASSERT(tag == Tag::Threads);
+        Q_ASSERT(cpu);
```
milianw, line 246: this will never be false, remove

src/models/eventmodel.cpp
```diff
@@ -204,47 +230,123 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
...
+    } else if (tag == Tag::Threads) {
         const auto process = m_processes.value(tagData(index.internalId()));
         const auto tid = process.threads.value(index.row());
         thread = m_data.findThread(process.pid, tid);
         Q_ASSERT(thread);
+    } else if (tag == Tag::Tracepoints) {
+        tracepoint = &m_data.tracepoints[index.row()];
+        Q_ASSERT(tracepoint);
```
milianw, line 254: dito, bogus - remove

src/models/eventmodel.cpp
```diff
@@ -204,47 +230,123 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
...
     }
 
     if (role == ThreadStartRole) {
-        return thread ? thread->time.start : m_time.start;
+        switch (tag) {
+        case Tag::Threads:
+            return thread->time.start;
+        default:
```
milianw, line 265: please don't use `default`, list the cases explicitly and add a `Q_UNREACHABLE()` below the switch. repeats below

src/models/eventmodel.cpp
```diff
@@ -336,17 +461,22 @@ QModelIndex EventModel::index(int row, int column, const QModelIndex& parent) co
...
+    case Tag::LAST_TAG:
     case Tag::Invalid: // leaf / invalid -> no children
     case Tag::Cpus:
+    case Tag::Tracepoints:
     case Tag::Threads:
         break;
     case Tag::Root: // root has the 1st level children: Overview
         return createIndex(row, column, static_cast<quintptr>(Tag::Overview));
     case Tag::Overview: // 2nd level children: Cpus and the Processes
         if (parent.row() == 0)
```
milianw: dito, switch

src/models/eventmodel.cpp
```diff
@@ -365,6 +496,9 @@ QModelIndex EventModel::parent(const QModelIndex& child) const
         return createIndex(0, 0, static_cast<quintptr>(Tag::Overview));
```
milianw, line 513: replace the magic numbers for the rows here too

src/parsers/perf/perfparser.cpp
```diff
@@ -872,6 +872,10 @@ class PerfParserPrivate : public QObject
         buildPerLibraryResult();
         buildCallerCalleeResult();
 
+        for (auto tracepoint = tracepoints.begin(); tracepoint != tracepoints.end(); tracepoint++) {
```
milianw, line 875:
```
for (auto it = tracepoints.constBegin(), end = tracepoints.constEnd(); it != end; ++it) {
    eventResult.tracepoints.push_back({strings[tracepoint.key()], tracepoint.value()});
}
```

src/parsers/perf/perfparser.cpp
```diff
@@ -1089,13 +1093,12 @@ class PerfParserPrivate : public QObject
...
-                tracepoint.time = event.time;
-                tracepoint.name = strings.value(attribute.name.id);
-                if (tracepoint.name != QLatin1String("sched:sched_switch")) {
-                    // sched_switch events are handled separately already
-                    tracepointResult.tracepoints.push_back(tracepoint);
+                if (eventResult.tracepointEventCostId == -1) {
+                    eventResult.tracepointEventCostId =
+                        addCostType(QStringLiteral("Tracepoint"), Data::Costs::Unit::Tracepoint);
                 }
+                auto& tracepointList = tracepoints[attribute.name.id];
```
milianw, line 1100: what's with the old code that handles sched switch?

milianw (changes requested):

src/models/timelinedelegate.cpp
```diff
@@ -133,6 +134,10 @@ Data::Events::const_iterator findEvent(const Data::Events::const_iterator& begin
     auto it = std::lower_bound(begin, end, time, byTime);
     // it points to the first item for which our predicate returns false, we want to find the item before that
     // so decrement it if possible or return begin otherwise
+    // if only one event is recorded, it will point to end it->time in this case will case asan to complain
```
milianw, line 137: grammar and duplicated words. rephrase. then inline the code:

```
// if only one event is recorded, `it` will point to end and we cannot access `it-time`
return (it == begin || (it != end && it->time == time)) ? it : (it - 1);
```

src/models/eventmodel.h
```diff
@@ -71,9 +71,13 @@ class EventModel : public QAbstractItemModel
         QString name;
     };
 
+public:
+    void addToFavourites(const QModelIndex& index);
```
milianw, line 75: use US spelling: Favorites

src/models/eventmodel.cpp
```diff
@@ -12,7 +12,9 @@
 
 #include <QSet>
 
-static bool operator<(const EventModel::Process& process, qint32 pid)
+// silence clang warnings
```
milianw, line 15: what warning is this fixing? can you paste it please into the commit message

src/models/eventmodel.cpp
```diff
@@ -173,6 +178,8 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
                 return tr("Processes");
             case OverViewRow::Tracepoint_Row:
                 return tr("Tracepoints");
+            case 3:
```
milianw, line 181: OverViewRow::Favorite_Row

src/models/eventmodel.cpp
```diff
@@ -250,6 +259,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         Q_ASSERT(thread);
     } else if (tag == Tag::Tracepoints) {
         tracepoint = &m_data.tracepoints[index.row()];
+    } else if (tag == Tag::Favorites) {
+        auto& favourite = m_favourites[index.row()];
```
milianw, line 263: const, or just inline directly into the `data()` call below

src/models/eventmodel.cpp
```diff
@@ -263,6 +275,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Processes:
         case Tag::Tracepoints:
             return m_time.start;
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 279: they

src/models/eventmodel.cpp
```diff
@@ -275,6 +290,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Processes:
         case Tag::Tracepoints:
             return m_time.end;
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 294: dito

src/models/eventmodel.cpp
```diff
@@ -288,6 +306,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Processes:
         case Tag::Tracepoints:
             return {};
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 310: dito

src/models/eventmodel.cpp
```diff
@@ -300,6 +321,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Processes:
         case Tag::Tracepoints:
             return Data::INVALID_TID;
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 325: dito

src/models/eventmodel.cpp
```diff
@@ -312,6 +336,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Processes:
         case Tag::Tracepoints:
             return Data::INVALID_PID;
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 340: dito

src/models/eventmodel.cpp
```diff
@@ -324,6 +351,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Threads:
         case Tag::Tracepoints:
             return Data::INVALID_CPU_ID;
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 355: dito

src/models/eventmodel.cpp
```diff
@@ -338,6 +368,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
         case Tag::Overview:
         case Tag::Processes:
             return QVariant::fromValue(Data::Events());
+        case Tag::Favorites:
+            // there are handled elsewhere
```
milianw, line 372: dito

src/models/eventmodel.cpp
```diff
@@ -353,6 +386,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
             case Tag::Overview:
             case Tag::Processes:
                 return {};
+            case Tag::Favorites:
+                // there are handled elsewhere
```
milianw, line 390: dito

src/models/eventmodel.cpp
```diff
@@ -367,6 +403,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
...
```
milianw, line 407: dito

src/models/eventmodel.cpp
```diff
@@ -386,6 +425,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
...
```
milianw, line 429: dito

src/models/eventmodel.cpp
```diff
@@ -426,6 +468,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
             case Tag::Overview:
             case Tag::Processes:
                 break;
+            case Tag::Favorites:
+                // there are handled elsewhere
```
milianw, line 472: dito

src/models/eventmodel.cpp
```diff
@@ -447,6 +492,9 @@ QVariant EventModel::data(const QModelIndex& index, int role) const
             case Tag::Overview:
             case Tag::Processes:
                 return {};
+            case Tag::Favorites:
+                // there are handled elsewhere
```
milianw, line 496: dito

src/models/eventmodel.cpp
```diff
@@ -554,11 +605,20 @@ QModelIndex EventModel::parent(const QModelIndex& child) const
...
-    }
 
     return {};
 }
+
+void EventModel::addToFavourites(const QModelIndex& index)
+{
+    const int position = m_favourites.size();
+    beginInsertRows(createIndex(Favorite_Row, 0, static_cast<quintptr>(Tag::Overview)), position, position);
+    m_favourites.push_back(index);
```
milianw, line 622: you must clear this list in `EventModel::setData`

src/models/eventmodel.h
```diff
@@ -71,9 +71,13 @@ class EventModel : public QAbstractItemModel
...
+
 private:
     Data::EventResults m_data;
     QVector<Process> m_processes;
+    QVector<QModelIndex> m_favourites;
```
milianw, line 80: I find it somewhat unfortunate to use model indices here. seemingly easy, but also deceptively wrong.

when we favorite the index `(0, 0, someParent)` and then later request the favorite for that index, but at a different column, we need to translate to keep the column request working

src/models/eventmodel.cpp
```diff
@@ -554,11 +605,20 @@ QModelIndex EventModel::parent(const QModelIndex& child) const
...
         return createIndex(parentRow, 0, static_cast<quintptr>(Tag::Processes));
     }
-    }
 
     return {};
 }
+
+void EventModel::addToFavourites(const QModelIndex& index)
+{
+    const int position = m_favourites.size();
```
milianw, line 620: assert that the index.model() == this _and_ that `.column() == 0`

return early if the index is in the list of favorites already

src/models/eventmodel.cpp
milianw: this will be wrong when we get a request for column 1 but the favorite is for column  0 or vice versa

tests/modeltests/tst_models.cpp
```diff
@@ -537,7 +537,7 @@ private slots:
         model.setData(events);
 
         QCOMPARE(model.columnCount(), static_cast<int>(EventModel::NUM_COLUMNS));
-        QCOMPARE(model.rowCount(), 2);
+        QCOMPARE(model.rowCount(), 4);
```
milianw, line 540: what are the new rows? 

please also add test coverage for the new favorite feature. we also need to be able to un-favorite a row

milianw (changes requested):

tests/modeltests/tst_models.cpp
lievenhey: Favourites and tracepoints

milianw: but without a favorite, there should be no additional (empty) row. similarly, without any tracepoints, there shouldn't be an empty row for that either

src/models/eventmodel.cpp
milianw: this was not done

src/models/timelinedelegate.cpp
```diff
@@ -135,7 +135,8 @@ Data::Events::const_iterator findEvent(const Data::Events::const_iterator& begin
     auto it = std::lower_bound(begin, end, time, byTime);
     // it points to the first item for which our predicate returns false, we want to find the item before that
     // so decrement it if possible or return begin otherwise
-    return (it == begin || it->time == time) ? it : (it - 1);
+    // if only one event is recorded, it will point to end it->time which will cause asan to complain
```
milianw, line 139: this should probably be its own commit, it's unrelated to the other changes in this patch

milianw (changes requested):
there are still no unit tests for any of the new functionality - we need both! i.e. tests for the new behavior of showing tracepoint events in the model, and tests for the favorite.

we also cannot yet undo a favorite action, i.e. remove a favorite

tests/modeltests/tst_models.cpp
lievenhey: I added a proxy to hide the empty rows in the timelinewidget

milianw (changes requested):
hey @lievenhey - I cleaned up the remaining nits and played around with this. It definitely goes into the right direction.

but there is more work remaining to make this actually useful:

the biggest issue is that all tracepoints for all threads and processes are lumped together into a single row - that is pretty useless.

the way I see it, our "Event Sources" should become a combo box that allows multi-selection. Then, each of these event sources will get one row _below_ the CPU or Thread group.

The proxy should then flatten this, if just a single cost is selected - i.e. such that we keep the view as-is for data files that don't contain any extra tracepoint data.

Right now with your proposed patch here, one can have "Event Source: <my tracepoint>" and then we'd see the per-thread tracepoints and then one more row with all tracepoints clumped together :shrug: that's not useful

additionally, the tooltips shown in the event delegate are wrong when hovering a tracepoint event, because it only looks at the events of the selected source type, meaning tracepoints are never considered.

once those things are fixed, I am OK with merging this work as a stepping stone. the following steps will then be:
- time measurement between enter/exit similar to off-CPU time
- somehow making sure that the favorite rows are always shown, independent of the vertical current scroll position - potentially through a second view?

lievenhey:
> somehow making sure that the favorite rows are always shown, independent of the vertical current scroll position - potentially through a second view?

using a second view is most likely the easiest option since `QWidgets` don't have stuff like `qml`s `header` or `footer`

milianw:
> > somehow making sure that the favorite rows are always shown, independent of the vertical current scroll position - potentially through a second view?
> 
> using a second view is most likely the easiest option since `QWidgets` don't have stuff like `qml`s `header` or `footer`

Agreed, we just need to sync the views and show just a single header somehow. But this should/ could be done in a follow-up patch and doesn't necessarily block integration of the work here as long as the other important issues are resolved.

lievenhey:
@milianw I finally found some time to work in this. Can you check if it to your satisfaction?

milianw:
are you sure you pushed the correct work? I don't think you attended the issue I raised above, namely:

> the biggest issue is that all tracepoints for all threads and processes are lumped together into a single row - that is pretty useless.

I just checked out your branch, rebased it on master and ran it on data I obtained with

```
perf record -z --call-graph dwarf -e cycles -e raw_syscalls:sys_enter -e raw_syscalls:sys_exit ./tests/test-clients/cpp-parallel/cpp-parallel
```

The result is still a single list of tracepoints across all threads, which is not useful. What's worse, some threads are suddenly showing up multiple times in the timelines above?


Screenshot_20231210_124332

lievenhey:
Ok, fixed that. Now all events are shown in one line (before there was one line per even type).
That is still somewhat wip since now it is really hard to differentiate different event types since only the tooltip reveals the type. I propose the following:
- different colors for different event types
- when hovering highlight not only all stacks but also all symbols with the same type

lievenhey:
Currently it looks like this:
grafik

lievenhey:
waiting for https://github.com/KDAB/perfparser/pull/37

GitMensch:
That _is_ a different view to the hover provided by #684 (and both may be useful), right?

PR #509, opened 2023-09-21
add copy function to disassembly views

lievenhey:
This allows the user to copy text from the sourcecodeview and the disassembly view. This is done by reimplementing QTreeView with a custom keyPressEvent function that copies everything from selectionModel() to the clipboard and not just currentIndex like the default implementation does.

### DIFF
--- a/src/copyabletreeview.h
+++ b/src/copyabletreeview.h
@@ -1,0 +1,21 @@
+/*
+    SPDX-FileCopyrightText: Lieven Hey <lieven.hey@kdab.com>
+    SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include <QTreeView>
+
+class CopyableTreeView : public QTreeView
+{
+    Q_OBJECT
+public:
+    explicit CopyableTreeView(QWidget* parent = nullptr);
+    ~CopyableTreeView() override;
+
+protected:
+    void keyPressEvent(QKeyEvent* event) override;
+};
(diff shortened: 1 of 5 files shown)


lievenhey:
This makes hotspot behave like the user expects. I would be confused if I select multiple cells and only get the last one into my clipboard.

src/copyabletreeview.h
```diff
@@ -0,0 +1,19 @@
...
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include <QTreeView>
+
+class CopyableTreeView : public QTreeView
+{
+    Q_OBJECT public : CopyableTreeView(QWidget* parent = nullptr);
```
milianw, line 14: bad formatting, missing `explicit`

src/models/codedelegate.cpp
```diff
@@ -12,8 +12,7 @@
 #include <QPainter>
 #include <QTextLine>
 
-#include "disassemblymodel.h"
-#include "sourcecodemodel.h"
+Q_DECLARE_METATYPE(QTextLine)
```
milianw, line 15: how is this related to the copy operation? separate patch?

milianw (changes requested):

src/copyabletreeview.h
lievenhey: oh wow, apparently this is valid clang formatting

src/models/codedelegate.cpp
milianw: can you please answer the question?

milianw (changes requested):

src/models/codedelegate.cpp
lievenhey: clangd was complaining the whole time about unused includes so I removed those

Issue #501, opened 2023-07-29
Main window cannot be resized to fit in a 1080x1920px screen


leandrosansilva:
**Describe the bug**

I use a HD (1920x1080px) screen "on the vertical", making it 1080x1920px, and the Hotspot window refuses to fit on it, as it refuses to be narrower than 1100px.

**To Reproduce**

Try to resize the window. It won't get narrower than 1100px. When maximizing the window, parts of it will be "off screen".

**Expected behavior**

**Screenshots**

![image](https://github.com/KDAB/hotspot/assets/1503334/6042c947-f934-499d-9899-f5361590536b)

Maximized window (note that part of the right area is missing, out of the screen):

![image](https://github.com/KDAB/hotspot/assets/1503334/c54f5d31-ceed-460c-a8b1-4a7a43a4fc3a)

**Version Info (please complete the following information):**

- Linux Kernel version:
- perf version: 6.3
- hotspot version (appimage? selfcompiled?): 1.4.1 from Manjaro/AUR
- if self-compiled hotspot, what version of elfutils:

**Additional context**
I am using Kwin 5.27 with Wayland. 1x scale.

PR #512, opened 2023-09-26
shove flamegraph controls into QToolBar

lievenhey:
This change removes the minimum width requirements of about 1100pixels, since Qt can hide widgets of a QToolBar in a menu.

fixes: #501


src/flamegraph.cpp
```diff
@@ -653,65 +676,144 @@ FlameGraph::FlameGraph(QWidget* parent, Qt::WindowFlags flags)
...
+            connect(searchInput, &QLineEdit::textChanged, this, &FlameGraph::setSearchValue);
+            connect(this, &FlameGraph::uiResetRequested, this, [this, searchInput] {
+                m_search.clear();
+                searchInput->clear();
+            });
+        },
+        this);
+
+    // use a QToolBar to automatically hide widgets in a menu that don't fit into the window
+    // the first 3 can stay as a widget since in their case the with is dominated by the number of tabs
```
milianw, line 793: typo: with -> width

milianw: also: why do you say "the first 3" - it looks like at least four widgets and three actions are added below - what does "the first 3" refer to?

milianw (changes requested):
minor nit, otherwise lgtm - thanks

milianw (approved):

PR #517, opened 2023-10-09
Update perfparser and use configure_file to generate config-perfparser.h

dfaure-kdab:
This also fixes the build error if libdebuginfod is installed but not libdebuginfo-devel, i.e. if debuginfod.h is missing.

Requires https://github.com/KDAB/perfparser/pull/29 to be merged first.


milianw:
thanks!

PR #518, opened 2023-10-10
Disable branches

lievenhey:


src/models/disassemblyoutput.cpp
```diff
@@ -271,10 +271,12 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
                                   toHex(symbol.relAddr + symbol.size)};
 
     // only available for objdump 2.34+
-    if (canVisualizeJumps(processPath))
+    auto canShowBranches = canVisualizeJumps(processPath);
```
GitMensch, line 274: This may be a good time to recheck why this function is called twice - when you use the disassembly function and it the jumps are not supported, then the message about missing support is seen twice.

GitMensch:
LGTM in general.

What do you think about the idea to have the jumps in a separate "column"?
In this case the column would be auto-hidden/disabled if `CanVisualize() == false` (or the visualization was turned of in the settings) and otherwise the user would be able to show/hide that column at any time.

If you don't want to follow that route, then I think it would be useful to (additionally?) have a toggle to be able to switch this directly on the disassembly page to re-read with visualization and update.

src/models/disassemblyoutput.cpp
lievenhey: somehow `disassemble` is called twice but I haven't investigated further

lievenhey:
> What do you think about the idea to have the jumps in a separate "column"?

The more I think about it, the better it sounds. In the current version I completely reload the model which will reset the view if I would add a checkbox to toggle this feature.  I originally didn't want to go that route since it involves a lot of dark magic (regexes) but it seems to be the better solution in the long run.

src/models/disassemblyoutput.cpp
GitMensch: I _guess_ fixing that will also speed up handling of huge functions. Thanks for having a look at this later.

GitMensch:
Just to recheck as it wasn't obvious from inspecting only the changes: is the new branches column "monospace text" only, or is it "syntax-highlighted", too? I _think_ it may also work with non-monospace (then looking better and saving space - but I'm not sure...)

lievenhey:
> The current resulting AppImage is much faster, instead of 24 seconds it only needs 8-9 seconds for opening.
If I globally disable the jumps in the settings, then it gets down to 5-6 seconds.

Interesting seem like qt has a problem with layouting if there are a lots of whitespaces

> Disabling the branch column "on demand" works fine. Note: the visible columns are not saved, if you disable it, then go back to the callee view, then use Disassemble on the same function the column is enabled again; same if the setting is of and the column is manually enabled.

I see this as an on demand feature for quick stuff. It is only saved, if you change it in the settings.

> Just to recheck as it wasn't obvious from inspecting only the changes: is the new branches column "monospace text" only, or is it "syntax-highlighted", too? I think it may also work with non-monospace (then looking better and saving space - but I'm not sure...)

it is monospace only

GitMensch:
> > Just to recheck as it wasn't obvious from inspecting only the changes: is the new branches column "monospace text" only, or is it "syntax-highlighted", too? I think it may also work with non-monospace (then looking better and saving space - but I'm not sure...)
> 
> it is monospace only

Can you give it a short try to see how it looks with the default font? If that looks better (or at least not worse and saves space), then this would be possibly a reasonable switch.

lievenhey:
> The current resulting AppImage is much faster, instead of 24 seconds it only needs 8-9 seconds for opening.
If I globally disable the jumps in the settings, then it gets down to 5-6 seconds.

That is way too long. It should only take a few milliseconds. I will ask some KDE People about this, since they wrote their own Layout Engine which they use in Kate. Kate can layout 2 million lines in like 10ms.

lievenhey:
> Can you give it a short try to see how it looks with the default font? If that looks better (or at least not worse and saves space), then this would be possibly a reasonable switch.

looks terrible since a space has a different width than a dash
grafik

src/models/disassemblyoutput.cpp
```diff
@@ -128,6 +129,7 @@ ObjectdumpOutput objdumpParse(const QByteArray& output)
     QString asmLine;
     QString sourceFileName;
     QString currentSourceFileName;
+    const QRegularExpression branchRegex(QStringLiteral("^([ |\\/\\\\>-]+)"));
```
milianw, line 132:
```
static const auto branchRegEx = QRegularExpression(QStringLiteral("..."));
```

that said, it would be trivial to rewrite this to not use a regex and just do a simple skip from front to split the line from its branches. that's going to be the best

src/resultsdisassemblypage.cpp
```diff
@@ -257,6 +252,10 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
                          ui->disasmSearchWidget, ui->disasmSearchEdit, ui->assemblyView, ui->disasmEndReachedWidget,
                          m_disassemblyModel, &m_currentDisasmSearchIndex, 0);
 
+    connect(settings, &Settings::showBranchesChanged, this, [this](bool showBranches) {
+        ui->assemblyView->setColumnHidden(DisassemblyModel::BranchColumn, !showBranches);
+    });
```
milianw, line 257: missing call to apply the lambda directly from the current settings value

src/resultsdisassemblypage.cpp
```diff
@@ -410,6 +411,9 @@ void ResultsDisassemblyPage::showDisassembly(const DisassemblyOutput& disassembl
     ResultsUtil::hideTracepointColumns(m_callerCalleeResults.selfCosts, ui->sourceCodeView,
                                        SourceCodeModel::COLUMN_COUNT);
 
+    auto settings = Settings::instance();
+    ui->assemblyView->setColumnHidden(DisassemblyModel::BranchColumn, !settings->showBranches());
```
milianw, line 415: ah here it is, DRY

milianw (changes requested):

src/models/disassemblyoutput.cpp
lievenhey: right dabbling with the dark arts was completely unnecessary

src/models/disassemblyoutput.cpp
```diff
@@ -199,8 +200,24 @@ ObjectdumpOutput objdumpParse(const QByteArray& output)
...
-        disassemblyLines.push_back(
-            {addr, assembly, extractLinkedFunction(asmLine), {currentSourceFileName, sourceCodeLine}});
+        int branchVisualisationRange = 0;
+        for (const auto& character : std::as_const(assembly)) {
+            // format is the following:
+            //    /-  a5 54 12 ...
+            //    |   64 a3 ....
+            //    \-> 65 23 ....
+            // so we can simply skip all characters until we meet a letter or a number
+            if (character.isNumber() || character.isLetter()) {
```
GitMensch, line 210: I think this should be either `if (character.isNumber() || character.isLower()) {` or, possibly change to a lookup string as this won't use the locale under the hood and the Qt doc for `isNumber()` explicit warns that it returns true for
>[Number_](https://doc.qt.io/qt-5/stylesheet-reference.html#number)* categories, not just 0-9

That would be possibly something like

```c
static QString const hex("0123456789abcdef");

        for (const auto& character : std::as_const(assembly)) {
            // format is the following:
            //    /-  a5 54 12 ...
            //    |   64 a3 ....
            //    \-> 65 23 ....
            // so we can simply skip all characters until we meet a letter or a number
            if (hex.indexOf(character) != -1) {
                break;
            }
```

src/models/disassemblyoutput.cpp
```diff
@@ -15,6 +15,7 @@
 #include <QFileInfo>
 #include <QLoggingCategory>
 #include <QProcess>
+#include <QRegularExpression>
```
milianw, line 18: not needed anymore

milianw (changes requested):

src/models/disassemblyoutput.cpp
```diff
@@ -120,6 +120,11 @@ QString findBinaryForSymbol(const QStringList& debugPaths, const QStringList& ex
     return {};
 }
 
+bool isHexCharacter(const QChar c)
+{
+    return (c >= QLatin1Char('0') && c <= QLatin1Char('9')) || (c >= QLatin1Char('a') && c <= QLatin1Char('f'));
```
GitMensch, line 125: That's nice, especially the calling!
I just wasn't sure if there will be a 1 to 4 QLatin1Char objects created when entering this function which may be prevented when declaring them as static const in the function.

GitMensch (commented):
With the new logic of keeping the branches in a separate column, I think we could simplify the split.

Reasoning (please recheck if that's correct):
* we have one big disassembly from objdump
* at least currently the start of the assembly is always at the same line position (that would change if we have colored input because of terminal sequences)

In this case we can check the start of the assembly exactly one time on the first line of the assembly, then use that offset for each record.

That would drop the distance-call from each line, calculating it only once. preferably outside.

src/models/disassemblyoutput.cpp
lievenhey: no need to, the compiler can optimize it away [see](https://godbolt.org/z/9qxd1c1eM)

GitMensch:
> I will have a look at coloured output soonish, so I would keep it as is. Especially since it is a cheap calculation.

Note that escape sequences would again match the hex-check :-)
But I totally agree that this can be revisited later.

lievenhey:
> This is now much better than before, I hope this gets pulled in and afterwards we may get a finished https://github.com/KDAB/hotspot/pull/494 (looks like 98% done) pulled in, too.

This will still take some time. We want to move the logic to the `perfparser` since it has all the logic for finding binaries etc.

milianw (approved):
I'm fine with this as an incremental step, but I dislike that toggling the visibility of the branch column directly is detached from the settings - can you please synchronize that to make it more obvious?

furthermore, I'm surprised to see that this patch has no test changes. That means this feature is currently completely untested? Can you please work on adding test coverage @lievenhey ?

If you prefer, you can merge this right away and work on both of the above in separate merge requests.

Thanks!

lievenhey:
I will add a test and then merge this

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+
+        bool ok = false;
+        const quint64 address = match.captured(1).toInt(&ok, 16);
+        Q_ASSERT(ok);
+        const quint64 size = match.captured(2).toInt(&ok, 10);
+        Q_ASSERT(ok);
+        return {address, size};
+    }
+
+    bool supportsVisualizeJumps()
```
milianw, line 248: const or static

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+        for (const auto& line : result.disassemblyLines) {
+            QVERIFY(!line.branchVisualisation.isEmpty());
+            if (!line.disassembly.isEmpty()) {
+                QVERIFY(isHex(line.disassembly[0]));
+            }
+        }
+    }
+
+private:
+    std::tuple<quint64, quint64> findAddressAndSizeOfFunc(const QString& library, const QString& name)
```
milianw, line 223: return a struct with named arguments, tuple is a code smell in non-generic code

make the method const or static

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+
+        QProcess readelf;
+        readelf.setProgram(QStringLiteral("readelf"));
+        readelf.setArguments({QStringLiteral("-s"), library});
+
+        readelf.start();
+        readelf.waitForFinished();
+
+        const auto output = readelf.readAllStandardOutput();
+        Q_ASSERT(!output.isEmpty());
```
milianw, line 235: this won't do anything in a release build, you need to use VERIFY_OR_THROW and friends instead

never use Q_ASSERT in tests

repeats a couple of times below

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+    std::tuple<quint64, quint64> findAddressAndSizeOfFunc(const QString& library, const QString& name)
+    {
+        QRegularExpression regex(QStringLiteral("[ ]+[0-9]+: ([0-9a-f]+)[ ]+([0-9]+)[0-9 a-zA-Z]+%1\\n").arg(name));
+
+        QProcess readelf;
+        readelf.setProgram(QStringLiteral("readelf"));
+        readelf.setArguments({QStringLiteral("-s"), library});
+
+        readelf.start();
+        readelf.waitForFinished();
```
milianw, line 232: VERIFY_OR_THROW(readelf.watiForFinished());

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+        }
+    }
+
+private:
+    std::tuple<quint64, quint64> findAddressAndSizeOfFunc(const QString& library, const QString& name)
+    {
+        QRegularExpression regex(QStringLiteral("[ ]+[0-9]+: ([0-9a-f]+)[ ]+([0-9]+)[0-9 a-zA-Z]+%1\\n").arg(name));
+
+        QProcess readelf;
+        readelf.setProgram(QStringLiteral("readelf"));
```
milianw, line 228: verify that this is actually available and fail the test otherwise, e.g.:

```
auto readElfBinary = QStandardPaths::findExecutable(QStringLiteral("readelf"));
VERIFY_OR_THROW(!readElfBinary.isEmpty());
```

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+    void testDetectBranches()
+    {
+        if (!supportsVisualizeJumps()) {
+            QSKIP("--visualize-jumps is not supported");
+        }
+
+        const auto lib = QFileInfo(findLib(QStringLiteral("libfib.so")));
+        auto [address, size] = findAddressAndSizeOfFunc(lib.absoluteFilePath(), QStringLiteral("_Z3fibi"));
+
+        const Data::Symbol symbol = {QStringLiteral("fib(int)"), address, size, QStringLiteral("libfib.so")};
```
milianw, line 203: const auto symbol = Data::Symbol{...};

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+
+        const QString objdump = QStandardPaths::findExecutable(QStringLiteral("objdump"));
```
milianw, line 205: const auto
QVERIFY(!objdump.isEmpty());

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+        QVERIFY(result.errorMessage.isEmpty());
+
+        auto isHex = [](const QChar c) {
+            return (c >= QLatin1Char('0') && c < QLatin1Char('9')) || (c >= QLatin1Char('a') && c <= QLatin1Char('f'));
+        };
+
+        for (const auto& line : result.disassemblyLines) {
+            QVERIFY(!line.branchVisualisation.isEmpty());
+            if (!line.disassembly.isEmpty()) {
+                QVERIFY(isHex(line.disassembly[0]));
```
milianw, line 217:
```
bool ok = false;
auto addr = line.disassembly[0].toULongLong(&ok, 16);
QVERIFY(ok);
QVERIFY(addr > 0);
```

and then remove `isHex`

milianw: can we also test then that the `addr` is in the range we got from `findAddressAndSizeOfFunc`?

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+        readelf.waitForFinished();
+
+        const auto output = readelf.readAllStandardOutput();
+        Q_ASSERT(!output.isEmpty());
+
+        auto match = regex.match(QString::fromUtf8(output));
+        Q_ASSERT(match.hasMatch());
+
+        bool ok = false;
+        const quint64 address = match.captured(1).toInt(&ok, 16);
```
milianw, line 241:
```
const auto address = ...toULongLong()
```

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -189,6 +190,80 @@ private slots:
...
+        Q_ASSERT(ok);
+        const quint64 size = match.captured(2).toInt(&ok, 10);
```
milianw, line 243: also toULongLong

milianw (changes requested):

tests/modeltests/tst_disassemblyoutput.cpp
lievenhey: its not the address, this is the hexdump, but yes will do

tests/modeltests/tst_disassemblyoutput.cpp
milianw: use readelfBinary from above

tests/modeltests/tst_disassemblyoutput.cpp
milianw: auto missing

milianw (changes requested):
I'll fix the small nits myself now

milianw (approved):

PR #525, opened 2023-10-13
update to appimage docker image to use devtoolset-12

GitMensch:
In the past I've sometimes got some elfutils related problems when using the AppImage; they _possibly_ are fixed with the newer elfutils (0.187, currently used devtoolset-11 has 0.185) and also the other parts (like GCC for building from source) are likely not bad either.

Note: as I actually don't "do docker" and am a merely user of the appImage I cannot test this other than trying the resulting CI build.

### DIFF
--- a/scripts/appimage/Dockerfile
+++ b/scripts/appimage/Dockerfile
@@ -65,7 +65,7 @@ RUN wget https://download.qt.io/official_releases/qt/5.15/${QT_VERSION}/single/q
     make -j && make install && cd .. && rm -Rf build qt-everywhere-*
 
 # appimage build tools
-RUN wget https://github.com/linuxdeploy/linuxdeploy/releases/download/1-alpha-20220822-1/linuxdeploy-x86_64.AppImage &&     chmod +x linuxdeploy-x86_64.AppImage && \
+RUN wget https://github.com/linuxdeploy/linuxdeploy/releases/download/1-alpha-20230713-1/linuxdeploy-x86_64.AppImage &&     chmod +x linuxdeploy-x86_64.AppImage && \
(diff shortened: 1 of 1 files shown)


milianw:
the docker images are not rebuild automatically, I have to do that manually. 

If you know how to do that automatically in a way that we can still test things locally then I'm all ears.

For now, I'll test these changes manually sometime this week and apply them then

Thanks!

GitMensch:
>If you know how to do that automatically in a way that we can still test things locally then I'm all ears.

me and docker: 🤷 

> For now, I'll test these changes manually sometime this week and apply them then

🚀 Sounds good!

> Thanks!

You're welcome, thanks for your work in Hotspot (I obviously like it)!

milianw:
it doesn't look like centos7 has devtoolset-12:

```
No package devtoolset-12 available.
No package devtoolset-12-elfutils-devel available.
No package devtoolset-12-elfutils-debuginfod available.
```

I intend to leverage KDE craft going forward, and then we'd have more up2date libraries as needed. for now I'll close this request (but I  include the second patch that updated linuxdeploy, thanks!)

GitMensch:
I see. Just for my understanding: The idea is to use CentOS7 to ensure that the AppImage uses a glibc version that is widely available, right?

Using [KDE Craft](https://github.com/KDE/craft) would mean not (primarily) using `apt` (and potentially a bunch of ppa's) but that packaging system to update the AppImage docker to have recent tools and libraries available, correct?

milianw:
Yes re Centos - it gives maximum compatibility basically when running the AppImage.

Craft compiles basically the world from scratch, and there's a binary cache that's filled from the KDE CI. We lose some compat but gain on the maintainability front and ease-of-use.

Issue #526, opened 2023-10-13
model search test: out of bounds compile warning


GitMensch:
**Describe the bug**
Just recognized that in the CI build log:

```
In file included from /__w/hotspot/hotspot/tests/modeltests/tst_search.cpp:15:
/__w/hotspot/hotspot/tests/modeltests/../../src/models/search.h: In function 'static void TestSearch::qt_static_metacall(QObject*, QMetaObject::Call, int, void**)':
/__w/hotspot/hotspot/tests/modeltests/../../src/models/search.h:32:10: warning: array subscript 6 is outside array bounds of 'const std::array<int, 5> [1]' [-Warray-bounds]
   32 |     auto start = current + 1;
      |          ^~~~~
/__w/hotspot/hotspot/tests/modeltests/tst_search.cpp:73:34: note: while referencing 'testArray'
   73 |         const std::array<int, 5> testArray = {1, 2, 3, 4, 5};
      |                                  ^~~~~~~~~
In file included from /__w/hotspot/hotspot/tests/modeltests/tst_search.cpp:15:
/__w/hotspot/hotspot/tests/modeltests/../../src/models/search.h:32:10: warning: array subscript 6 is outside array bounds of 'const std::array<int, 5> [1]' [-Warray-bounds]
   32 |     auto start = current + 1;
      |          ^~~~~
/__w/hotspot/hotspot/tests/modeltests/tst_search.cpp:73:34: note: while referencing 'testArray'
   73 |         const std::array<int, 5> testArray = {1, 2, 3, 4, 5};
      |                                  ^~~~~~~~~
```

**To Reproduce**
Check the build log - or compile with GCC 1.21 locally.

PR #530, opened 2023-10-17
fix out of range access

lievenhey:
start was set to current + 1 which will cause undefined behavior if current is the last element
this patch adds a check for this and wraps around in that case

fixes: #526

### DIFF
--- a/src/models/search.h
+++ b/src/models/search.h
@@ -29,7 +29,7 @@ int search_impl(iter begin, iter end, iter current, SearchFunc searchFunc, EndRe
     if (begin == end)
         return -1;
 
-    auto start = current + 1;
+    const auto start = (current == end) ? begin : (current + 1);
     auto found = std::find_if(start, end, searchFunc);
 
     if (found != end) {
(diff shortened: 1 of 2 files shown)


lievenhey:
I have to check the output of the appimage build since the warning is only shown there

milianw (changes requested):
add a test for this case please

lievenhey:
How do I test this? Since gcc 5 `std::find_if` returns `end` if `begin` > `end` (figured out by playing with godbolt, since this error only happens on old compilers)

src/models/search.h
```diff
@@ -29,7 +29,13 @@ int search_impl(iter begin, iter end, iter current, SearchFunc searchFunc, EndRe
     if (begin == end)
         return -1;
 
-    auto start = current + 1;
+    iter start;
```
milianw, line 32:
```
auto start = (current == end) ? begin : (current + 1);
```

milianw (changes requested):

Issue #527, opened 2023-10-13
Disassembly view: allow to disable hex by moving it to a separate column


GitMensch:
Follow-up to #516, to allow hiding it like KCachegrind
![kcachegrind disassembly](https://github.com/KDAB/hotspot/assets/6699539/5889b83f-caa1-4e98-b0c7-a9ee0b3e51e6)
as _originally requested by @milianw in https://github.com/KDAB/hotspot/issues/523#issuecomment-1760994607_

> @lievenhey if possible, could you also have a look at how kcachegrind fills its disassembly view? [...] I also like that it splits up more of the content into columns one could individually hide (e.g. the hex stuff wouldn't be too interesting for me).

milianw:
I pushed some improvements for the branch visualization now, which allows a similar compact visualization to what is shown in the screenshot above. more fine grained splitting of the disassembly column would still be appreciated

milianw:
can you please upload the raw output from `objdump` for the above command here? e.g. use `strace` to see how we run `objdump` internally, then do it manually from the CLI and pipe the output to a file, then upload here. then I can fix it - previously it was broken for objdump on my end... I guess the output simply isn't too stable and we need to support multiple formats somehow

GitMensch:
I've tried, but there's no useful output in strace when running from the appinfo... while I extract and rerun that, I've created https://github.com/KDAB/hotspot/pull/535 which will help to get there faster in the next time and may also provide useful in general

GitMensch:
Here's a bit of output using the same parameters `objdump -d -l -C --visualize-jumps` from this machine:

```
objdump -d -l -C --start-address 0x17fe --stop-address  0x1884 --visualize-jumps $(which cobcrun)

/usr/bin/cobcrun:     file format elf64-x86-64


Disassembly of section .text:

00000000000017fe <_start@@Base+0x39e>:
    17fe:              48 8d 3d d3 0e 00 00     lea    0xed3(%rip),%rdi        # 26d8 <_IO_stdin_used@@Base+0x6d8>
    1805:              e9 46 f8 ff ff           jmpq   1050 <puts@plt>
    180a:              66 0f 1f 44 00 00        nopw   0x0(%rax,%rax,1)
    1810:              48 85 ff                 test   %rdi,%rdi
    1813:              0f 84 9f 00 00 00        je     18b8 <_start@@Base+0x458>
    1819:              41 56                    push   %r14
    181b:              41 55                    push   %r13
[...]

PR #535, opened 2023-10-22
adjustments to objdump call 

GitMensch:
After recognizing that the process path of objdump is resolved first, I've adjusted the PR to use that (so the additional `stat` calls seen in #527 are not needed).

* invoke directly from the processPath already resolved before
* improve error message for objdump invocation issues helping to find problems related to the call

### DIFF
--- a/src/models/disassemblyoutput.cpp
+++ b/src/models/disassemblyoutput.cpp
@@ -268,6 +268,10 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
         disassemblyOutput.errorMessage = QApplication::tr("Empty symbol ?? is selected");
         return disassemblyOutput;
     }
+    if (symbol.relAddr == 0 || symbol.size == 0) {
+        disassemblyOutput.errorMessage = QApplication::tr("Symbol with unknown details is selected");
+        return disassemblyOutput;
+    }
 
     const auto processPath = QStandardPaths::findExecutable(objdump);
     if (processPath.isEmpty()) {
--- a/src/models/disassemblyoutput.cpp
+++ b/src/models/disassemblyoutput.cpp
@@ -265,18 +265,26 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
     DisassemblyOutput disassemblyOutput;
     disassemblyOutput.symbol = symbol;
     if (symbol.symbol.isEmpty()) {
-        disassemblyOutput.errorMessage = QApplication::tr("Empty symbol ?? is selected");
+        disassemblyOutput.errorMessage =
+            QApplication::translate("DisassemblyOutput",
+                                    "<qt>Empty symbol <tt>"
+                                    "??" // note: using string concatenation to prevent -Wtrigraph
+                                    "</tt> is selected.");
         return disassemblyOutput;
     }
     if (symbol.relAddr == 0 || symbol.size == 0) {
-        disassemblyOutput.errorMessage = QApplication::tr("Symbol with unknown details is selected");
+        disassemblyOutput.errorMessage =
+            QApplication::translate("DisassemblyOutput", "<qt>Symbol <tt>%1</tt> with unknown details is selected.")
+                .arg(symbol.symbol);
         return disassemblyOutput;
     }
 
     const auto processPath = QStandardPaths::findExecutable(objdump);
     if (processPath.isEmpty()) {
         disassemblyOutput.errorMessage =
-            QApplication::tr("Cannot find objdump process %1, please install the missing binutils package for arch %2")
+            QApplication::translate("DisassemblyOutput",
+                                    "<qt>Cannot find objdump process <tt>%1</tt>, please install "
+                                    "the missing binutils package for arch <tt>%2</tt>.")
                 .arg(objdump, arch);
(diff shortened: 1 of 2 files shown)


lievenhey:
@GitMensch lgtm, but can you ran `clang-format` so that the ci no longer complains?

GitMensch:
applied `clang-format -i` and pushed

src/models/disassemblyoutput.cpp
```diff
@@ -308,22 +308,26 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
     }
     arguments.append(binary);
 
-    asmProcess.start(objdump, arguments);
+    asmProcess.start(processPath, arguments);
 
     if (!asmProcess.waitForStarted()) {
-        disassemblyOutput.errorMessage += QApplication::translate("DisassemblyOutput", "Process was not started.");
+        disassemblyOutput.errorMessage += QApplication::translate("DisassemblyOutput", "Process was not started.")
```
milianw, line 314: string puzzle, this is impossible to translate properly. Make this one large string and then use arg to fill in the placeholders and also include the `QProcess::errorString`

```
..., "<qt>Process failed to start: <tt>%1 %2</tt> returned <tt>%2</tt>").arg(processPath, arguments.join(...), asmProcess.errorString())
```

src/models/disassemblyoutput.cpp
```diff
@@ -308,22 +308,26 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
...
+            + QStringLiteral("\n")
+            + QApplication::tr("command: %1 %2").arg(processPath, arguments.join(QLatin1Char(' ')));
         return disassemblyOutput;
     }
 
     if (!asmProcess.waitForFinished()) {
         disassemblyOutput.errorMessage +=
-            QApplication::translate("DisassemblyOutput", "Process was not finished. Stopped by timeout");
+            QApplication::translate("DisassemblyOutput", "Process was not finished. Stopped by timeout")
```
milianw, line 322: dito

milianw (changes requested):

src/models/disassemblyoutput.cpp
GitMensch: will fix the string puzzles, I likely was concerned about not breaking the existing translations

GitMensch:
* string puzzle fixed, other invocations of error messages changed accordingly
* additional commit fixing a related issue (missing check -> error message)

src/models/disassemblyoutput.cpp
```diff
@@ -272,7 +272,9 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
     const auto processPath = QStandardPaths::findExecutable(objdump);
     if (processPath.isEmpty()) {
         disassemblyOutput.errorMessage =
-            QApplication::tr("Cannot find objdump process %1, please install the missing binutils package for arch %2")
+            QApplication::translate("DisassemblyOutput",
```
milianw, line 283: these hunks here are unrelated to what the commit message talks about - split that up into a separate commit

src/models/disassemblyoutput.cpp
```diff
@@ -303,27 +305,32 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
...
     if (binary.isEmpty()) {
         disassemblyOutput.errorMessage +=
-            QApplication::translate("DisassemblyOutput", "Could not find %1").arg(symbol.binary);
+            QApplication::translate("DisassemblyOutput", "<qt>Could not find <tt>%1</tt>").arg(symbol.binary);
         return disassemblyOutput;
     }
     arguments.append(binary);
 
-    asmProcess.start(objdump, arguments);
+    asmProcess.start(processPath, arguments);
```
milianw, line 321: ideally this here is also its own patch - whenever your commit message is something like "do A and B" then it's an indication that you should have split it up into two patches instead

src/models/disassemblyoutput.cpp
```diff
@@ -265,7 +265,15 @@ DisassemblyOutput DisassemblyOutput::disassemble(const QString& objdump, const Q
     DisassemblyOutput disassemblyOutput;
     disassemblyOutput.symbol = symbol;
     if (symbol.symbol.isEmpty()) {
-        disassemblyOutput.errorMessage = QApplication::tr("Empty symbol ?? is selected");
+        disassemblyOutput.errorMessage =
```
milianw, line 268: this doesn't really help - if you really want to wrap the `??` then just do that directly, no need for `.arg()`

milianw (changes requested):

src/models/disassemblyoutput.cpp
GitMensch: That's what I did, just to get a compile warning about trigraphs., which was the reason to split that. Another option is to use inline hex values, bit that is likely not useful for the translation.

GitMensch:
rebased and split from 2 to 3 separate commits (I do like the result more than the original, too); tested manually in the GUI using the files under "tests" (all but the timeout, checked this with the appImage and also works as expected).

Adjusted one test as needed and added some of the missing tests for the adjusted messages.

milianw:
thanks!

PR #538, opened 2023-10-23
Better asm syntax highlighting

lievenhey:


lievenhey:
left old, right new
grafik

lievenhey:
TODO:
- test
- hide format chooser if ansi codes are used
- add option to hide hexdump

GitMensch:
I'm a bit confused, why does "Syntax Highlighting: None" have any color? But nice to see that those issues are tackled!

lievenhey:
> I'm a bit confused, why does "Syntax Highlighting: None" have any color? But nice to see that those issues are tackled!

someone didn't implement synchronisation yet

GitMensch:
As this auto-drops the highlighting on the hex values and also changes the code involved here, I highly suggest to implement #527 (moving the hex parts out as requested by Milian) first. Could be done in this PR, too.

src/models/disassemblyoutput.cpp
```diff
@@ -207,29 +207,25 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
...
-            const auto addrEnd = lineRest.indexOf(suffix, prefix.size());
-            if (addrEnd == -1)
+        const auto addr = [addrString = parts[0].trimmed(), &asmLine]() -> uint64_t {
+            const auto suffix = QLatin1Char(':');
+            if (!addrString.endsWith(suffix))
                 return 0;
 
             bool ok = false;
-            auto ret = lineRest.mid(0, addrEnd).toULongLong(&ok, 16);
+            auto ret = addrString.midRef(0, addrString.size() - 1).toULongLong(&ok, 16);
```
milianw, line 249: the midRef isn't needed anymore now, it should be the same as just `addrString.toULongLong(...)`

src/models/disassemblyoutput.cpp
```diff
@@ -207,29 +207,25 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
...
-                return 0;
+        // a line looks like this:
+        // [spaces]addr:\t [branch visualisazion] [hexdump]\tdiassembly
+        // we can simplify parsing by splitting it into three parts
+
+        const auto parts = asmLine.split(QLatin1Char('\t'));
 
-            const auto addrEnd = lineRest.indexOf(suffix, prefix.size());
-            if (addrEnd == -1)
+        const auto addr = [addrString = parts[0].trimmed(), &asmLine]() -> uint64_t {
```
milianw, line 243: `parts.value(0)` for extra safety

src/models/disassemblyoutput.cpp
```diff
@@ -238,19 +234,23 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
...
-            auto size = std::distance(lineRest.cbegin(), firstHexIt);
-            auto ret = lineRest.mid(0, size);
-            lineRest = lineRest.mid(size);
-            return ret;
+        struct BranchesAndHexdump
+        {
+            QString branchVisualisation;
+            QString hexdump;
+        };
+        const auto [branchVisualisation, hexdump] = [branchesAndHex = parts[1]]() -> BranchesAndHexdump {
```
milianw, line 269: `parts.value(1)` for extra safety

src/models/disassemblyoutput.cpp
```diff
@@ -238,19 +234,23 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
...
+            auto size = std::distance(branchesAndHex.cbegin(), firstHexIt);
+            auto branchVisualisation = branchesAndHex.mid(0, size);
+            auto hexdump = branchesAndHex.mid(size);
+            return {branchVisualisation, hexdump};
         }();
 
         disassemblyLines.push_back({addr,
-                                    lineRest.toString(),
-                                    branchVisualisation.toString(),
+                                    parts.size() == 3 ? parts[2].trimmed() : QString {},
```
milianw, line 278: simplify to `parts.value(2).trimmed()`

src/models/disassemblymodel.cpp
```diff
@@ -82,6 +82,8 @@ QVariant DisassemblyModel::headerData(int section, Qt::Orientation orientation,
         return tr("Address");
     else if (section == BranchColumn)
         return tr("Branches");
+    else if (section == HexdumpColumn)
```
milianw, line 74: this patch is missing test coverage

src/models/disassemblyoutput.cpp
```diff
@@ -17,28 +17,46 @@
 #include <QProcess>
 #include <QStandardPaths>
 
+#include "formattingutils.h"
+
 namespace {
 Q_LOGGING_CATEGORY(disassemblyoutput, "hotspot.disassemblyoutput")
 
-bool canVisualizeJumps(const QString& objdump)
+QString objdumpHelp(const QString& objdump)
```
milianw, line 25: just return a `QByteArray` and use that then with `QByteArrayLiteral` - then we don't need the extra `QString` conversion

src/models/disassemblyoutput.cpp
```diff
@@ -17,28 +17,46 @@
...
+{
+    return objdumpHelp(objdump).contains(QStringLiteral("--disassembler-color"));
 }
 
-DisassemblyOutput::LinkedFunction extractLinkedFunction(const QString& disassembly)
+DisassemblyOutput::LinkedFunction extractLinkedFunction(const QString& disassemblyWithAnsi)
 {
     DisassemblyOutput::LinkedFunction function = {};
 
+    const auto escapeChar = QLatin1Char('\u001B');
```
milianw, line 52: add a comment - what is this char?

src/models/disassemblyoutput.cpp
```diff
@@ -213,6 +232,14 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
 
         const auto parts = asmLine.split(QLatin1Char('\t'));
 
+        if (parts.size() == 1 && asmLine.endsWith(QLatin1Char(':'))) {
```
milianw, line 235: it's unfortunate that you do so much in this patch - I would really appreciate if you could split it up - one patch that just ports away from the old QTextDocument usage to the new highlighter, then another patch for adding the custom highlighting

src/models/formattingutils.h
```diff
@@ -0,0 +1,14 @@
...
+    SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include <QString>
+
+namespace utils {
```
milianw, line 12: afair we use `Util` in all other areas already, use the same casing here too

src/models/formattingutils.h
```diff
@@ -0,0 +1,14 @@
+/*
+    SPDX-FileCopyrightText: Lieven Hey <lieven.hey@kdab.com>
+    SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include <QString>
```
milianw, line 10: forward declare

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,281 @@
...
+    {
+        m_formats.clear();
+        m_offset = 0;
+
+        KSyntaxHighlighting::State state;
+        for (const auto& line : text) {
+            state = highlightLine(line, state);
+
+            // KSyntaxHighlighting uses line offsets but QTextLayout uses global offsets
+            // the +1 is for the newline at the end of the line
```
milianw, line 49: there is no +1 here

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,281 @@
...
+    m_layout->setCacheEnabled(true);
+    m_layout->setFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
+}
+
+HighlightedText::~HighlightedText() = default;
+
+void HighlightedText::setText(const QStringList& text)
+{
+    const bool usesAnsi = std::any_of(text.cbegin(), text.cend(),
+                                      [](const QString& line) { return line.contains(QLatin1Char('\u001B')); });
```
milianw, line 198: again this magic char, what does it mean? if it has to be in sync with the other place, introduce it as a constexpr global with a sane name

src/resultsdisassemblypage.cpp
```diff
@@ -177,9 +177,11 @@ class BranchDelegate : public QStyledItemDelegate
 
         auto pen = QPen(option.palette.color(QPalette::Link), 1);
         pen.setCosmetic(true);
+        const auto oldPen = painter->pen();
```
milianw, line 180: the commit message for this patch has an error - duplicate `will will`

milianw (changes requested):
the cleanup/fixup patches should go into a separate MR because they can land directly. the rest of this patch needs some cleanup

src/models/disassemblymodel.cpp
lievenhey: ah it moved into the next patch, dammit

src/models/highlightedtext.cpp
lievenhey: well its know as escape character
if you look at the ascii table you can see it there
```
       032   26    1A    SUB (substitute)            132   90    5A    Z
       033   27    1B    ESC (escape)                133   91    5B    [
       034   28    1C    FS  (file separator)        134   92    5C    \  '\\'
```

milianw: you cannot expect people have the ASCII table in their mind when reading the code. the point is: your code is unreadable as-is. make it readable by introducing a named variable:

```
constexpr auto escapeCode = QLatin1Char('\u001B');
```

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -207,30 +207,32 @@ private slots:
...
+        };
+
         for (const auto& line : result.disassemblyLines) {
             QVERIFY(!line.branchVisualisation.isEmpty());
 
             // check that we only captures valid visualisation characters
             QVERIFY(std::all_of(line.branchVisualisation.cbegin(), line.branchVisualisation.cend(),
                                 isValidVisualisationCharacter));
 
-            QVERIFY(!line.disassembly.isEmpty());
```
milianw, line 225: why did you remove the old code that checks that the disassembly wasn't empty? that should still be the case, no?

src/models/disassemblyoutput.cpp
```diff
@@ -17,28 +17,43 @@
 #include <QProcess>
 #include <QStandardPaths>
 
+#include "formattingutils.h"
+
 namespace {
 Q_LOGGING_CATEGORY(disassemblyoutput, "hotspot.disassemblyoutput")
 
-bool canVisualizeJumps(const QString& objdump)
+QByteArray objdumpHelp(const QString& objdump)
```
milianw, line 25: unrelated cleanup -> separate patch

src/models/disassemblyoutput.cpp
```diff
@@ -17,28 +17,43 @@
...
-    return help.contains("--visualize-jumps");
+    return process.readAllStandardOutput();
+}
+
+bool canVisualizeJumps(const QString& objdump)
+{
+    return objdumpHelp(objdump).contains(QByteArrayLiteral("--visualize-jumps"));
+}
+
+bool canUseSyntaxHighlighting(const QString& objdump)
```
milianw, line 42: the patch that introduces disassembler highlighting should be a separate one again

src/models/disassemblyoutput.cpp
milianw: this wasn't done yet?

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    void themeChanged() override
+    {
+        if (!m_repository) {
+            return;
+        }
+
+        KSyntaxHighlighting::Repository::DefaultTheme theme;
+        if (QPalette().base().color().lightness() < 128) {
```
milianw, line 62: this cannot work, can it? see: https://doc.qt.io/qt-6/qpalette.html#QPalette

you want to use https://doc.qt.io/qt-6/qguiapplication.html#palette or something instead, no?

tests/modeltests/tst_formatting.cpp
```diff
@@ -0,0 +1,45 @@
...
+class TestFormatting : public QObject
+{
+    Q_OBJECT
+private slots:
+    void testRemoveAnsi_data()
+    {
+        QTest::addColumn<QString>("ansiString");
+        QTest::addColumn<QString>("ansiFreeString");
+
+        QTest::addRow("no ansi sequence") << QStringLiteral("[30m A B [0m C") << QStringLiteral("[30m A B [0m C");
```
milianw, line 25: ansi highlighting should be done in a separate patch

src/models/highlightedtext.cpp
milianw: you even introduced this now (in the wrong patch) - use it here too

milianw (changes requested):

src/models/highlightedtext.cpp
lievenhey: good to know, I copied that code from another location, will fix both

tests/modeltests/tst_disassemblyoutput.cpp
lievenhey: no, sometimes we encounter `nop`s like this:
```asm
    124a:       f2 0f 10 4c 24 08       movsd  0x8(%rsp),%xmm1
    1250:       f2 0f 10 44 24 10       movsd  0x10(%rsp),%xmm0
    1256:       f2 0f 58 0d aa 0d 00    addsd  0xdaa(%rip),%xmm1        # 2008 <_IO_stdin_used+0x8>
    125d:       00 
    125e:       ff c3                   inc    %ebx
    1260:       81 fb 40 42 0f 00       cmp    $0xf4240,%ebx
    1266:       0f 85 64 ff ff ff       jne    11d0 <main+0x50>
```

GitMensch: That's important information, maybe add here as comment?

GitMensch:
This also fixes #527 (currently not related, you may want to do that on the commit message).

Can you please include the logs https://github.com/KDAB/hotspot/files/13114752/dump.log and https://github.com/KDAB/hotspot/files/13114535/log.txt in the testsuite to ensure they can be processed?

tests/modeltests/tst_disassemblyoutput.cpp
milianw: I agree, add a comment - better yet explicitly catch it - i.e. expect empty when the hexdump is `00`, otherwise expect non-empty

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include <KSyntaxHighlighting/State>
+#include <KSyntaxHighlighting/SyntaxHighlighter>
+#include <KSyntaxHighlighting/Theme>
+#endif
+
+#include "formattingutils.h"
+
+namespace {
+#if KFSyntaxHighlighting_FOUND
+class KSyntaxHighlightingImplementation : public HighlightingImplementation,
```
milianw, line 30: don't use the `K` prefix here, to ensure we don't run into name clashes in the future

src/models/highlightedtext.h
```diff
@@ -0,0 +1,67 @@
...
+class Definition;
+class Repository;
+}
+
+class HighlightingImplementation
+{
+    Q_DISABLE_COPY(HighlightingImplementation)
+public:
+    HighlightingImplementation() = default;
+    virtual ~HighlightingImplementation() = default;
```
milianw, line 29: move the `= default` to the .cpp - we don't want to inline the vtable into every TU that includes this header

src/models/highlightedtext.h
```diff
@@ -0,0 +1,67 @@
...
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include <QObject>
+#include <QTextLayout>
+#include <QTextLine>
+
+#include <KColorScheme>
```
milianw, line 14: should be unused in the header?

src/models/highlightedtext.h
```diff
@@ -0,0 +1,67 @@
...
+    SPDX-FileCopyrightText: Lieven Hey <lieven.hey@kdab.com>
+    SPDX-FileCopyrightText: 2023 Klarälvdalens Datakonsult AB, a KDAB Group company, info@kdab.com
+
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#pragma once
+
+#include <QObject>
+#include <QTextLayout>
```
milianw, line 11: forward declare

src/models/highlightedtext.h
```diff
@@ -0,0 +1,67 @@
...
+#include <QTextLine>
```
milianw, line 12: dito

src/models/highlightedtext.h
```diff
@@ -0,0 +1,67 @@
...
+private slots:
+    void applyFormatting();
+
+private:
+    void updateHighlighting();
+
+#if KFSyntaxHighlighting_FOUND
+    KSyntaxHighlighting::Repository* m_repository;
+#endif
+    std::unique_ptr<HighlightingImplementation> m_highlighter;
```
milianw, line 87: missing <memory> include?

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    SPDX-License-Identifier: GPL-2.0-or-later
+*/
+
+#include "highlightedtext.h"
+
+#include <QGuiApplication>
+#include <QPalette>
+
+#include <memory>
```
milianw, line 13: move to header

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+#include <KColorScheme>
```
milianw, line 15: empty line below

src/models/highlightedtext.h
```diff
@@ -0,0 +1,67 @@
...
+}
+
+class HighlightingImplementation
+{
+    Q_DISABLE_COPY(HighlightingImplementation)
+public:
+    HighlightingImplementation() = default;
+    virtual ~HighlightingImplementation() = default;
+
+    virtual QVector<QTextLayout::FormatRange> format(const QStringList& text) = 0;
```
milianw, line 31: why is this a pure virtual interface? there is only one implementation at compile time - remove all the virtuals here from the header

milianw: also, this class is not public - meaning people should only deal with HighlightedText below which uses it internally but otherwise doesn't provide a getter to this class. As such, forward declare this here in the header and move all code to the .cpp

src/models/highlightedtext.cpp
milianw: see comments in the header: I don't understand why you have this indirection at all. Just make this the `class HighlightingImplementation` and be done with it - no need for multiple inheritance or the `KSyntaxHighlightingImplementation` at all

src/models/disassemblyoutput.cpp
```diff
@@ -20,17 +20,21 @@
...
+        return {};
     }
-    const auto help = process.readAllStandardOutput();
-    return help.contains("--visualize-jumps");
+    return process.readAllStandardOutput();
+}
+
+bool canVisualizeJumps(const QString& objdump)
+{
+    return objdumpHelp(objdump).contains(QByteArrayLiteral("--visualize-jumps"));
```
milianw, line 39: sorry, I was wrong - the QBAL isn't needed here (see the old code - just `.contains("--visualize-jumps")`

src/models/highlightedtext.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include <KSyntaxHighlighting/AbstractHighlighter>
+#include <KSyntaxHighlighting/Definition>
+#include <KSyntaxHighlighting/Format>
+#include <KSyntaxHighlighting/Repository>
+#include <KSyntaxHighlighting/State>
+#include <KSyntaxHighlighting/SyntaxHighlighter>
+#include <KSyntaxHighlighting/Theme>
+#endif
+
+#include "formattingutils.h"
```
milianw, line 26: the .cpp for this is only added to CMakeLists.txt in a later patch

src/models/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ add_library(
     disassemblyoutput.cpp
     eventmodel.cpp
     filterandzoomstack.cpp
+    formattingutils.cpp
```
milianw, line 14: a previous patch already used the header for this

src/models/disassemblyoutput.cpp
```diff
@@ -37,12 +39,21 @@ bool canVisualizeJumps(const QString& objdump)
     return objdumpHelp(objdump).contains(QByteArrayLiteral("--visualize-jumps"));
 }
 
-DisassemblyOutput::LinkedFunction extractLinkedFunction(const QString& disassembly)
+bool canUseSyntaxHighlighting(const QString& objdump)
+{
+    return objdumpHelp(objdump).contains(QByteArrayLiteral("--disassembler-color"));
```
milianw, line 44: sorry, dito - no QBAL needed

src/models/highlightedtext.cpp
milianw: this isn't done

tests/modeltests/tst_disassemblyoutput.cpp
```diff
@@ -126,6 +126,7 @@ private slots:
         }
 
         file->write(text.toUtf8());
+        file->flush();
```
milianw, line 129: why was this not detected on the CI so far?

milianw (changes requested):

tests/modeltests/tst_disassemblyoutput.cpp
lievenhey: :shrug: it worked on older versions, so I guess ether the host or the ci container got an update which changed  how the write calls are handled

GitMensch:
Rechecked: while the appimage 211 showed the syntax-highlighted disassembly and used "non-7bit arrows" (and a warning message `unexepected jump visualization char: +`)
highlight
the current appimage has seems to have no syntax-highlighting any more
no-highlight
along with the warning
```
hotspot.disassemblyoutput: objdump binary does not support `--disassembler-color`: "/usr/bin/objdump"
```
changing the highlighter in the pull-down doesn't seem to have any effect at all

src/models/disassemblyoutput.cpp
```diff
@@ -238,19 +261,23 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
...
+        {
+            QString branchVisualisation;
+            QString hexdump;
+        };
+        const auto [branchVisualisation, hexdump] = [branchesAndHex = parts.value(1)]() -> BranchesAndHexdump {
+            auto firstHexIt = std::find_if(branchesAndHex.cbegin(), branchesAndHex.cend(), isHexCharacter);
+            auto size = std::distance(branchesAndHex.cbegin(), firstHexIt);
+            auto branchVisualisation = branchesAndHex.mid(0, size);
+            auto hexdump = branchesAndHex.mid(size);
+            return {branchVisualisation, hexdump};
```
GitMensch, line 274: Can you use `return {branchVisualisation, hexdump..trimmed()};` here? Otherwise we have trailing spaces, leading to eliding symbol "..." (and unnecessary amount of trailing spaces kept in memory).

GitMensch (changes requested):

lievenhey:
> ```
> hotspot.disassemblyoutput: objdump binary does not support `--disassembler-color`: "/usr/bin/objdump"
> ```
that error should definitely not happen, since I check if that flag is supported, I will take another look at the code

GitMensch:
Seems like the cmake-format CI error is easy to solve, but the failing test cases need more work.

GitMensch:
Just wanted to share that #561 seems to be fixed with the last commit (only checked two _different_ examples that did not work before) - if it should be then please add its reference to the commits / this PR.
Can you add the disassembly output provided in this issue to the testsuite?

One downside: it seems that the syntax highlighting in the Disassembly does not work (AppImage).

lievenhey:
@GitMensch can check you the most recent appimage? I notices that QTextLayout is very slow (multiple seconds for 50k+ lines) and added a cache. You complained about slow performance so maybe this finally fixes it.

GitMensch:
Checked one extreme example (all with appimages):

* "old" hotspot: ~110 seconds
* current master: 48-60 seconds
* this PR previous push: more seconds
* this PR after your last commit: 11-12 seconds - but the syntax highlighting in the disassembly is "kind of gone", compare
   master with
   this PR's state

lievenhey:
That is most likely, because `KColorScheme::ColorSet::Complementary` is not available. I will think of another way to get colors.

GitMensch:
> That is most likely, because `KColorScheme::ColorSet::Complementary` is not available. I will think of another way to get colors.

... and/or because of the following?

> Note that I do get a message
> 
> > hotspot.disassemblyoutput: objdump binary does not support `--disassembler-color`: "/usr/bin/objdump"

lievenhey:
Ohh, right, in that case it should fallback to using `KSyntaxHighlighter`. In that case it should use the original formatting. I will into it.

lievenhey:
@GitMensch I looked into it. Currently I only copy the font color from `KSyntaxHighlighter::Format`. If that is what you mean.

GitMensch:
Would there be any "thickness" it the objdump colored output that may should be copied over?

Note: in any case I'd "expect" to have the objdump coloring as an extra entry of the highlighting (create a manual list including all syntax highlighters that have "asm" in them, the add "objdump" on top, if it is available).
Last time checked there the original syntax highlighter doesn't work any more for disassembly at all.

GitMensch:
If it is more likely to get included faster, then we may split this PR in 4:
* separate hex column - at least according to my manual tests this seems to work fine
* a later follow-up PR for a possible automated testcase for the above
* move to QTextLayout including caching (seems to be finished and provide better performance - in this case an early distribution would be very useful) -> possibly in this PR
* disassembly syntax highlighting with objdump

milianw:
code lgtm, probably needs a rebase now but otherwise this should be ready to go in IIUC? @GitMensch is your highlighting bug resolved now in the latest iteration of this pull request?

GitMensch:
> GitMensch is your highlighting bug resolved now in the latest iteration of this pull request?

No, sadly #561 is not yet solved. Just tested with [objdump.log](https://github.com/KDAB/hotspot/files/13629286/objdump.cob_move_packed_to_display.log), generated by "GNU objdump (GNU Binutils for Debian) 2.35.2", which doesn't support `--disassembler-color`, with `objdump -d -l -C --start-address 0x2ba15 --stop-address 0x2bccf --visualize-jumps /tmp/gnucobol-bug-hunting/libcob/.libs/libcob.so.4.2.0`, result 
broken disassembly

I guess that the actual highlighting is not the main problem but the splitting of the jump-visualisation and hex-values.

I'm keen to recheck after a rebase (shouldn't make a big difference, I guess) and even more after a fix to this.

Just a debugging note (you're likely aware of this): you may add an "objdump" script that only does a `cat` on this file, to see the result in your environment.

lievenhey:
Thanks, I finally understand your problem and managed to reproduce it.

lievenhey:
@GitMensch It seems that #566 fixed the issue, can you check again?

GitMensch:
Yes, it is and the current appimage from this PR works fine there, too.
Also tested with a bunch of other recordings. 🥳 

As Milian is also fine with this PR it seems its ready for merge :-)

GitMensch:
@milianw can you merge that in, please?

milianw (approved):

Issue #477, opened 2023-03-24
recognize perfparser format (peek into file on open)


GitMensch:
**Is your feature request related to a problem? Please describe.**
When naming a file something.perfparser (or even"banana") opening the file via hotspot does not work "The hotspot-perfparser binary existed with code 4 (invalid perf data file)."

**Describe the solution you'd like**
peek into the file specified at command line or via open dialog 

**Describe alternatives you've considered**
Always remember to name the file correctly...

PR #539, opened 2023-10-24
peek into file before open

GitMensch:
* check file open, providing nice and early error
* explicit check for perfparser (magic number QPERFSTREAM) opening directly without the need to fall back to file name
* explicit check for V1 perf data which isn't supported by perfparser
* explicit check for unknown file type

Fixes: https://github.com/KDAB/hotspot/issues/477


src/parsers/perf/perfparser.cpp
```diff
@@ -1583,23 +1583,35 @@ void PerfParser::startParseFile(const QString& path)
...
+            return;
+        }
+
+        if (file.peek(11) == "QPERFSTREAM") {
             d.setInput(&file);
             while (!file.atEnd() && !d.stopRequested) {
                 if (!d.tryParse()) {
-                    emit parsingFailed(tr("Failed to parse file"));
+                    // TODO: provide reason
+                    emit parsingFailed(tr("Failed to parse file %1: %2").arg(path, QStringLiteral("unknown reason")));
```
milianw, line 1597: inconsistent capitalization - note the `File format unknown` starts upper case here, but this message and the one in the middle uses lower case.

milianw (changes requested):
one minor issue, otherwise lgtm

GitMensch:
So should I capitalize Unknown?

milianw:
> So should I capitalize Unknown?

yes, and the second message in the middle too. ideally, rephrase it to the more terse

> Unsupported V1 perf data

GitMensch:
I've started to add some tests. I can already use the tests/integrationtests/custom_cost_aggregation_testfiles/custom_cost_aggregation.perfparser file (use it both under its original name and copy it to a local file "fruitper" (opening works as expected in both cases), but I do miss a matching perf.data file. Can you create and upload it (for later tests maybe multiple versions [with/without `--call-graph`])? Do you have a PERFFILE (perf V1) anywhere [or even a Linux < 3.7 to create it]? I've only hex-edited a text-file, to trigger that message so far.

GitMensch:
texts adjusted; note that the place to do the checks was initially wrong, fixed that along with starting to add tests (not finished yet)

GitMensch:
tests finished so far, we can add perfv1+perfv2 tests when those are available; I consider that ready to pull now

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+    void testFileErrorHandling()
+    {
+        PerfParser parser(this);
+        QSignalSpy parsingFailedSpy(&parser, &PerfParser::parsingFailed);
+        QString message;
+
+        const auto notThereFile = QLatin1String("not_here");
+        parser.initParserArgs(notThereFile);
+
+        COMPARE_OR_THROW(parsingFailedSpy.count(), 1);
```
milianw, line 184: just `QCOMPARE` - the `COMPARE_OR_THROW` is only needed in helper functions that don't return `void`

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
         m_capabilities = host.perfCapabilities();
     }
 
+    void testFileErrorHandling()
+    {
+        PerfParser parser(this);
+        QSignalSpy parsingFailedSpy(&parser, &PerfParser::parsingFailed);
+        QString message;
+
+        const auto notThereFile = QLatin1String("not_here");
```
milianw, line 181: `QStringLiteral`

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        parser.initParserArgs(notThereFile);
+
+        COMPARE_OR_THROW(parsingFailedSpy.count(), 1);
+        message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
+        // qDebug("Error message is '%s'", qPrintable(message));
```
milianw, line 186: cleanup

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        QVERIFY(message.contains(notThereFile));
+        QVERIFY(message.contains(QLatin1String("does not exist")));
+
+        const auto parentDirs = QLatin1String("../..");
```
milianw, line 190: QStringLiteral

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        parser.initParserArgs(parentDirs);
+        // note: initializing parser args reset the attached spy counter
+        COMPARE_OR_THROW(parsingFailedSpy.count(), 1);
```
milianw, line 193: QCOMPARE

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
+        QVERIFY(message.contains(parentDirs));
+        QVERIFY(message.contains(QLatin1String("is not a file")));
+
+        const auto noRead = QLatin1String("no_r_possible");
```
milianw, line 198: dito

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        const auto shell = QStandardPaths::findExecutable(QStringLiteral("bash"));
+        if (!shell.isEmpty()) {
+            QProcess::execute(
+                shell, {QLatin1String("-c"), QStringLiteral("rm -rf %1 && touch %1 && chmod a-r %1").arg(noRead)});
+            parser.initParserArgs(noRead);
+
+            COMPARE_OR_THROW(parsingFailedSpy.count(), 1);
```
milianw, line 205: dito

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+
+        const auto perfData = QFINDTESTDATA("custom_cost_aggregation_testfiles/custom_cost_aggregation.perfparser");
+        QVERIFY(!perfData.isEmpty() && QFile::exists(perfData));
+
+        parser.startParseFile(perfData);
+
+        VERIFY_OR_THROW(parsingFinishedSpy.wait(6000));
+
+        // Verify that the test passed
+        COMPARE_OR_THROW(parsingFailedSpy.count(), 0);
```
milianw, line 227: QCOMPARE

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        QSignalSpy parsingFailedSpy(&parser, &PerfParser::parsingFailed);
+        QSignalSpy parsingFinishedSpy(&parser, &PerfParser::parsingFinished);
+        QString message;
+
+        const auto perfData = QFINDTESTDATA("custom_cost_aggregation_testfiles/custom_cost_aggregation.perfparser");
+        QVERIFY(!perfData.isEmpty() && QFile::exists(perfData));
+
+        parser.startParseFile(perfData);
+
+        VERIFY_OR_THROW(parsingFinishedSpy.wait(6000));
```
milianw, line 224: QVERIFY

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+
+        // Verify that the test passed
+        COMPARE_OR_THROW(parsingFailedSpy.count(), 0);
+        COMPARE_OR_THROW(parsingFinishedSpy.count(), 1);
```
milianw, line 228: dito

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+
+        const auto shell = QStandardPaths::findExecutable(QStringLiteral("bash"));
+        if (!shell.isEmpty()) {
+            const auto perfDataSomeName = QLatin1String("fruitper");
```
milianw, line 232: QStringLiteral

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+            QProcess::execute(shell,
+                              {QLatin1String("-c"), QStringLiteral("cp -f %1 %2").arg(perfData, perfDataSomeName)});
```
milianw, line 234: QStringLiteral for the first one, QLatin1String for the ().arg'ed one

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+
+            parser.startParseFile(perfDataSomeName);
+
+            VERIFY_OR_THROW(parsingFinishedSpy.wait(6000));
```
milianw, line 238: QVERIFY

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+
+            // Verify that the test passed
+            COMPARE_OR_THROW(parsingFailedSpy.count(), 0);
```
milianw, line 241: QCOMPARE

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+            COMPARE_OR_THROW(parsingFinishedSpy.count(), 2);
```
milianw, line 242: dito

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        }
+
+        // check for invalid data
+        const auto noPerf = QFINDTESTDATA("custom_cost_aggregation_testfiles/by_cpu.txt");
+        QVERIFY(!noPerf.isEmpty() && QFile::exists(noPerf));
+        parser.startParseFile(noPerf);
+        COMPARE_OR_THROW(parsingFailedSpy.count(), 1);
```
milianw, line 249: dito

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +172,107 @@ private slots:
...
+        message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
+        QVERIFY(message.contains(noPerf));
+        QVERIFY(message.contains(QLatin1String("File format unknown")));
+
+        // TODO: check for PERFv1 (missing: data file)
```
milianw, line 254: remove from patch then - never commit commented out code

milianw (changes requested):

tests/integrationtests/tst_perfparser.cpp
GitMensch: both `QCOMPARE` and `QVERIFY` were taken from the other test codes in the exact same file... but I'll change all of them in the same commit, then...

GitMensch:
Requested changes done, PERFv1 and PERFv2 added, fixed opening of perf.data.gz (now also covered by the testsuite).

GitMensch:
I've split this into two commits: first the missing tests, then the adjustments which also provide the original requested feature, with the tests adjusted. Those test changes also show that we don't need to wait for a failing parse - because we get the parsing errors before returning, theres's no perfparser binary started any more in case of obvious errors.

Note that tinkering with the tests showed one issue: we seem to cache elf files, but only those that we find, it would be good to also cache "invalid" / empty returns, in which case it should be possible to decrease the wait time for those tests; I
lll add that piece to #551.

.gitignore
GitMensch, line 1: just a note: I always got an actual.txt file when running the tests that showed as unstaged change - so I've adjusted this entry

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -743,8 +837,8 @@ private slots:
 
         VERIFY_OR_THROW(recordingFinishedSpy.wait(10000));
 
-        COMPARE_OR_THROW(recordingFailedSpy.count(), 0);
```
GitMensch, line 746: this and the changes below come from the first/second review from @milianw; I've adjusted this in both the new and the old tests (which were used as a template for the original ones)

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +173,99 @@ private slots:
...
+
+        // check for PERFv2
+        const auto perf2 = QFINDTESTDATA("file_content/perf.data.true.v2");
+        QVERIFY(!perf2.isEmpty() && QFile::exists(perf2));
+        parser.startParseFile(perf2);
+        QVERIFY(parsingFinishedSpy.wait(
+            8000)); // FIXME: this waittime and the one below must be so high as unavailable elf files are checked
+                    // again and again via PerfSymbolTable::registerElf, as seen in the output
+        QCOMPARE(parsingFailedSpy.count(), 0);
+        QCOMPARE(parsingFinishedSpy.count(), 2);
```
GitMensch: This is _by design_ a valid perf file which cannot be completely parsed.

Running this results in multiple identical outputs of

> PerfUnwind::MissingElfFile: Could not find ELF file for 64/ld-2.17.so. This can break stack unwinding and lead to missing symbols.

and the test hints at that this is related to this test taking so long (it runs faster if I add that file).

We seem to cache elf files (return of ` PerfSymbolTable::registerElf()`), but only those that we can resolve.
In this case it would be good to also cache "invalid" / empty returns; allowing to also decrease the wait time for those tests here.
l'm copying this piece to #551, as working on that is totally unrelated to this PR.

GitMensch:
Any idea why:

* the ubuntu-2204 test environment does not include perf?
* the test, that is not run at all, SIGSEGVs?

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -172,6 +173,99 @@ private slots:
...
+        Settings::instance()->setAppPath(
+            QFileInfo(QStandardPaths::findExecutable(QStringLiteral("true"))).dir().path());
+        Settings::instance()->setExtraLibPaths(QFINDTESTDATA("file_content"));
+        PerfParser parser(this);
+        QSignalSpy parsingFailedSpy(&parser, &PerfParser::parsingFailed);
+        QSignalSpy parsingFinishedSpy(&parser, &PerfParser::parsingFinished);
+        QString message;
+
+        // check pre-exported perfparser
+        const auto perfData = QFINDTESTDATA("file_content/true.perfparser");
```
milianw, line 223: please make this a data driven test, i.e. use

```
void testFileContent_data()
{
    QTest::addColumn<QString>("file");
    QTest::addColumn<bool>(isValid);
    QTest::addRow("true.perfparser") << QFINDTESTDATA("file_content/true.perfparer");
    ...
}

void testFileContent()
{
    QFETCH(QString, perfFile);
    QFETCH(bool, isValid);
    ...
    QTRY_COMPARE(parsingFinishedSpy.count(), 1);
    QCOMPARE(parsingFailedSpy.count(), isValid ? 0 : 1);
    ...
}
```

tests/integrationtests/tst_perfparser.cpp
milianw: great, thanks - but please make this a separate commit - first do the cleanup, then add the new work in a second patch on top

milianw (changes requested):

tests/integrationtests/tst_perfparser.cpp
GitMensch: done (with an explicit timeout and the message instead of the boolean)

GitMensch:
lol, clang_lazy complains about `QTRY_COMPARE_WITH_TIMEOUT` (which is also used by `QTRY_COMPARE`, just with the default timeout 5000) to use `QString::fromUtf8("QTestLib: This test case check (\"%1\") failed because the requested timeout (%2 ms) was too short, %3 ms would have been sufficient this time.");` instead of `QStringLiteral`. The only option to solve that would to tinker with the QT macro or completely refrain from using it.

milianw:
> lol, clang_lazy complains about `QTRY_COMPARE_WITH_TIMEOUT`

odd bug, esp. considering we do `-isystem /usr/include/qt` - I'll investigate and potentially update clazy or so. we can ignore that for now - but there are two little nits I'd like to see fixed first please

milianw:
> Any idea why:
> 
>     * the ubuntu-2204 test environment does not include perf?

we cannot run perf properly anyways inside the VM I believe, though maybe the task-clock backend is good enough for some basic testing. generally, it's a tricky one to get working in a portable fashion - I am not very fond of it anymore. it breaks every other day, when we update the compiler, elfutils or perf...

>     * the test, that is not run at all, SIGSEGVs?

what test, backtrace?

GitMensch:
Thanks for the information; the SIGSEGV does not occur any more...

GitMensch:
> > lol, clang_lazy complains about `QTRY_COMPARE_WITH_TIMEOUT`
> 
> odd bug, esp. considering we do `-isystem /usr/include/qt`

I agree.

> I'll investigate and potentially update clazy or so. we can ignore that for now

Thanks.

> but there are two little nits I'd like to see fixed first please

Which ones? While learning about QT and its `_data` test was quite interesting, I'd really like to get that out of my view soon :-)

milianw:
> > but there are two little nits I'd like to see fixed first please
> 
> Which ones? While learning about QT and its `_data` test was quite interesting, I'd really like to get that out of my view soon :-)

Don't you see the two review comments of mine?

GitMensch:
> Don't you see the two review comments of mine?

Only about splitting and making this data riven, both were pushed [currently 2 days ago](https://github.com/KDAB/hotspot/pull/539#event-10889844645)

milianw (changes requested):
for some reasons these review changes didn't get published :shrug:

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -223,17 +224,25 @@ private slots:
...
-        QTest::addRow("invalid data") << QFINDTESTDATA("tst_perfparser.cpp") << QStringLiteral("invalid perf data file")
-                                      << 1000;
-        QTest::addRow("PERF v1") << QFINDTESTDATA("file_content/perf.data.true.v1")
-                                 << QStringLiteral("invalid perf data file") << 1000;
+        const auto perfData = QFINDTESTDATA("file_content/true.perfparser");
+        QTest::addRow("pre-exported perfparser") << perfData << QString() << 2000;
+        const auto shell = QStandardPaths::findExecutable(QStringLiteral("bash"));
+        if (!shell.isEmpty()) {
+            const auto perfDataSomeName = QStringLiteral("fruitper");
+            QProcess::execute(shell,
```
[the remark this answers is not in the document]
GitMensch: Surely, as noted: Currently, I'm neither a C++ nor QT guy. Will adjust and drop the additional include which must have get into by IDE automation and missed to check that before commit.

Thank for the review.

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -181,6 +181,90 @@ private slots:
...
+    {
+        PerfParser parser(this);
+        QSignalSpy parsingFailedSpy(&parser, &PerfParser::parsingFailed);
+        QString message;
+
+        const auto notThereFile = QStringLiteral("not_here");
+        parser.initParserArgs(notThereFile);
+
+        QCOMPARE(parsingFailedSpy.count(), 1);
+        message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
```
milianw, line 194: `qvariant_cast<QString>(var)` -> `var.toString()`, repeats a few times below

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -181,6 +181,90 @@ private slots:
...
+        const auto parentDirs = QStringLiteral("../..");
+        parser.initParserArgs(parentDirs);
+        // note: initializing parser args reset the attached spy counter
+        QCOMPARE(parsingFailedSpy.count(), 1);
+        message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
+        QVERIFY(message.contains(parentDirs));
+        QVERIFY(message.contains(QLatin1String("is not a file")));
+
+        const auto noRead = QStringLiteral("no_r_possible");
+        const auto shell = QStandardPaths::findExecutable(QStringLiteral("bash"));
```
milianw, line 207: don't use bash and external processes for removing a file creating a new one and setting chmod on it - just use `QTemporaryFile` and the `QFile` API to set the permissions

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -181,6 +181,90 @@ private slots:
...
+           again and again via PerfSymbolTable::registerElf, as seen in the output */
+        QTest::addRow("PERF v2") << QFINDTESTDATA("file_content/perf.data.true.v2") << QString() << 8000;
+#if KFArchive_FOUND
+        QTest::addRow("PERF v2 gz") << QFINDTESTDATA("file_content/perf.data.true.v2.gz") << QString() << 9000;
+#endif
+    }
+
+    void testFileContent()
+    {
+        Settings::instance()->setAppPath(
```
milianw, line 246: why does this matter, can you please add a comment? we shouldn't need to do this

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -181,6 +181,90 @@ private slots:
...
+        QSignalSpy parsingFinishedSpy(&parser, &PerfParser::parsingFinished);
+
+        QFETCH(QString, perfFile);
+        QFETCH(QString, errorMessagePart);
+        QFETCH(int, waitTime);
+
+        QVERIFY(!perfFile.isEmpty() && QFile::exists(perfFile));
+        parser.startParseFile(perfFile);
+
+        if (errorMessagePart.isNull()) {
```
milianw, line 261: isEmpty

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -181,6 +181,90 @@ private slots:
...
+            QTRY_COMPARE_WITH_TIMEOUT(parsingFinishedSpy.count(), 1, waitTime);
+            QCOMPARE(parsingFailedSpy.count(), 0);
+        } else {
+            QTRY_COMPARE_WITH_TIMEOUT(parsingFailedSpy.count(), 1, waitTime);
+            QCOMPARE(parsingFinishedSpy.count(), 0);
+            const auto message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
```
milianw, line 267: toString

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -181,6 +181,90 @@ private slots:
...
+
+        const auto noRead = QStringLiteral("no_r_possible");
+        const auto shell = QStandardPaths::findExecutable(QStringLiteral("bash"));
+        if (!shell.isEmpty()) {
+            QProcess::execute(
+                shell, {QLatin1String("-c"), QStringLiteral("rm -rf %1 && touch %1 && chmod a-r %1").arg(noRead)});
+            parser.initParserArgs(noRead);
+
+            QCOMPARE(parsingFailedSpy.count(), 1);
+            message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
```
milianw, line 214: toString

tests/integrationtests/tst_perfparser.cpp
milianw: also, instead of testing two things one after the other in a bigger function, make this data driven too please - the only difference is whether the file exists or whether it's not readable - you can simply pass a bool to indicate these two states and handle it in the setup accordingly

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -223,17 +223,21 @@ private slots:
         QTest::addColumn<QString>("errorMessagePart");
         QTest::addColumn<int>("waitTime");
 
-        QTest::addRow("pre-exported perfparser") << QFINDTESTDATA("file_content/true.perfparser") << QString() << 2000;
-        QTest::addRow("invalid data") << QFINDTESTDATA("tst_perfparser.cpp") << QStringLiteral("invalid perf data file")
-                                      << 1000;
-        QTest::addRow("PERF v1") << QFINDTESTDATA("file_content/perf.data.true.v1")
-                                 << QStringLiteral("invalid perf data file") << 1000;
+        const auto perfData = QFINDTESTDATA("file_content/true.perfparser");
```
milianw, line 226: these hunks should be in the parent patch, or?

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -262,6 +266,7 @@ private slots:
             QCOMPARE(parsingFinishedSpy.count(), 0);
             const auto message = qvariant_cast<QString>(parsingFailedSpy.takeFirst().at(0));
             QVERIFY(message.contains(errorMessagePart));
+            QVERIFY(message.contains(perfFile));
```
milianw, line 269: dito, parent patch

milianw (changes requested):

tests/integrationtests/tst_perfparser.cpp
GitMensch: The AppPath? That removes noise because the binary for the recording is not necessarily in the same place as the one recorded. See #551 for the problem, should I reference this issue here and let it be?
If/when this issue is solved then _maybe_ this isn't needed any more, but this PR actually gives you the direct option to test and possibly do something about it.

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -184,6 +181,95 @@ private slots:
...
+        QTest::addRow("pre-exported perfparser") << perfData << QString() << 2000;
+        const auto perfDataSomeName = QStringLiteral("fruitper");
+        QFile::copy(perfData, perfDataSomeName); // we can ignore errors (file exist) here
+        QTest::addRow("pre-exported perfparser \"bad extension\"") << perfDataSomeName << QString() << 2000;
+        QTest::addRow("no expected magic header")
+            << QFINDTESTDATA("tst_perfparser.cpp") << QStringLiteral("File format unknown") << 1000;
+        QTest::addRow("PERF v1") << QFINDTESTDATA("file_content/perf.data.true.v1") << QStringLiteral("V1 perf data")
+                                 << 1000;
+
+        /* FIXME: this waittime and the one below must be so high as unavailable elf files are checked
```
GitMensch, line 236: I've thought I did create an issue for this, seems this isn't the case.
Do we need one now and then just leave a link to the issue here (possibly with the current text(?

tests/integrationtests/tst_perfparser.cpp
GitMensch: No because:
* only the new checks "the real patch here" do output the file name in the error message
* the new patch adjusts the error messages as per your requests (most were not checked at all), so we can't check them before the patch

If you want to then I can add to the "pre"-testing an 'expected' "can't parse the file if it doesn't match the expected filename", but I don't think that's anything useful to test.
Please note if there are other things you want to be included in the previous commit.

tests/integrationtests/tst_perfparser.cpp
GitMensch: dito - not possible in previous patch as until with the "real patch" the filename was not part of the error message

tests/integrationtests/tst_perfparser.cpp
GitMensch: this explicit matches `QString()`, which creates a Null String. `isEmpty` would catch both `""` and `Null`, but that is really intended to match the Null version. Should I still adjust that?

tests/integrationtests/tst_perfparser.cpp
GitMensch: Did as requested, sadly the temp file cannot be created in the "data" function, so I've used a null string as filename to request that to be done in the test function.

GitMensch:
@milianw Just a friendly ping - Is there anything open with this or the other PRs from my side?

tests/integrationtests/tst_perfparser.cpp
milianw: it shouldn't matter in practice whether it's null or empty - imo it's bogus that QString even gives you the ability to distinguish. I don't want us to use isNull anywhere on strings - it really shouldn't matter

tests/integrationtests/tst_perfparser.cpp
milianw: what I don't understand: the first patch introduces

```
       QTest::addRow("pre-exported perfparser") << QFINDTESTDATA("file_content/true.perfparser") << QString() << 2000;
```

then this patch here changes it to:

```
        const auto perfData = QFINDTESTDATA("file_content/true.perfparser");
        QTest::addRow("pre-exported perfparser") << perfData << QString() << 2000;
```

this part should be cleaned up and implemented directly in this way in the initial patch such that there is no change hunk at all here.

dito for the "PERF v1" case below

milianw (changes requested):

tests/integrationtests/tst_perfparser.cpp
milianw: alright, let's leave it as is then

milianw (approved):
OK, I removed one more usage of `QString::isNull` and added a patch to disable the clazy issues to get the CI green again - let's see if that's really the case then this can be merged

milianw:
yep, lgtm now, I'll merge it - thanks for your contribution and patience with me :)

PR #540, opened 2023-10-24
minor typo correction: unknown

GitMensch:
additional: reordered checks in brushSystem

### DIFF
--- a/src/flamegraph.cpp
+++ b/src/flamegraph.cpp
@@ -379,13 +379,10 @@ QBrush brushSystem(const Data::Symbol& symbol)
     static const auto user = QBrush(QColor(200, 200, 0, 125));
     static const auto unknown = QBrush(QColor(50, 50, 50, 125));
 
-    // I have seen [ only on kernel calls
+    // remark lievenhey: I have seen [ only on kernel calls
     if (symbol.path.isEmpty() || symbol.path.startsWith(QLatin1Char('['))) {
         return unknown;
-    } else if (isSystemPath(symbol.path)) {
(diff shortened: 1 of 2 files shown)


milianw (changes requested):
please split this up into two patches, otherwise this lgtm

src/main.cpp
```diff
@@ -193,6 +193,9 @@ int main(int argc, char** argv)
             << parser.helpText();
         return 1;
     }
+    const auto originalArguments = QCoreApplication::arguments();
```
milianw, line 196: why this change?

GitMensch:
Is one PR with two commits fine or so you want separate PRs?

src/main.cpp
GitMensch: The idea (I've reverted all but moving that declaration up-front) was to check the arguments before creating the window.
But it's not part of this PR, so I'm just reverting that.

There is now only one thing in this PR, so there's no split useful anymore, is it? If you still want the split: is one commit each (one for the name, the other for the message) enough or do you really like two PRs?

milianw: it's fine to have it in one merge request, but you have to make atomic commits - how many I don't care, but better more than less. each commit gets its own message with a motivational text to set the stage for me as a reviewer why a change is needed.

milianw:
> Is one PR with two commits fine or so you want separate PRs?

one merge request, many commits

GitMensch:
split as requested

milianw (approved):

PR #547, opened 2023-10-30
use python 3.11 to run pre-commit on the ci

lievenhey:
~the old pre-commit action broke and is no longer working I relpaced it with pre-commit-lite which support auto fixup~

python got updated to 3.12 which broke pre-commit


milianw (approved):
lgtm, thanks

PR #559, opened 2023-11-12
extend old binary filter instead of replacing it

### LINKED ISSUE #475: Bug: "Filter Out By Binary" cannot be stacked
Especially when recording system-wide or when otherwise recording multiple processes, for example via a shell script to get "combined results" as suggested in #464, you often end up with a profiling that includes not relevant system processes (for example `bash`, `make`, running compiles/link, ...).
Also when inspecting a "common" recording you may see two "big" functions that you are not interested in when checking the details.

Because of #4 and #292 we can filter by function/binary (works quite good and is often enough) or filter out a _single_ function/binary.

All "exclude" filters should "stack" automatic (a manual "reset filter" will be possible to empty the filter stack).
This allows for example to:
* filter in by binary INTEREST
* filter out by symbol A
* filter out by symbol B
or:
* filter out by binary shell
* filter out by binary make
* filter out by binary cc1

For the binaries: record separately, then combine (as noted in the linked issue); but with the functions that still would be nice.

GitMensch:
fixes #475

Nice to see that the work on #558 brought something more useful - an understanding how the context menu works and with that the option to debug the related issue :-)

### DIFF
--- a/src/models/filterandzoomstack.cpp
+++ b/src/models/filterandzoomstack.cpp
@@ -196,6 +196,9 @@ void FilterAndZoomStack::applyFilter(Data::FilterAction filter)
         filter.excludeSymbols += lastFilter.excludeSymbols;
         filter.includeSymbols += lastFilter.includeSymbols;
         filter.includeSymbols.subtract(filter.excludeSymbols);
+        filter.excludeBinaries += lastFilter.excludeBinaries;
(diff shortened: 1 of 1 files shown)


milianw:
thank you! the CI error false positive was fixed by Lieven in #565

PR #562, opened 2023-11-13
Fix KF6 build

nicolasfella:
TerminalInterfaceV2 was merged into TerminalInterface

Since Hotspot doesn't use any of the features of V2 we can just use TerminalInterface everywhere


milianw:
thanks!

PR #566, opened 2023-12-05
add option to hide hexdump

### LINKED ISSUE #527: Disassembly view: allow to disable hex by moving it to a separate column

lievenhey:
closes #527

### DIFF
--- a/src/disassemblysettingspage.ui
+++ b/src/disassemblysettingspage.ui
@@ -32,14 +32,14 @@
    <property name="bottomMargin">
     <number>0</number>
    </property>
-   <item row="1" column="0">
+   <item row="2" column="0">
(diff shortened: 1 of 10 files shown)


lievenhey:
I just moved that here so that #538 is smaller

milianw (approved):

milianw:
thanks

Issue #521, opened 2023-10-11
Disassembly view: limit syntax highlighters


GitMensch:
**Is your feature request related to a problem? Please describe.**
I've selected by error HTML for the disassembly, with the result that I've got 100% cpu for several minutes, then no syntax highlighting visible.
The list (at least on the right side) is too long it really should only contain a reduced list.

**Describe the solution you'd like**
Filter the list for the syntax highlighting on the right, either only for "None" and "GNU Assembler" or, if `objdump` may also return other formats (I'm not sure) also for everything that matches `ssembler\|ASM`.

**Describe alternatives you've considered**
Take longer time to search for the right assembly and for going from None to GNU Assembler.

**Additional context**
It would also be reasonable to filter on the left side, but I'm not sure which ones could be filtered out. I _think_ it would be reasonable to filter everything that generates no executable code (HTML, XML, JSON, Yaml, ...). But this issue is _mostly_ about the right side.

lievenhey:
`objdump` supports coloured output. It would be nice if we somehow could leverage that.
![grafik](https://github.com/KDAB/hotspot/assets/82457690/908e0115-2a9a-4fb8-9c7c-3319961fc847)

GitMensch:
I agree that using `--disassembler-color=extended-color` and, when jumps are available also `--visualize-jumps=extended-color`  and show this in the QT window would be really nice. Using these options likely lead to terminal sequences, so if there isn't an QT object that can handle it we would need to manually color that using regex - this is likely too slow.

If that could be done, then we may do not need any syntax highlighting in there any more, right?
... but as that is a binutils Oct 2022 feature, we would need the current one at least as a fallback (another Disassembly setting :-)

@lievenhey Do you agree that it is reasonable to limit the usable syntax highlighting here? If it is _just_ GNU Assembler or None, then we could replace that with a checkbox (enabled by default, stored "internally" as the column orders in callee are stored, if they are).

PR #567, opened 2023-12-05
limit assembly highlighters

lievenhey:
KSyntaxHighligthing::Definition provides section() which can be used to filter out all non assembly highlighters, since they are not usefull and only cluttering the combobox.

closes: #521


milianw:
thanks

Issue #569, opened 2023-12-07
Fails to build with kddockwidgets-2.0.0: error: no viable conversion from 'const QVector<KDDockWidgets::Core::DockWidget *>' to 'const QVector<KDDockWidgets::DockWidgetBase *>'


yurivict:
**Describe the bug**
```
wrkdirs/usr/ports/devel/hotspot/work/hotspot-1.4.1/src/mainwindow.cpp:265:36: error: no viable conversion from 'const QVector<KDDockWidgets::Core::DockWidget *>' to 'const QVector<KDDockWidgets::DockWidgetBase *>'
    m_resultsPage->initDockWidgets(restored);
                                   ^~~~~~~~
/usr/local/include/qt5/QtCore/qvector.h:72:12: note: candidate constructor not viable: no known conversion from 'const Vector<Core::DockWidget *>' (aka 'const QVector<KDDockWidgets::Core::DockWidget *>') to 'const QVector<KDDockWidgets::DockWidgetBase *> &' for 1st argument
    inline QVector(const QVector<T> &v);
           ^
/usr/local/include/qt5/QtCore/qvector.h:75:5: note: candidate constructor not viable: no known conversion from 'const Vector<Core::DockWidget *>' (aka 'const QVector<KDDockWidgets::Core::DockWidget *>') to 'QVector<KDDockWidgets::DockWidgetBase *> &&' for 1st argument
    QVector(QVector<T> &&other) noexcept : d(other.d) { other.d = Data::sharedNull(); }
    ^
/usr/local/include/qt5/QtCore/qvector.h:79:12: note: candidate constructor not viable: no known conversion from 'const Vector<Core::DockWidget *>' (aka 'const QVector<KDDockWidgets::Core::DockWidget *>') to 'std::initializer_list<DockWidgetBase *>' for 1st argument
    inline QVector(std::initializer_list<T> args);
           ^
/usr/local/include/qt5/QtCore/qvector.h:70:14: note: explicit constructor is not a candidate
    explicit QVector(int size);
             ^
/usr/local/include/qt5/QtCore/qvector.h:83:14: note: explicit constructor is not a candidate
    explicit QVector(QArrayDataPointerRef<T> ref) noexcept : d(ref.ptr) {}
             ^
/wrkdirs/usr/ports/devel/hotspot/work/hotspot-1.4.1/src/resultspage.h:55:73: note: passing argument to parameter 'restored' here
[...]

PR #570, opened 2023-12-07
add support for kddockwidgets 2.0

lievenhey:
this patch fixes the compile errors with kddockwidgets 2.0 but keeps compatibility with earlier versions

fixes: #569


lievenhey:
I plan to keep building ubuntu 20.04 with kddw 1.6 and let the rest use kddw 2.0. This should take care that we stay compatible with earlier versions.

src/resultspage.cpp
```diff
@@ -254,7 +261,7 @@ void ResultsPage::onOpenEditor(const Data::Symbol& symbol)
 void ResultsPage::selectSummaryTab()
 {
     m_summaryPageDock->show();
-    m_summaryPageDock->setFocus();
+    // m_summaryPageDock->setFocus();
```
milianw, line 264: ?

src/resultspage.cpp
```diff
@@ -312,14 +319,26 @@ void ResultsPage::showError(const QString& message)
     QTimer::singleShot(5000, ui->errorWidget, &KMessageWidget::animatedHide);
 }
 
-void ResultsPage::initDockWidgets(const QVector<KDDockWidgets::DockWidgetBase*>& restored)
+void ResultsPage::initDockWidgets(const QVector<CoreDockWidget*>& restored)
 {
-    Q_ASSERT(restored.contains(m_summaryPageDock));
+#if KDDOCKWIDGETS_VERSION < KDDOCKWIDGETS_VERSION_CHECK(2, 0, 0)
```
milianw, line 324: make this a util function and call it here, i.e.
```
auto summaryPageDock = toDockWidget(m_summaryPageDock)
```

src/resultspage.cpp
```diff
@@ -312,14 +319,26 @@ void ResultsPage::showError(const QString& message)
...
+#endif //  KDDOCKWIDGETS_VERSION < KDDOCKWIDGETS_VERSION_CHECK
+
+    Q_ASSERT(restored.contains(summaryPageDock));
 
     const auto docks = {m_bottomUpDock, m_topDownDock,     m_flameGraphDock, m_callerCalleeDock,
                         m_timeLineDock, m_disassemblyDock, m_frequencyDock};
     for (auto dock : docks) {
-        if (!dock || restored.contains(dock))
+#if KDDOCKWIDGETS_VERSION < KDDOCKWIDGETS_VERSION_CHECK(2, 0, 0)
+        CoreDockWidget* dockWidget = dock;
```
milianw, line 336: dito

src/dockwidgets.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <kddockwidgets/kddockwidgets_version.h>
+
+#if KDDOCKWIDGETS_VERSION < KDDOCKWIDGETS_VERSION_CHECK(2, 0, 0)
+namespace KDDockWidgets {
+class DockWidgetBase;
+class DockWidget;
+class MainWindow;
+}
+using DockWidget = KDDockWidgets::DockWidget;
```
milianw, line 18: add empty line above

src/dockwidgets.h
```diff
@@ -0,0 +1,36 @@
...
+using DockMainWindow = KDDockWidgets::MainWindow;
+using CoreDockWidget = KDDockWidgets::DockWidgetBase;
+#else
+namespace KDDockWidgets {
+namespace QtWidgets {
+class DockWidget;
+class MainWindow;
+}
+namespace Core {
```
milianw, line 27: add empty line above

src/dockwidgets.h
```diff
@@ -0,0 +1,36 @@
...
+class DockWidget;
+}
+
```
milianw, line 30: remove empty line

milianw (changes requested):

GitMensch:
Can you please also update https://github.com/KDAB/hotspot/blob/master/scripts/appimage/Dockerfile to use the new version? While the docs say that 2.0 doesn't brings much from the user-view, it seems to be reasonable to update KDAB software in its appimages.

src/resultspage.cpp
lievenhey: forgot to change that back

Issue #534, opened 2023-10-19
applying a filter in huge recordings take very long (`QWidget::changeEvent()` calling `qHash`)


GitMensch:
**Describe the bug**
Applying a filter in huge recordings take very long. Maybe this can this be improved?

**To Reproduce**
Open a huge recording, then wait until it is opened, then apply a exclusion filter for a symbol - wait very long.

**Expected behavior**
Not taking several minutes for a filter to apply.

**Additional context**
I've recorded this long standing filtering from a separate hotspot instance (profiling for some seconds, then stopped), then opened that recording with hotspot.
Here are the results (recursive `QWidget::changeEvent()` calling `qHash(QString)` with ):
![flamegraph](https://github.com/KDAB/hotspot/assets/6699539/5bc1d444-76a1-41ba-b776-8c91be82118c)
![caller/callee](https://github.com/KDAB/hotspot/assets/6699539/41fe2344-c25a-4641-84c5-d1cbf964c2a0)

If useful, then I can upload the inspected perf.data.hotspot-filter (38,435,235 Bytes), taken from appimage, analyzed with appPath and debugPath set.

milianw:
the backtrace looks pretty broken to me, why would `QWidget::changeEvent` do a qHash lookup? are you really sure that the debug information matches the build used in the appimage?

milianw:
yeah, just double checked, and this is not it - I suspect the wrong Qt5Core debug information is used for some reason?

```
void QWidget::changeEvent(QEvent * event)
{
    switch(event->type()) {
    case QEvent::EnabledChange: {
        update();
#ifndef QT_NO_ACCESSIBILITY
        QAccessible::State s;
        s.disabled = true;
        QAccessibleStateChangeEvent event(this, s);
        QAccessible::updateAccessibility(&event);
#endif
        break;
    }

    case QEvent::FontChange:
    case QEvent::StyleChange: {
        Q_D(QWidget);
        update();
        updateGeometry();
        if (d->layout)
            d->layout->invalidate();
        break;
    }

    case QEvent::PaletteChange:
        update();
        break;

    case QEvent::ThemeChange:
        if (QGuiApplication::desktopSettingsAware() && windowType() != Qt::Desktop
            && qApp && !QCoreApplication::closingDown()) {
            if (testAttribute(Qt::WA_WState_Polished))
                QApplication::style()->unpolish(this);
            if (testAttribute(Qt::WA_WState_Polished))
                QApplication::style()->polish(this);
[...]

PR #571, opened 2023-12-10
[enhancement] cut 60% time for PerfParser::filterResults

MuchToMyDelight:
When change to "aggregate cost by symbol", filterResults spends a lot of time doing string comparison. This patch will reduce it dramatically.

Test on a 240MB perf.data/6 core CPU. The time of switching to "aggregate cost by symbol" is reduced from 50.x seconds to 19.x seconds.

### DIFF
--- a/src/callgraphgenerator.cpp
+++ b/src/callgraphgenerator.cpp
@@ -25,10 +25,10 @@ QHash<Data::Symbol, QString> writeGraph(QTextStream& stream, const Data::Symbol&
            << "\"]\n";
 
     stream << "node" << parentId << " [label=\"";
-    if (symbol.prettySymbol.isEmpty()) {
+    if (symbol.prettySymbol().isEmpty()) {
         stream << "??";
     } else {
-        stream << symbol.prettySymbol;
+        stream << symbol.prettySymbol();
     }
     stream << "\", color=\"" << settings->callgraphActiveColor().name() << "\"]\n";
 
@@ -52,7 +52,7 @@ void resultsToDot(int height, Direction direction, const Data::Symbol& symbol, c
         return;
     }
 
-    if (symbol.prettySymbol.isEmpty())
+    if (symbol.prettySymbol().isEmpty())
         return;
 
     if (results.selfCosts.numTypes() == 0) {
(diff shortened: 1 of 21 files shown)


milianw (changes requested):
I'm fine with this in principle, but you'll have to make all the members private and add getter/setter  API, as otherwise it's too easy to break the code by accidentally mutating a symbol member without updating the hash.

furthermore, while at it, the `qHash` code should be inlined and `qHash(Symbol)` should just forward the cached hash then directly

GitMensch:
Thank you for inspecting this. The approach seems quite reasonable.

Note: #534 is related - is the `qHash()` commonly called by that `std:tie()`?
In general: could you provide a perfparser file (File->save) of a recording before and after the change?

MuchToMyDelight:
> I'm fine with this in principle, but you'll have to make all the members private and add getter/setter API, as otherwise it's too easy to break the code by accidentally mutating a symbol member without updating the hash.
> 
> furthermore, while at it, the `qHash` code should be inlined and `qHash(Symbol)` should just forward the cached hash then directly

Hello milianw! I totally agree with you about that all members should be private, actually that was what I did for my first version. The only thing is that it involves tons of change to the code as struct Symbol is used widely. So I think it's better to show the idea firstly. I'll update the PR later.

MuchToMyDelight:
Updated the PR and here is how it's tested:
Firstly, use perf to record a systemwide 60s. Then open it with hotspot, profling the time cost of switching from aggreate cost by process to symbol, under the tab "Bottom up"

The base version: origin/master, notice that operator== takes 50.6% time. total time **24.x** seconds
base


The optimized version: the cost of operator== has gone, the time is reduced to **10.x** seconds
optimized

lievenhey:
Don't worry about the neonqt6 error, that one is already fixed in master, just rebase.

MuchToMyDelight:
passed clany-tidy and pre-commit. 
However, I found that this test failed in origin/master both with or without this PR
``` text
QWARN  : TestPerfParser::testCppInliningEventCyclesInstructions(leader-sampling) hotspot.perfparser: Unexpected attribute id: -1 Only know about 2 attributes so far
QWARN  : TestPerfParser::testCppInliningEventCyclesInstructions(leader-sampling) hotspot.perfparser: Unexpected attribute id: -1 Only know about 2 attributes so far
QFATAL : TestPerfParser::testCppInliningEventCyclesInstructions(leader-sampling) ASSERT failure in QVector<T>::operator[]: "index out of range", file /usr/include/qt/QtCore/qvector.h, line 457
FAIL!  : TestPerfParser::testCppInliningEventCyclesInstructions(leader-sampling) Received a fatal error.
   Loc: [Unknown file(0)]
Totals: 10 passed, 15 failed, 0 skipped, 0 blacklisted, 5196ms
********* Finished testing of TestPerfParser *********


89% tests passed, 1 tests failed out of 9

Total Test time (real) =  12.99 sec

The following tests FAILED:
          9 - tst_perfparser (Subprocess aborted)
Errors while running CTest

````

MuchToMyDelight:
In comparison, I submitted another pull #574 . I'm undecided which one is better. What's your opinion? @milianw @GitMensch

milianw:
Are you ok if we abandon/close this one in favor of the other pull request which is imo much nicer?

MuchToMyDelight:
closing this one, prefer #574

Issue #515, opened 2023-10-04
FR: in Caller/Callee view add Dissassembly option also in the "location"


GitMensch:
**Is your feature request related to a problem? Please describe.**
I can easily have a look at the function and from there the source lines that take most of the cpu instructions spent.
... but with large functions that span multiple source files it is really a hazzle to get to the code in the disassembly.

**Describe the solution you'd like**
* add "Disassembly" also in the location view
* on click: disassemble the current selected function, then auto-scroll to the source position

**Describe alternatives you've considered**
All manually, but that's really not nice for multi-source / large functions.

GitMensch:
While the feature is visible in the Location view, it seems to be always disabled ?!?

![grafik](https://github.com/KDAB/hotspot/assets/6699539/b9e35c9c-4808-4ed9-a3d6-48a8e8718c5e)

@lievenhey _Should_ this work already?

Should this be reopened?

lievenhey:
ah, it is because I am dump

GitMensch:
Seems to be fixed by #593 and of course raises the question if we could test anything here...

PR #573, opened 2023-12-12
CallerCallee add jump to source location in disassembly

lievenhey:
this patch adds a context menu that allows the user to select if he wants to open the source code or opens the disassembly and focus the selected location

closes: #515

### DIFF
--- a/src/resultscallercalleepage.cpp
+++ b/src/resultscallercalleepage.cpp
@@ -92,11 +92,6 @@ ResultsCallerCalleePage::ResultsCallerCalleePage(FilterAndZoomStack* filterStack
                                    ResultsUtil::CallbackAction::ViewDisassembly});
     ResultsUtil::setupHeaderView(ui->callerCalleeTableView, contextMenu);
     ResultsUtil::setupCostDelegate(m_callerCalleeCostModel, ui->callerCalleeTableView);
-    connect(ui->callerCalleeTableView, &QTreeView::activated, this, [this](const QModelIndex& index) {
-        auto symbol = index.data(CallerCalleeModel::SymbolRole).value<Data::Symbol>();
-        if (symbol.isValid())
-            emit jumpToDisassembly(symbol);
-    });
 
     connect(parser, &PerfParser::callerCalleeDataAvailable, this, [this](const Data::CallerCalleeResults& data) {
         m_callerCalleeCostModel->setResults(data);
@@ -170,7 +165,6 @@ ResultsCallerCalleePage::ResultsCallerCalleePage(FilterAndZoomStack* filterStack
     ui->sourceMapView->setContextMenuPolicy(Qt::CustomContextMenu);
     connect(ui->sourceMapView, &QTreeView::customContextMenuRequested, this,
             &ResultsCallerCalleePage::onSourceMapContextMenu);
-    connect(ui->sourceMapView, &QTreeView::activated, this, &ResultsCallerCalleePage::onSourceMapActivated);
 
     connect(ui->callerCalleeTableView->selectionModel(), &QItemSelectionModel::currentRowChanged, this,
             [selectCallerCaleeeIndex](const QModelIndex& current, const QModelIndex&) {
@@ -234,21 +228,23 @@ void ResultsCallerCalleePage::onSourceMapContextMenu(QPoint point)
 
     QMenu contextMenu;
     auto* viewCallerCallee = contextMenu.addAction(tr("Open in Editor"));
-    auto* action = contextMenu.exec(QCursor::pos());
-    if (action == viewCallerCallee) {
-        emit navigateToCode(location.path, location.lineNumber, 0);
-    }
-}
+    connect(viewCallerCallee, &QAction::triggered, this, [this, location, index] {
+        if (location) {
+            emit navigateToCode(location.path, location.lineNumber, 0);
+        } else {
+            const auto fileLine = index.data(SourceMapModel::FileLineRole).value<Data::FileLine>();
+            emit navigateToCodeFailed(tr("Failed to find file for location '%1'.").arg(fileLine.toString()));
+        }
+    });
 
-void ResultsCallerCalleePage::onSourceMapActivated(const QModelIndex& index)
-{
-    const auto location = toSourceMapLocation(index);
-    if (location) {
-        emit navigateToCode(location.path, location.lineNumber, 0);
-    } else {
-        const auto fileLine = index.data(SourceMapModel::FileLineRole).value<Data::FileLine>();
-        emit navigateToCodeFailed(tr("Failed to find file for location '%1'.").arg(fileLine.toString()));
-    }
+    auto line = index.data(SourceMapModel::FileLineRole).value<Data::FileLine>();
+    auto disassemblyAction = contextMenu.addAction(tr("Disassembly"));
+    const auto symbol = index.data(CallerCalleeModel::SymbolRole).value<Data::Symbol>();
+    disassemblyAction->setEnabled(symbol.canDisassemble());
+    connect(disassemblyAction, &QAction::triggered, this,
+            [this, symbol, line] { emit jumpToSourceCode(symbol, line); });
+
+    contextMenu.exec(QCursor::pos());
 }
 
 void ResultsCallerCalleePage::setSysroot(const QString& path)
--- a/src/resultscallercalleepage.h
+++ b/src/resultscallercalleepage.h
@@ -47,12 +47,12 @@ public:
 
 private slots:
     void onSourceMapContextMenu(QPoint pos);
-    void onSourceMapActivated(const QModelIndex& index);
 
 signals:
     void navigateToCode(const QString& url, int lineNumber, int columnNumber);
     void navigateToCodeFailed(const QString& message);
     void selectSymbol(const Data::Symbol& symbol);
+    void jumpToSourceCode(const Data::Symbol& symbol, const Data::FileLine& line);
     void jumpToDisassembly(const Data::Symbol& symbol);
 
 private:
--- a/src/resultsdisassemblypage.cpp
+++ b/src/resultsdisassemblypage.cpp
@@ -641,4 +641,12 @@ void ResultsDisassemblyPage::changeEvent(QEvent* event)
     }
 }
 
+void ResultsDisassemblyPage::jumpToSourceLine(const Data::FileLine& line)
+{
+    if (line.isValid()) {
+        ui->sourceCodeView->scrollTo(m_sourceCodeModel->indexForFileLine(line));
+        ui->assemblyView->scrollTo(m_disassemblyModel->indexForFileLine(line));
+    }
+}
+
 #include "resultsdisassemblypage.moc"
--- a/src/resultsdisassemblypage.cpp
+++ b/src/resultsdisassemblypage.cpp
@@ -274,19 +274,55 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
     connect(ui->assemblyView, &QTreeView::entered, this, updateFromDisassembly);
     connect(ui->sourceCodeView, &QTreeView::entered, this, updateFromSource);
 
-    ui->sourceCodeView->setContextMenuPolicy(Qt::CustomContextMenu);
-    connect(ui->sourceCodeView, &QTreeView::customContextMenuRequested, this, [this](QPoint point) {
-        const auto index = ui->sourceCodeView->indexAt(point);
-        const auto fileLine = index.data(SourceCodeModel::FileLineRole).value<Data::FileLine>();
-        if (!fileLine.isValid())
-            return;
+    auto createContextMenu = [](QTreeView* view, auto* model, int role, auto&& addEntries) {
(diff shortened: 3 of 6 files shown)


GitMensch:
Thank you for working on this.

I did check the location pane in caller/callee tab and do see both context menu entries - but
* in one case (sourcefile deleted) there is no context menu at all, no message, no output in stderr/logging
* in all others the Disassembly entry did not do anything (also no output in stderr/logging), while "open in file" does work and positions to the right place and Disassembly on the function level (all other panes and views) works fine

Checking F6 yields in the "go to line" being shown, placing "115" in there and pressing Enter does not seem to work either.

Note: all of CTRL+F (in both panes) and F6 (only on the source) has a remarkable time (1-2 seconds) to be shown ?!?

In any case: CTRL+G is already taken, what about ALT+G or CTRL+SHIFT+G?

src/resultscallercalleepage.cpp
```diff
@@ -234,21 +233,17 @@ void ResultsCallerCalleePage::onSourceMapContextMenu(QPoint point)
...
-        emit navigateToCode(location.path, location.lineNumber, 0);
-    }
-}
+    connect(viewCallerCallee, &QAction::triggered, this,
+            [this, location] { emit navigateToCode(location.path, location.lineNumber, 0); });
+
+    auto line = index.data(SourceMapModel::FileLineRole).value<Data::FileLine>();
+    auto disassemblyAction = contextMenu.addAction(tr("Disassembly"));
+    connect(disassemblyAction, &QAction::triggered, this, [this, line] {
+        if (m_symbol.isValid())
```
milianw, line 242: use the validity to enable/disable the action in the first place - to prevent the slot from being called in situations where your code would silently not do anything

src/resultscallercalleepage.cpp
```diff
@@ -234,21 +233,17 @@ void ResultsCallerCalleePage::onSourceMapContextMenu(QPoint point)
...
+            emit jumpToSourceCode(m_symbol, line);
+    });
 
-void ResultsCallerCalleePage::onSourceMapActivated(const QModelIndex& index)
-{
-    const auto location = toSourceMapLocation(index);
-    if (location) {
-        emit navigateToCode(location.path, location.lineNumber, 0);
-    } else {
-        const auto fileLine = index.data(SourceMapModel::FileLineRole).value<Data::FileLine>();
```
milianw, line 249: this old validation code and message got lost in translation now no?

src/resultsdisassemblypage.cpp
```diff
@@ -253,6 +253,16 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
     ui->sourceCodeView->setMouseTracking(true);
     setCostHeader(ui->sourceCodeView);
 
+    ui->gotoWidget->hide();
+
+    connect(ui->gotoLineEdit, &QLineEdit::returnPressed, this, [this]() {
+        int lineNumber = ui->gotoLineEdit->text().toInt();
+        int offset = m_sourceCodeModel->index(0, SourceCodeModel::SourceCodeLineNumber).data().value<int>();
+        // TODO: range validation, maybe, qt should handle out of range, at least test
```
milianw, line 261: please verify and then remove this todo

src/resultsdisassemblypage.cpp
```diff
@@ -275,8 +285,24 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+    layout->addWidget(label);
+    auto edit = new QLineEdit(gotoMenuWidget);
+    layout->addWidget(edit);
+    auto gotoAction = new QWidgetAction(this);
+    gotoAction->setDefaultWidget(gotoMenuWidget);
+
+    connect(edit, &QLineEdit::returnPressed, this, [this, edit] {
+        int lineNumber = edit->text().toInt();
+        int offset = m_sourceCodeModel->index(0, SourceCodeModel::SourceCodeLineNumber).data().value<int>();
+        // TODO: range validation, maybe, qt should handle out of range, at least test
```
milianw, line 300: dito

src/resultscallercalleepage.cpp
```diff
@@ -93,9 +93,9 @@ ResultsCallerCalleePage::ResultsCallerCalleePage(FilterAndZoomStack* filterStack
     ResultsUtil::setupHeaderView(ui->callerCalleeTableView, contextMenu);
     ResultsUtil::setupCostDelegate(m_callerCalleeCostModel, ui->callerCalleeTableView);
     connect(ui->callerCalleeTableView, &QTreeView::activated, this, [this](const QModelIndex& index) {
-        auto symbol = index.data(CallerCalleeModel::SymbolRole).value<Data::Symbol>();
-        if (symbol.isValid())
-            emit jumpToDisassembly(symbol);
+        m_symbol = index.data(CallerCalleeModel::SymbolRole).value<Data::Symbol>();
```
milianw, line 96: `activated` gets emitted on double click on my system - it's imo meaningless to set some member variable then. instead, the context menu should  check the current index and ask it for its symbol

src/resultscallercalleepage.cpp
milianw: also, it's imo not nice to have an action called "disassembly" silently not do anything if the disassembly isn't showing anything relating to the symbol at hand. that would happen with this code, or? since it only scrolls - it doesn't reload the data in the disassembly view?

src/resultsdisassemblypage.cpp
```diff
@@ -253,6 +253,16 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
     ui->sourceCodeView->setMouseTracking(true);
     setCostHeader(ui->sourceCodeView);
 
+    ui->gotoWidget->hide();
+
+    connect(ui->gotoLineEdit, &QLineEdit::returnPressed, this, [this]() {
+        int lineNumber = ui->gotoLineEdit->text().toInt();
```
milianw, line 259: here and a couple times below: const auto

src/resultsdisassemblypage.cpp
```diff
@@ -275,8 +285,24 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+    auto gotoMenuWidget = new QWidget(this);
+    auto layout = new QHBoxLayout(gotoMenuWidget);
+    auto label = new QLabel(QStringLiteral("Goto: "), gotoMenuWidget);
+    layout->addWidget(label);
+    auto edit = new QLineEdit(gotoMenuWidget);
+    layout->addWidget(edit);
+    auto gotoAction = new QWidgetAction(this);
+    gotoAction->setDefaultWidget(gotoMenuWidget);
+
+    connect(edit, &QLineEdit::returnPressed, this, [this, edit] {
```
milianw, line 297: why only on return pressed? why not on textEdited / textChanged?

src/resultsdisassemblypage.cpp
```diff
@@ -286,6 +312,10 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
         auto* openEditorAction = contextMenu.addAction(QCoreApplication::translate("Util", "Open in Editor"));
         QObject::connect(openEditorAction, &QAction::triggered, &contextMenu,
                          [this, fileLine]() { emit navigateToCode(fileLine.file, fileLine.line, -1); });
+
+        // TODO: fix focus, pressing enter will trigger both actions
```
milianw, line 316: yes please do that then

milianw (changes requested):

src/resultscallercalleepage.cpp
milianw: I think I (hopefully?) misunderstand the code. can you clarify: if I select this on a different entry than is currently shown in the disassembly view, it will select that one first and then scroll to the line afterwards?

lievenhey: ugh, there is a lot of bad naming going on `ResultsCallerCalleePage::jumpToSourceCode` calls `ResultsPage::onJumpToSourceCode` which calls `ResultsDisassemblyPage::onJumpToDisassembly` which will actually change the file. Afterwords `ResultsPage::onJumpToSourceCode` calls `ResultsDisassemblyPage::jumpToSourceLine`

lievenhey: so yeah it will change the file and then scroll to the line

src/resultscallercalleepage.cpp
lievenhey: yes, fixed it

lievenhey:
I added it also for the disassembly, but it doesn't work there (currently)

src/resultscallercalleepage.cpp
milianw: would be nice if you could clean this up eventually

src/resultscallercalleepage.h
```diff
@@ -10,6 +10,7 @@
 
 #include <QWidget>
 
+#include "data.h"
```
milianw, line 13: forward declare instead

src/resultsdisassemblypage.cpp
```diff
@@ -274,19 +274,58 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
-    connect(ui->sourceCodeView, &QTreeView::customContextMenuRequested, this, [this](QPoint point) {
-        const auto index = ui->sourceCodeView->indexAt(point);
-        const auto fileLine = index.data(SourceCodeModel::FileLineRole).value<Data::FileLine>();
-        if (!fileLine.isValid())
-            return;
+    auto createContextMenu = [](QTreeView* view, auto* model, int role, auto&& addEntries) {
+        auto gotoMenuWidget = new QWidget(view);
+        auto layout = new QHBoxLayout(gotoMenuWidget);
+        layout->setContentsMargins(0, 0, 0, 0);
+        auto label = new QLabel(QCoreApplication::translate("Util", "Goto: "), gotoMenuWidget);
```
milianw, line 281: you can just use `tr` here, no?

src/resultsdisassemblypage.cpp
```diff
@@ -274,19 +274,58 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+        layout->addWidget(label);
+        auto edit = new QLineEdit(gotoMenuWidget);
+        layout->addWidget(edit);
+        auto gotoAction = new QWidgetAction(view);
+        gotoAction->setDefaultWidget(gotoMenuWidget);
+
+        auto gotoLine = [view, model, role, edit] {
+            const auto lineNumber = edit->text().toInt();
+
+            // model->index()->data is somehow not possible
```
milianw, line 291: what does this mean?

src/resultsdisassemblypage.cpp
```diff
@@ -274,19 +274,58 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+            const QModelIndex index = model->index(0, role);
```
milianw, line 292: auto

src/resultsdisassemblypage.cpp
```diff
@@ -274,19 +274,58 @@ ResultsDisassemblyPage::ResultsDisassemblyPage(CostContextMenu* costContextMenu,
...
+            contextMenu.exec(QCursor::pos());
+        });
+    };
+
+    createContextMenu(ui->sourceCodeView, m_sourceCodeModel, SourceCodeModel::SourceCodeLineNumber,
+                      [this](QMenu* menu, const QModelIndex& index) {
+                          const auto fileLine = index.data(SourceCodeModel::FileLineRole).value<Data::FileLine>();
+                          if (fileLine.isValid()) {
+                              auto* openEditorAction =
+                                  menu->addAction(QCoreApplication::translate("Util", "Open in Editor"));
```
milianw, line 320: just `tr`?

src/resultsdisassemblypage.ui
```diff
@@ -87,7 +87,7 @@
         </property>
         <property name="icon">
          <iconset theme="go-previous">
-          <normaloff>.</normaloff>.</iconset>
+          <normaloff>../../../.designer/backup</normaloff>../../../.designer/backup</iconset>
```
milianw, line 90: unrelated changes?

milianw (changes requested):

src/resultsdisassemblypage.cpp
lievenhey: If I try `model->index(..)->data()` the compiler throws an error

lievenhey: same if I replace `QModelIndex` with auto.
Error message:
```
hotspot/src/resultsdisassemblypage.cpp:293:46: error: use 'template' keyword to treat 'value' as a
 dependent template name
            const auto offset = index.data().value<int>();
                                             ^
                                             template 
1 error generated.
```

src/resultsdisassemblypage.ui
lievenhey: huh? Yes no idea what QDesigner is doing

lievenhey: maybe it is because I pressed restore changes after it crashed

Issue #575, opened 2023-12-15
[com.kdab.kddockwidgets] [error] Couldn't find dock widget frequency


JoJoDeveloping:
**Describe the bug**
Hi all, I've been trying to compile `hotspot` locally (using Arch Linux). I think I have installed all dependency (see bonus bugs below), and the app builds, but it fails to start.
Instead, there are a few error messages, and then it immediately segfaults:

```
]$ ./bin/hotspot 
kf.coreaddons: "Could not find plugin konsolepart"
[2023-12-15 17:18:19.277] [com.kdab.kddockwidgets] [error] Couldn't find dock widget name=frequency
[2023-12-15 17:18:19.277] [com.kdab.kddockwidgets] [error] Couldn't find dock widget frequency
Segmentation fault (core dumped)
```

I assume that I am missing some dependency. But which?

**To Reproduce**

1. Install Arch Linux
2. Install the dependencies
  2.1 Try to install the dependencies listed here: https://github.com/KDAB/hotspot/blob/master/HACKING.md#arch-linux
  2.2 Notice that this command is completely wrong, since the listed packages simply do not exist
  2.3 Instead install all dependencies listed here: https://aur.archlinux.org/packages/hotspot
3. `mkdir make_dir; cd make_dir; cmake ../; make`
4. `./bin/hotspot`
5. Observe the segfault

**Expected behavior**
The app should not segfault immediately. I think it is supposed to show a window.

**Version Info (please complete the following information):**

- Linux Kernel version: `6.6.2-arch1-1`
- perf version: not applicable
- hotspot version (selfcompiled): `6068053c97395f99826445b887c51269adf153a1`
- if self-compiled hotspot, what version of elfutils: elfutils 0.190-1

**Additional context**

Bonus bug 1: The steps for compiling on Arch Linux specify dependencies that do not exist. See
https://github.com/KDAB/hotspot/blob/master/HACKING.md#arch-linux

PR #594, opened 2024-01-16
add ci test for non optional build

lievenhey:
This commit adds a ci build to build hotspot without optional dependencies. This should prevents bugs like #575 (fails to build without qcustomplot).


lievenhey:
Nice, the ci found the first bug already

lievenhey:
Multi line comments are now highlighted correctly
grafik

lievenhey:
dammit, it broke the disassembler

GitMensch:
Nice work!

the PR's title may should be adjusted, or the changes split into two PRs (I don't think that's necessary as long as it is split in the commits)

src/models/highlightedtext.cpp
```diff
@@ -263,9 +305,12 @@ void HighlightedText::setText(const QStringList& text)
     }
 
     m_highlightedLines.reserve(text.size());
```
milianw, line 307: don't we need to clear this container when we setText? otherwise repeated calls will just append to it?

src/models/highlightedtext.cpp
```diff
@@ -275,9 +320,13 @@ void HighlightedText::setText(const QStringList& text)
 
 void HighlightedText::setDefinition(const KSyntaxHighlighting::Definition& definition)
 {
+#if KFSyntaxHighlighting_FOUND
```
milianw, line 323: this sounds like a separate patch?

tests/modeltests/tst_formatting.cpp
```diff
@@ -46,9 +46,8 @@ private slots:
         QTest::addColumn<QStringList>("ansiStrings");
         QTest::addColumn<QVector<QVector<QTextLayout::FormatRange>>>("formatting");
 
-        QTest::addRow("no ansi sequence")
-            << QStringList {QStringLiteral(" A  B  C  D  E ")}
-            << QVector<QVector<QTextLayout::FormatRange>> {{{0, 15, {}}}}; // only default formatting
+        QTest::addRow("no ansi sequence") << QStringList {QStringLiteral(" A  B  C  D  E ")}
```
milianw, line 49: unrelated change?

milianw (changes requested):

tests/modeltests/tst_formatting.cpp
lievenhey: yes, I wonder why `clang-format` didn't pick this up

lievenhey:
I split the ci change off into its own pr

lievenhey:
It has. Until this patch the Highlighter was called on demand. Now it is called once for the whole file.

GitMensch:
I guess that's good to go after solving the merge conflict.

GitMensch:
Just wondering: Is there anything blocking the pull here?

lievenhey:
Milians high workload.

milianw (approved):
can you also add a test for this please? otherwise feel free to merge as-is for now already

Issue #577, opened 2023-12-20
Disassembly displays old source (caching issue?) and search is broken


GitMensch:
**Describe the bug**
Use of Disassembly shows _partially_ (and not always) the source of a previous disassembly.
Furthermore the search function seems to be totally broken.

@lievenhey I guess you can reproduce and want to have a look at this. Search definitely worked in older appimages and "caching atifacts" also seem to be a new thing.

milianw:
how do you trigger this? can you give more detailed information please?

GitMensch:
* open disassembly of a function in source1
* open disassembly of a function in source2

result: you mostly see the source of source1, instead of source2, but the title is set to and the syntax language is used according to source2

lievenhey:
Just to make sure, you are using the correct version of your executable? Disassembler works with addresses not and dwarf information. In theory a compiler can generate two different binaries for the same source code. Depending if he reorders the binary or not (used in obfuscated code to make leaked debug infos unusable).

GitMensch:
Sometimes the binaries are out of date, often just recompiled. If there are differences then the counters are off by some lines (or plain wrong when the code around it changed).

But this is about totally showing the wrong source, and this is "new" and happens since the last big Disassembly overhaul.

milianw:
also: for the situation you describe @lievenhey we should ideally catch it and noisily show a warning to the user. we can detect it by comparing buildids.

GitMensch:
I guess you cannot reproduce that with the appimage?
Is that related to `QT_QPA_PLATFORM` as in #339?

Do we need something like an explicit update call to also change the content of the lines?

lievenhey:
Ah, found it (actually @milianw but in another context). In `HighlightedText::setText` the `QVector` is reset correctly. This causes the weird behaviour you are seeing.

Issue #523, opened 2023-10-11
Show inclusive costs in Assembly/Disassembly View


GitMensch:
**Is your feature request related to a problem? Please describe.**
I'm looking at disassembly that is generated from COBOL (so COBOL on the left side) where one COBOL statement may yield in multiple C functions being called.
The Source View already have "cycles (self)" [= local code] and "cycles (incl.)", but that's only one inclusive cost for the complete line (which is totally right there).
The Disassembly View shows where the local code actually spends its time and it shows all called C functions - but their actual cost cannot be seen as the `callq` instructions commonly have no or nearly no cost attached and "behind that" is where the "cycles (incl.)" from the left actually are (but again one cannot see in which of the called functions).

**Describe the solution you'd like**
Preferably have the cost of the called functions added to the cycles for callq, otherwise renamed the current "cycles" column to "cycles (self)" and add "cycles (incl.)".

**Describe alternatives you've considered**
Check the called functions on their own in the caller/callee tab and see if the function currently under test is listed, but then the there's only a guess to the hint "is this relevant for that code path", while it would be directly visible in the disassembly view otherwise.

**Additional context**
Here's a _very_ simple recording showing this.
![hotspot](https://github.com/KDAB/hotspot/assets/6699539/de449ca5-a9e2-46b2-9e53-0b781259e023)
Hotspot shows the sampling to have 11% on line 9 but the assembly has no cyles, because apart from simple single-instructions it is all done in a called function. But there's no way to see in which function.
Compare with kcachegrind where each call has the amount of instructions per call and one can see that the third call to `cob_append_string` is the expensive one
![kcachegrind disassembly](https://github.com/KDAB/hotspot/assets/6699539/5889b83f-caa1-4e98-b0c7-a9ee0b3e51e6)

milianw:
To answer the original question: Yes that should be possible, the perfparser gives us the exact address for the call, we simply don't use that anywhere else and thus "lose" that data (the caller/callee view aggregates by source file/line after all). We would have to keep that data around for this view here, then we could show the cost there as you request.

I also agree that it would be very useful to have!

milianw:
@lievenhey if possible, could you also have a look at how kcachegrind fills its disassembly view? The screenshot above doesn't look too bad - certainly better than what we have when it comes to jumps and the like. I also like that it splits up more of the content into columns one could individually hide (e.g. the hex stuff wouldn't be too interesting for me).

GitMensch:
[...]

PR #595, opened 2024-01-16
fix source code search

lievenhey:
The search functions only works on a limited range, but the model contained the complete source code. This patch runs the search function on only a subset of that code.
fixes: #577

### DIFF
--- a/src/models/sourcecodemodel.cpp
+++ b/src/models/sourcecodemodel.cpp
@@ -254,8 +254,8 @@ void SourceCodeModel::find(const QString& search, Direction direction, int curre
     auto searchFunc = [&search](const QString& line) { return line.indexOf(search, 0, Qt::CaseInsensitive) != -1; };
 
     auto endReached = [this] { emit searchEndReached(); };
-
-    const int resultIndex = ::search(m_lines.mid(m_startLine, m_numLines), current, direction, searchFunc, endReached);
+    const int resultIndex = ::search(m_lines.cbegin() + m_startLine, m_lines.cbegin() + m_startLine + m_numLines,
+                                     current, direction, searchFunc, endReached);
 
     if (resultIndex >= 0) {
         emit resultFound(createIndex(resultIndex + 1, SourceCodeColumn));
--- a/src/models/sourcecodemodel.cpp
+++ b/src/models/sourcecodemodel.cpp
@@ -255,7 +255,7 @@ void SourceCodeModel::find(const QString& search, Direction direction, int curre
 
     auto endReached = [this] { emit searchEndReached(); };
 
-    const int resultIndex = ::search(m_lines, current, direction, searchFunc, endReached);
+    const int resultIndex = ::search(m_lines.mid(m_startLine, m_numLines), current, direction, searchFunc, endReached);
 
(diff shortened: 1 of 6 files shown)


GitMensch:
Is there any option to adjust the search model _test_ for that?

src/models/sourcecodemodel.cpp
```diff
@@ -254,8 +254,8 @@ void SourceCodeModel::find(const QString& search, Direction direction, int curre
     auto searchFunc = [&search](const QString& line) { return line.indexOf(search, 0, Qt::CaseInsensitive) != -1; };
 
     auto endReached = [this] { emit searchEndReached(); };
-
-    const int resultIndex = ::search(m_lines, current, direction, searchFunc, endReached);
+    qDebug() << m_lines.mid(m_startLine, m_numLines);
```
milianw, line 257: remove

src/models/sourcecodemodel.cpp
```diff
@@ -254,8 +254,8 @@ void SourceCodeModel::find(const QString& search, Direction direction, int curre
...
+    const int resultIndex = ::search(m_lines.mid(m_startLine, m_numLines), current, direction, searchFunc, endReached);
```
milianw, line 258: this requires a copy, which is unfortunate... can we pass iterators instead maybe to prevent that?

if only we could use `std::span` already...

milianw (changes requested):
I would also appreciate a better test coverage to handle this ideally

src/models/sourcecodemodel.cpp
lievenhey: `midRef`? Otherwise update to c++20?

lievenhey: Ah, `midRef` doesn't exists for Lists. And cant :facepalm:

lievenhey:
I rewrote the search functions to use iterators instead + added some tests

GitMensch:
Looking forward to get this merged.

tests/modeltests/tst_search.cpp
```diff
@@ -100,9 +85,21 @@ private slots:
...
-                     testArray, 1, Direction::Forward, [](int i) { return i == 0; }, [] {}),
+                     testArray.cbegin(), testArray.cend(), 1, Direction::Forward, [](int i) { return i == 0; }, [] {}),
                  0);
     }
+
+    void testSearchOnIterators()
+    {
+        const std::array<int, 5> testArray = {0, 1, 2, 3, 0};
+
+        for (int i = 1; i < 3; i++) {
```
milianw, line 96: make this go from 0 to testArray.size() please

milianw (approved):
one minor nit then this can go in

thank you!

GitMensch:
I've pot some time into testing the appimage - the source handling in the Disassembly window now seems to work fine.

The only thing that feels strange is the go to function:

* seems to be only usable with the context menu
* CTRL+G is identical to ALT+G and selects/closes the Flame Graph; I guess that's the QT default we want to keep *but maybe CTRL+SHIFT+G can then open the go to dialogue as well

And of course #523 would be very helpful.

Issue #580, opened 2023-12-22
copyright year needs an adjustment


GitMensch:
the license file has

> Copyright (C) 2016-2020 Klaralvdalens Datakonsult AB

Side note: Would you consider "updating" the license to GPLv3+?

milianw:
why do you want to update the license?

GitMensch:
The copyright year change in the was last done in 6f4dc6a0904c239c8844e2325a1668b2d4aabcb9.

I think that one either considers to specify the copyright for the license, in which case the current entry makes no sense, or to specify not also the license but also the copyright for the software in a single place, in which case the years should be updated.

@winterz just adjusted the license in https://github.com/KDAB/KDDockWidgets/blob/2.0/LICENSE.GPL.txt to explicit allow _both_ license versions to be used (sadly it does not include "or any later version", but that's up to the copyright owner, of course).

For a general note about possible reasons to upgrade see https://www.gnu.org/licenses/rms-why-gplv3.html (for many the reason "I can link to GPLv3" will be the most relevant reason).

milianw:
I believe the year is pretty meaningless - at least outside the US. But my colleague Allen (@winterz) regularly updates the year, which is fine by me.

Generally we try to use the same licenses for all our FOSS products here at kdab, are you saying KDDW is incompatible with hotspot? I doubt that, and if need be we can change the two licenses.

But I definitely do not intend to change the license "just because".

PR #596, opened 2024-01-16
Remove the date range from KDAB copyrights

winterz:
Keep the file creationYear as part of the copyright but no longer worry about extending the date range to include the currentYear.

This standardizes to KDAB copyright policy.


GitMensch:
Thank you or this work. Standard is always useful!
#580 should be referenced in this issue, for it to be closed on merge.

Just as a question: What do you think of dual-licensing Hotspot under GPLv2 / GPLv3 (like KDDockWidgets) or license under "GPL version 2 or later" (the license text may be unchanged then), which would allow both license versions for future use (depending on the dependencies, of course)?

milianw (approved):

PR #599, opened 2024-01-22
fix wrong source code shown

lievenhey:
The vectors containing the source code are not reset correctly. This results one some old code being shown. This patch resets them correctly.

fixes: #577

### DIFF
--- a/src/models/highlightedtext.cpp
+++ b/src/models/highlightedtext.cpp
@@ -186,6 +186,7 @@ private:
 class HighlightedLine
 {
 public:
+    HighlightedLine() = default;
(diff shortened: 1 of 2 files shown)


milianw:
I'm going to pick a different approach

GitMensch:
At least with the new approach-the problem persists.
I can now reproduce: disassembly shown, then showing a different disassembly - the code pane was adjusted to show other code, but the code lines shown mismatches the cycles and the disassembly.
If I restart hotspot and show the disassembly of that other function "directly" then there's a perfect match.

I suggest to reopen the originating issue....

Issue #485, opened 2023-05-05
AppImage does not start on non-glibc (gnu_dev_makedev: symbol not found) and needs local-installed libfuse2


GitMensch:
**Describe the bug**
Starting the run-everywhere AppImage doesn't work.

**To Reproduce**
Try to start that on any system that has no libfuse2 (still allows to run extracted) or does not use glibc (does not start at all, error message shown in ldd).

**Expected behavior**
Should open.

**Additional context**
If I understood https://github.com/AppImage/AppImageKit/issues/877 correctly then the AppImage needs to be built differently using `appimagetool --runtime-file` with the type2-runtime.

GitMensch:
According to https://github.com/AppImage/AppImageKit/wiki/FUSE:
> Some AppImages require FUSE version 2 to run. Newer AppImages have FUSE version 3 built in. 

Can we have a newer AppImage, please? Both the last release and continuous can't run on recent RHEL or Debian testing - as fuse2 was removed there.

That seems to be related to pkg2appimagebuilds .., or is it "just" necessary to build/use https://github.com/AppImage/AppImageKit ?

iamsergio:
> According to https://github.com/AppImage/AppImageKit/wiki/FUSE:
> 
> > Some AppImages require FUSE version 2 to run. Newer AppImages have FUSE version 3 built in.
> 
> Can we have a newer AppImage, please? Both the last release and continuous can't run on recent RHEL or Debian testing - as fuse2 was removed there.
> 
> That seems to be related to pkg2appimagebuilds .., or is it "just" necessary to build/use https://github.com/AppImage/AppImageKit ?

Could fuse2 be a red herring ? At least on my Archlinux I see `fusermount3` being invoked

GitMensch:
debian testing only has libfuse3 (previous versions have oldstable/libfuse2) - and after the install I still get

~~~
Cannot mount AppImage, please check your FUSE setup.
You might still be able to extract the contents of this AppImage
if you run it with the --appimage-extract option.
See https://github.com/AppImage/AppImageKit/wiki/FUSE
for more information
open dir error: No such file or directory
~~~

PR #600, opened 2024-01-23
build appimage with static runtime

lievenhey:
This allows the user to run the appimage on non fuse2 systems (all moderns systems use fuse3). This is archived by using the static appimage runtime. Since linuxdeploy doesn't support changing the runtime I only use it to create the appdir directory. Then I use the original appimagetool to package the appimage using the static runtime.

This should also make it possible to run the appimage on a non glibc system but I don't have access to one so I can't test it.

fixes: #485


lievenhey:
@GitMensch since you seem to have access to a non glibc system (unlike me) can you test if the appimage works there? I can confirm it works on ubuntu:23.04 which uses fuse3.

scripts/appimage/Dockerfile
```diff
@@ -127,6 +127,12 @@ RUN cd /opt && git clone --recursive https://github.com/KDAB/KDDockWidgets.git -
     cd KDDockWidgets && mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX=/usr -DKDDockWidgets_EXAMPLES=0 -DKDDockWidgets_FRONTENDS="qtwidgets" .. && \
     make -j && make install && cd /opt && rm -Rf KDDockWidgets
 
+RUN wget https://github.com/AppImage/type2-runtime/releases/download/continuous/runtime-x86_64 && \
+    wget https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage && \
+    chmod +x appimagetool-x86_64.AppImage && mv appimagetool-x86_64.AppImage /usr/bin/ && \
+    mv runtime-x86_64 /opt/ && \
+    yum install -y desktop-file-utils
```
milianw, line 134: move up to where we install other deps with yum already

scripts/appimage/Dockerfile
```diff
@@ -127,6 +127,12 @@ RUN cd /opt && git clone --recursive https://github.com/KDAB/KDDockWidgets.git -
     cd KDDockWidgets && mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX=/usr -DKDDockWidgets_EXAMPLES=0 -DKDDockWidgets_FRONTENDS="qtwidgets" .. && \
     make -j && make install && cd /opt && rm -Rf KDDockWidgets
 
+RUN wget https://github.com/AppImage/type2-runtime/releases/download/continuous/runtime-x86_64 && \
```
milianw, line 130: move up to the "appimage build tools" section

scripts/appimage/build_appimage.sh
```diff
@@ -74,4 +74,8 @@ linuxdeploy-x86_64.AppImage --appdir appdir --plugin qt \
     -d "./appdir/usr/share/applications/com.kdab.hotspot.desktop" \
     --output appimage
 
+# package appdir with type 2 runtime so we don't depend on glibc and fuse2
+appimagetool-x86_64.AppImage --runtime-file /opt/runtime-x86_64 appdir
+
+mv Hotspot-x86_64.AppImage "/output/hotspot-static-$gitversion-x86_64.AppImage"
 mv Hotspot*x86_64.AppImage "/output/hotspot-$gitversion-x86_64.AppImage"
```
milianw, line 81: why do we need to retain this image? isn't it sufficient to only keep the static version?

milianw (changes requested):

scripts/appimage/build_appimage.sh
lievenhey: it was just an idea in case the new one creates problems

lievenhey:
@milianw I applied your changes. I will push an update once I build the docker container.

GitMensch:
Rechecked old appimage on Alpine:

```
Alpine:~$ ./hotspot-v1.4.1-294-g7474981-x86_64.AppImage
-ash: ./hotspot-v1.4.1-294-g7474981-x86_64.AppImage: not found
Alpine:~$ ldd ./hotspot-v1.4.1-294-g7474981-x86_64.AppImage
        /lib64/ld-linux-x86-64.so.2 (0x7fdddd997000)
        libdl.so.2 => /lib64/ld-linux-x86-64.so.2 (0x7fdddd997000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7fdddd997000)
        libz.so.1 => /lib/libz.so.1 (0x7fdddd97d000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7fdddd997000)
Error relocating ./hotspot-v1.4.1-294-g7474981-x86_64.AppImage: gnu_dev_makedev: symbol not found
```

new one:

```
Alpine:~$ ./hotspot-v1.4.1-295-gcd4518d-x86_64.AppImage
fuse: failed to exec fusermount: Permission denied

Cannot mount AppImage, please check your FUSE setup.
You might still be able to extract the contents of this AppImage
if you run it with the --appimage-extract option.
See https://github.com/AppImage/AppImageKit/wiki/FUSE
for more information
open dir error: No such file or directory
```

Note: I do get the same error message on Debian, while the old one worked :-(
Seems like the first thing is to get the appimage to run again on Debian, before we check back on Alpine.

```
Alpine:~$ file ./hotspot-*
./hotspot-v1.4.1-295-gcd4518d-x86_64.AppImage: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
./hotspot-v1.4.1-294-g7474981-x86_64.AppImage: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=30e06184968532b6a9aa36f44ada39e4af0bda56, for GNU/Linux 2.6.32, stripped
```

One of the nice things: the new one is around 12% smaller than the old one.

GitMensch:
note: the Debian one has libfuse2 installed, the Alpine environment no libfuse, seems the static part of this is missing.

lievenhey:
Test results so far:
Arch Linux: works
Debian 11: works
Debian 12: works
Ubuntu 22.10: works
Ubuntu 23.10: works
CentOS 7: works
Alpine 3.19: no, missing symbols

all cleans installs (except Arch Linux CentOS 7 and Alpine where I installed kde)

GitMensch:
Do you have any idea why the _previous_ appimage works fine on my Debian 11.8, but the one from this one fails with a fuse error (and my Alpine experiencing the exact same thing)?

lievenhey:
Alpine has lots of unresolved symbols, debian no idea

GitMensch:
Until we find out why that appimage does not run on environments it previously did, I suggest to mark this PR as draft.

scripts/appimage/build_appimage.sh
```diff
@@ -74,4 +74,7 @@ linuxdeploy-x86_64.AppImage --appdir appdir --plugin qt \
     -d "./appdir/usr/share/applications/com.kdab.hotspot.desktop" \
     --output appimage
 
-mv Hotspot*x86_64.AppImage "/output/hotspot-$gitversion-x86_64.AppImage"
+# package appdir with type 2 runtime so we don't depend on glibc and fuse2
+appimagetool-x86_64.AppImage --runtime-file /opt/runtime-x86_64 appdir
```
milianw, line 78: I don't quite get it - isn't the `linuxdeploy --output appimage` command above already creating an AppImage? I think that should be changed, no? We should only create on AppImage. AFAIR `linuxdeploy` can just create the AppDir (maybe just remove the `--output appimage`?) and then we can run appimagetool here next

scripts/appimage/Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN sed -i 's#enabled=1#enabled=0#' /etc/yum/pluginconf.d/fastestmirror.conf &&
         # kgraphviewer
         boost boost-devel graphviz-devel \
         # appimages
-        fuse fuse-libs bzip2 && \
+        fuse fuse-libs bzip2 desktop-file-utils && \
```
milianw, line 46: out of interest: what is this required for?

milianw (changes requested):

GitMensch:
Rechecked this PRs appimage on updated Debian 11.9: still getting

```
/tmp/hotspot-v1.4.1-295-g2a27d2a-x86_64.AppImage
fuse: failed to exec fusermount: Permission denied

Cannot mount AppImage, please check your FUSE setup.
You might still be able to extract the contents of this AppImage
if you run it with the --appimage-extract option.
See https://github.com/AppImage/AppImageKit/wiki/FUSE
for more information
open dir error: No such file or directory
```

while the appimage outside of this PR works fine.

I really like to have the appimage be more portable, including to non-glibc Linux environments, but until this is solved we liely should mark it as draft.

PR #601, opened 2024-01-29
Elfutils debug

### LINKED ISSUE #560: perfparser: add elfutlis version info to `--version` (and its debug symbols in the debuginfo appimage package)
I'm inspecting issues seen in perfparser and find it hard to go on with the appimage provided version.

One of the reasons is that appimage creation uses centos7 (I _guess_ it is still at "debuginfod only for centos8+9 state) and more important: the elfutils used, which make the biggest part of perfutils (`perf record` shows 98.5% spent in hotspot-perfparser is in `PerfRecordSample::PerfRecordSample(PerfRecordSample&&)`), is statically linked, so not well visible with `ldd` either.

Include elfutils version in an extra line in `hotspot-perfparser` - both to make it more clear what is used and to give credits to the library that seems to do most of the actual work.
Furthermore it would be very good if the appimage debuginfo download could include the elfutils debuginfo package.

None.

lievenhey:
This will finally close #560. I successfully tested it with a local running `debuginfod` instance. I didn't managed to get it running by using the ci bundle, because the `run_debuginfod_in_docker.sh` didn't like my `debuginfo.zip` file.

### DIFF
--- a/scripts/appimage/Dockerfile
+++ b/scripts/appimage/Dockerfile
@@ -7,7 +7,7 @@
 FROM centos/devtoolset-7-toolchain-centos7 as intermediate
 
 ARG QT_VERSION=5.15.11
-ARG KF5_VERSION v5.110.0
+ARG KF5_VERSION=v5.110.0
 ARG KDDockWidgets_VERSION=2.0
 ARG QCustomPlot_VERSION=2.1.1
 ARG rust_demangler_VERSION=0.1.23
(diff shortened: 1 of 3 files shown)


lievenhey:
Adding the debug info blows up the image size to 8GB

GitMensch:
> Adding the debug info blows up the image size to 8GB. Since I haven't pushed the image, the build fails.

That would be a quite stable download, no?
How big is the current one?

That Is this the complete debuginfo for QT and elfutils, isn't it? I _guess_ that, as QT is not compiled explicit, the debuginfo could be downloaded "somewhere", but as the elfutils is build on the container we'd need to also provide its debuginfo?

lievenhey:
On the image in the ci, which will make the ci super slow since it needs to download the hole image every time it runs.

GitMensch:
Isn't the distributed QT not enough? If it is then QT can be installed from the system package manager and the link to the matching download of debug infos can just be shared.

If the system-provided QT (and/or rest) isn't usable it should be possible to:
* configure and build QT (and everything else) with full debug info (that's just the creation of the image)
* before the creation is completed: either split the debuginfo "clean" and upload that (=once during container creation) or go easy and upload the 8GB result somewhere, then `strip`` the binaries and delete the sources completely -> the final container is small (=starts fast)

milianw:
can we instead output a separate `debuginfo.zip` for the base image and host that somewhere on github? then people interested in that could download it. but we wouldn't have to pay the price on every CI run

GitMensch:
Note: it should be possible to have a CI workflow building this, which is *only manually triggered*.
This would allow to do that whenever the QT/elfutils version is changed, and after the build the debuginfo.zip can then be attached/updated to the release page (making it usable without logging in).

lievenhey:
I think I can create a workflow which is only run manually to create the image and upload it to github. This way the workflow could contain the debuginfo as an artifact.

GitMensch:
@lievenhey Looks like a perfect candidate to finish after #648 :-) ... .or should that be included there?

lievenhey:
solved in #648

PR #603, opened 2024-02-06
update ci files

lievenhey:
Github deprecates node js 16 actions. This patch updates the ci to use checkout@v4 which is the new version.


lievenhey:
:facepalm: nodejs 20 does no longer support centos7, so it doesn't run on our ci

PR #616, opened 2024-04-11
HighlightedText: Check m_highlighter for nullptr

kbroulik:
Otherwise Hotspot crashes when changing the color scheme before a trace has been loaded.

---

Wanted to debug why Plasma’s color settings module was slow when changing the color and instead had Hotspot crash whenever I did.


milianw:
thank you

PR #625, opened 2024-04-28
Clear the error when changing the record type

alois31:
When changing the record type, previous errors may no longer be applicable (for example the application having been unset in the "launch application" mode is not relevant at all to the "attach to process" mode). Clear them to allow starting the recording in the new mode again (as long as that doesn't create legitimate errors again).


milianw:
ty

Issue #605, opened 2024-02-22
Latest AppImage download without GitHub account


HybridDog:
**Describe the bug**

The Readme explains a way to download the latest AppImage from the build jobs, but it does not work unless the reader is registered and logged in at GitHub, which is not mentioned.

> Head over to our list of AppImage build jobs. When you click on a job, you'll see a page with an "Artifacts" section that contains an "appimage" binary you can then download.

**To Reproduce**

1. Log out from GitHub
2. Go to the [mentioned link](https://github.com/KDAB/hotspot/actions/workflows/build-appimage.yml?query=branch%3Amaster)
3. Click on the first "workflow run result"
4. Scroll down to "Artifacts" and click on "appimage" next to the wireframe cube icon
5. Observe how "appimage" is not a link but only text which can be marked

**Expected behavior**
The Readme could mention that downloading the build artifact requires a GitHub account or explain another way to download it.

**Screenshots**
![2024-02-22-110536_990x373_scrot](https://github.com/KDAB/hotspot/assets/3192173/27f831e9-6c43-4dd9-abef-6f744f7ce191)

**Version Info (please complete the following information):**

- Linux Kernel version: not applicable
- perf version: not applicable
- hotspot version (appimage? selfcompiled?): master
- if self-compiled hotspot, what version of elfutils: not applicable

**Additional context**
Downloading the older AppImage from the releases webpage works without GitHub account.

GitMensch:
There is no reasonable way to provide the Appimage _on each CI build **outside** of the CI artifacts_ (which is needed for the download) - but it would definitely be useful to update the CI image on the release page from time to time, kind of "stable appimage" (which may be a new release).

As I personally see some issues [already worked on] that break the Disassembly view, I'd suggest to do that afterwards.

Another option is to run a nightly/weekly workflow to update the release [as pointed out by @Kagami](https://github.com/actions/upload-artifact/issues/51#issuecomment-1893927846)

> ```yaml
> - name: Release nightly
>   uses: softprops/action-gh-release@v1
>   with:
>     prerelease: true
>     name: nightly
>     tag_name: nightly
>     files: file.txt
>     fail_on_unmatched_files: true
> ```
> 
> And instruct your users to download from `https://github.com/OWNER/REPO/releases/download/nightly/file.txt` Example [here](https://github.com/Kagami/ninnin/blob/0aa2e0638b28315da9bd44e3804d5c8ab4fbd08b/.github/workflows/build.yml).
> 
[...]

PR #627, opened 2024-04-29
Automate release

lievenhey:
This PR adds an automatic latest release aswell as release-please which provides a pr that can be merged to create a new release.

closes: #605

### DIFF
--- a/.github/workflows/release-appimage.yml
+++ b/.github/workflows/release-appimage.yml
@@ -36,3 +36,9 @@ jobs:
         artifacts: |
           ${{ github.workspace }}/hotspot-debuginfo-*
           ${{ github.workspace }}/*.AppImage
+
+    - name: Create release PR
+      uses: google-github-actions/release-please-action@v4
+      with:
+        token: ${{ github.token }}
+        release-type: simple
--- a/.github/workflows/release-appimage.yml
+++ b/.github/workflows/release-appimage.yml
@@ -1,0 +1,38 @@
+name: Create latest release
+
+on:
+  push:
+    branches:
+    - 'master'
+
+jobs:
+  latest-release:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-appimage.yml
```diff
@@ -0,0 +1,38 @@
...
+      with:
+        fetch-depth: 0
+        submodules: recursive
+
+    - name: Build
+      uses: docker://ghcr.io/kdab/kdesrc-build:latest
+
+    - name: Delete latest release
+      run: |
+        gh release delete latest --cleanup-tag || true
```
milianw, line 24: is "latest" the name here? can we call it "continuous" to revive the existing old continuous release from back then?

lievenhey: yes that is possible

milianw (approved):
lgtm, thank you

GitMensch:
I currently fail to understand that - the idea is that for each commit on master there is a nightly release, no?
Then I neither see where this defined nor where the nightly release is that would solve the now closed #605.

lievenhey:
Done, seems like ether `ncipollo/release-action` or github automatically add a `draft` tag. Now I use the github cli to remove that tag.

I don't think changing the name to `hotspot-continuous.AppImage` is a good idea. The current name is a perfect description of the commit that was used to build the appimage. If we get a bug report we can then use the name to identify which exact version was used. Maybe the bug is already fixed but the user didn't update, etc ...

GitMensch:
`./hotspot... --version` has the output, as well as the version info in the GUI.
The idea here is to get a _stable_ download URL, and for that we need - additional to the updated release - an identical name.

If you still want the versioned name: what do you think about having the appimage twice (once with a stable name)?

PR #631, opened 2024-05-06
Fix flamegraph

lievenhey:


Waqar144:
can confirm this patch fixes the linked issue

milianw:
thanks, nice find

PR #638, opened 2024-05-15
ci: rename artifacts before uploading in continuous build

lievenhey:
This creates a stable download url where the newest version can always be downloaded.

@GitMensch there you go


GitMensch:
Thank you; this has the file with the hash in the CI artifact (which is fine) and should have the stable name in the continuous.
... that reminds me that we should adjust the README, I'm going  to create a PR for that now.

milianw (approved):

PR #639, opened 2024-05-15
docs: update, mostly for AppImage notes

GitMensch:
CONTRIBUTING.md:
* add license information
* add conventional commits convention

README.md:
* updated AppImage notes - point to latest/continuous release instead of CI artifacts
* case adjustments for Hotspot and AppImage
* http -> https
* some minor rewording

HACKING.md:
* drop the unzip step as it is not needed any more
* replace shown version string by a clear pseudo-version
* update shown dates to a more current one

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -153,22 +153,20 @@ To do that, you will need to download the debuginfo artifact that matches the Ap
 
 ### AppImage-Debugging using manual symbol loading
 
-For using this approach you need to first unpack the debug information:
+For using this approach you need to first unpack the debug information
+downloaded from the release:
 
 ```bash
-$ unzip debuginfo.zip
-Archive:  debuginfo.zip
-  inflating: hotspot-debuginfo-v1.4.0-43-g02fd82c-x86_64.tar.bz2
 $ tar -xvf hotspot-debuginfo-*.tar.bz2
-hotspot-debuginfo-v1.4.0-43-g02fd82c/usr/bin/hotspot
-hotspot-debuginfo-v1.4.0-43-g02fd82c/usr/lib64/libexec/hotspot-perfparser
+hotspot-debuginfo-v1.2.3-45-g02fd82c/usr/bin/hotspot
+hotspot-debuginfo-v1.2.3-45-g02fd82c/usr/lib64/libexec/hotspot-perfparser
 ```
 
 Starting the debugger is possible directly with executing the AppImage as follows:
 
 ```bash
-$ ./hotspot-v1.4.0-43-g02fd82c-x86_64.AppImage &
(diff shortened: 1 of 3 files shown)


lievenhey:
we are now using conventional commits. Please prefix your commit message with `chore: `

GitMensch:
Wouldn't docs be more appropriate?

GitMensch:
Added a second commit for the conventional commits (I've totally missed that) in CONTRIBUTING.md and also updated the HACKING file. Note that Markdownlint is wrong - dropping the double whitespace will lead to the text being all on one line (the formatting rule was always two spaces and a line break means a line break). Should we still pull this or use "one long rendered line" or "adding a paraghraph for each sentence?

HACKING.md
```diff
@@ -153,22 +153,20 @@ To do that, you will need to download the debuginfo artifact that matches the Ap
...
+downloaded from the release:
 
 ```bash
-$ unzip debuginfo.zip
-Archive:  debuginfo.zip
-  inflating: hotspot-debuginfo-v1.4.0-43-g02fd82c-x86_64.tar.bz2
 $ tar -xvf hotspot-debuginfo-*.tar.bz2
-hotspot-debuginfo-v1.4.0-43-g02fd82c/usr/bin/hotspot
-hotspot-debuginfo-v1.4.0-43-g02fd82c/usr/lib64/libexec/hotspot-perfparser
+hotspot-debuginfo-v1.5.0-17-g0ed41cf/usr/bin/hotspot
```
milianw, line 161: here and below. unrelated changes?

milianw (commented):
why the changes to the version + sha? that's just an example

GitMensch:
I wanted to use a newer version and date as this makes the document look less outdated to me. If you want to I'll only update the 1.3 one to the 1.4 (to make samples use the same) and don't change the date.

Am 17. Mai 2024 09:44:09 MESZ schrieb Milian Wolff ***@***.***>:

HACKING.md
GitMensch: We could make it possibly even more obvious: leaving the sha as is, but changing the version to 1.2.3.-45 - Do you agree?
Are you fine with changing the date in the example below or should I leave this as-is?

lievenhey:
@GitMensch can you rebase this patch please. Master contains a patch that should stop `pre-commit` from complaining about that formatting.

GitMensch:
Done. To care for @milianw's comments I've also used the previous SHA in the version numbers referenced (less changes) and changed the version number itself top a clear pseudo-version so it won't be likely for any one to feel the need to update them.

milianw (approved):

PR #648, opened 2024-05-27
ci: automate debug info generation

lievenhey:
This patch updates the docker container to create a debuginfo file for qt and kde. The file is automatically published as a release when the container is pushed.


lievenhey:
If necessary I can throw some comments in the dockerfile

milianw (approved):

GitMensch:
Looks nice, can you also add the note about how to get and use them as an optional step in the HACKING file, possibly after merging #639?

PR #656, opened 2024-06-04
Fix crash on load

lievenhey:


milianw (changes requested):
please also include the backtrace from the bugreport in the commit message

generally this is pretty dirty though and the reason is that we call emit from the background tasks without checking whether the context object is still alive.

please add that info to the commit message, and clearly mark the dtor code as a HACK with a short explanation too

milianw (approved):

PR #659, opened 2024-06-07
ignore empty command line arguments

GitMensch:
Those can be passed by tools calling HotSpot and lead to "File '' does not exist." error messages.

One example "tool" is vscode - you can either start external programs with no arguments or with user-inputs (for example a , but if you have an empty user input (aka "just start hotspot") you get that error message.


GitMensch:
Hm, thinking about that; maybe it would be nice to allow opening hotspot without "perf.data", even if it exists - this could also be triggered by the empty file "".

Thoughts?

src/main.cpp
```diff
@@ -202,6 +202,10 @@ int main(int argc, char** argv)
     applyCliArgs(settings);
 
     auto files = parser.positionalArguments();
+
+    // remove empty arguments that may be added by tools calling HotSpot
+    files.removeAll(QLatin1String(""));
```
milianw, line 207: init the empty QString directly: `files.removeAll(QString())`

milianw:
> Hm, thinking about that; maybe it would be nice to allow opening hotspot without "perf.data", even if it exists - this could also be triggered by the empty file "" (or a new binary option).
> 
> Thoughts?

Up to you, if you see a need for it I'm not going to oppose a patch for that.

milianw (changes requested):

lievenhey:
@GitMensch I wouldn't prefix the commit with `chore` since that should be used to dependency updates. Use `feat` in this case.

GitMensch:
done

milianw:
thanks!

PR #660, opened 2024-06-07
feat: provide option to start with recording page

GitMensch:
This is a draft to showcase the working feature.

The main question is: Do we want a single option (as in this PR) that opens the recording page when no files are given (then we may want to output an error if the `QStringList files` is not empty); or do we want an option like "openPage" with a set of names that works in general (also for files opened, which then needs to be postponed after a possible load is done to switch from the summary page to the specified one).

... and of course I'm open to a better name than `record` or `openPage`.

after everything is clear the README will get updated as well to include the new option in the help output


milianw:
Imo, `--record` sounds too similar to `perf record`, meaning it would expect it to directly start recording whatever else is passed after it. But this would require (much) more work, with little gain - one can just copy'n'paste the record step into a script and only visualize the data afterwards after all. That said, I'm not opposed to someone working on that if there is a desire for such functionality.

For your use-case, I think we should rather approach it differently, i.e. like you say here:
>  an option like "openPage" with a set of names that works in general

PR #664, opened 2024-07-14
Use almalinux8 for the appimage generation

milianw:
Update our dependencies accordingly

### DIFF
--- a/scripts/appimage/Dockerfile
+++ b/scripts/appimage/Dockerfile
@@ -13,6 +13,7 @@ ARG QCustomPlot_VERSION=2.1.1
 ARG rust_demangler_VERSION=0.1.23
 ARG d_demangler_VERSION=0.0.2
 ARG KGraphViewer_VERSION=v2.4.3a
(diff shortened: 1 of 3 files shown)


GitMensch:
Would that mean that the resulting appimage would be unlikely to run on RHEL7  any more?
alma has elfutils 0.190 which is also nice - but I don't know if this would only be available on almalinux9.

In any case I definitely like the newer dependencies :-) and guess you'll fix the current issue for the appimage generation soon

> /github/workspace/scripts/appimage/build_appimage.sh: line 21: /opt/rh/gcc-toolset-13/enable

milianw:
RHEL7 is EOL: https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance

almalinux8 already has elfutils 0.190

milianw:
and: the build fails b/c I didn't yet push the new docker image. I'll see when I get the time, wanted to give @lievenhey some time to give this a look over (which he did already)

Issue #671, opened 2024-08-28
Disassembly: per-instruction cost is not shown for inlined code


arvidfm:
**Describe the bug**
When disassembling functions containing inlined code, the disassembly view does not show per-line/instruction costs for code inlined from other functions.

**To Reproduce**

Compile this program with `gcc -g -O2 -o fib main.c `:
```c
#include <stdio.h>

unsigned long fib(unsigned long n) {
  unsigned long a = 0, b = 1;
  for (unsigned long i = 0; i < n; i++) {
    unsigned long tmp = a;
    a = a + b;
    b = tmp;
  }
  return a;
}

int main() {
  printf("%d\n", fib(10000000000));
}
```

Then profile the binary using Hotspot and open the disassembly view for `main`.

**Expected behavior**
The disassembly view should show per-line/instruction costs even for inlined code.

**Screenshots**
The profile in `perf report`:
![image](https://github.com/user-attachments/assets/897be423-9238-49cd-9b66-b75dd041fa7d)

The same profile in Hotspot's disassembly view:

![image](https://github.com/user-attachments/assets/365477a3-16c5-4ec7-b39b-fd94eb05fe02)

If I compile with `-fno-inline` and disassemble `fib` instead, the costs show as expected:

![image](https://github.com/user-attachments/assets/8788f6db-7a4b-4c66-9251-f1cdeb22dd13)

**Version Info (please complete the following information):**

- Linux Kernel version: 6.9.10
- perf version: 6.9
- hotspot version: 1.5.1 (tested both AppImage and compiled from AUR)

**Additional context**
Originally found while profiling Go code, but it seems Go and C binaries behave the same here.

Issue #586, opened 2024-01-08
Attribute inline cost to caller


lievenhey:
From #568:

I try to explain what I've thought to understand and what I think should be the UI for that.

* "real" inline (the compiler is free to inline or not after all) means that the compiler includes the code for a function within another function
* the actual generation (and therefore the diassembly) _may_ be different in each place concerning the context (for example a compiler may recognize that specific parts can only be taken in specific cases and remove the other parts)
* `perf` should trace the execution of that as it is tracing non-inlined code

Running `objdump` shows the disassembly for the "outer" function, therefore the inlined code is part of the disassembly, possibly multiple times (=  we can "see" the specific generation for this specific place).
Code-wise we see a function call.

If we can deduce the place of that disassembly to be part of the "calling" place (where it is inlined) then we can directly attribute the costs there.

If we can't do it, then we still have the "original" place (source reference and/or function name) and on the other hand we have the costs in the function of "calling" it (this is something that can be seen in the caller/callee tab, also with this PR applied).
Ideally we'd then be able to attribute the "summed" costs for the disassembly to the calling place -but I do recognize that in this case we'd have to match that in the code, which is not possible to always get correct because of the preprocessor work (for example macros) - we shouldn't start that path.

This leaves the question: Is there a way to deduce the "calling" source reference for functions that are inlined? If we yes, then we should attribute the total costs to that place.

lievenhey:
Here is the rest of the conversation:
![grafik](https://github.com/KDAB/hotspot/assets/82457690/6f0433f0-44bb-418c-aba7-29b19c052c4f)

GitMensch:
Here's a bit more information in pictures:

the callee view shows (likely correct) costs of inline functions:
![grafik](https://github.com/KDAB/hotspot/assets/6699539/65c6ceba-e051-4c19-baa5-ad854ebed4cd)
as we found, there is no guaranteed single place for inlined functions, so "Disassemble" is not active

the location view shows the cost per location, like 
![grafik](https://github.com/KDAB/hotspot/assets/6699539/64200815-dd57-415d-98dc-3a328645da76)

We therefore can deduce from the caller/callee tab, that:
* most cost is in this inline function
* most cost is in line 418

[...]

PR #672, opened 2024-09-02
Fix cost attribution of inline frames in disassembly view

milianw:
Previously, we put the cost of an inline frame into a container indexed by the inline symbol. But during disassembly, we never get to query that data again, since we cannot disassemble an inline frame. Instead, we need to be able to query the cost for arbitrary binary offsets, independent of their originating symbol.

The patch here achieves this by lifting the OffsetLocationCostMap out of the CallerCalleeEntryMap into CallerCalleeResults, but mapped by the binary name. This way we can efficiently store and lookup the data of a given offset within a specific binary during disassembly, which allows us to show the cost for inlined code in the disassembly view.

Fixes: https://github.com/KDAB/hotspot/issues/671

### DIFF
--- a/src/models/data.h
+++ b/src/models/data.h
@@ -683,18 +683,6 @@ struct CallerCalleeEntry
         return *it;
     }
 
-    LocationCost& offset(quint64 addr, int numTypes)
-    {
-        auto it = offsetMap.find(addr);
-        if (it == offsetMap.end()) {
-            it = offsetMap.insert(addr, {numTypes});
-        } else if (it->inclusiveCost.size() < static_cast<size_t>(numTypes)) {
-            it->inclusiveCost.resize(numTypes);
-            it->selfCost.resize(numTypes);
-        }
-        return *it;
-    }
-
     ItemCost& callee(const Symbol& symbol, int numTypes)
     {
         auto it = callees.find(symbol);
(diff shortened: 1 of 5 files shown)


milianw:
@arvidfm @lievenhey  @GitMensch care to try out the appimage in here for inline cost attribution in the disassembly addr view? I think that should be fixed now

arvidfm:
@milianw Is [this](https://github.com/KDAB/hotspot/actions/runs/10662946666) the right artifact to test? I can't seem to find any artifacts for the latest commit (c3e1682), only the pre-force push one (8e20e25).

GitMensch:
@arvidfm you find the right artifact here: https://github.com/KDAB/hotspot/actions/runs/10662946666/artifacts/1881066271 (go to "build appimage, then to its summary, then click on "skipped by" until you get to actual binaries)

arvidfm:
Giving it a try on one of my projects, the inlined instructions seem to be getting annotated properly now, but the corresponding source lines on the left are still missing the costs, is that expected?
image

GitMensch:
> but the corresponding source lines on the left are still missing the costs

I _think_ this is what #586 is about.
Interesting in this area would be #523 as well - but that's an enhancement;-)

milianw:
ok cool, I'll merge this as-is for now and then look at the source file/line problem next when I find the time. the solution to that is probably going to look similar, but we need to make sure that we don't show confusing numbers when the same function is inlined into multiple ones

Issue #666, opened 2024-07-18
caller/callee: optional "extended" search


GitMensch:
**Is your feature request related to a problem? Please describe.**
Commonly I need to inspect the cost for multiple symbols.

**Describe the solution you'd like**
Allow the search to be "extended", for example when starting with a "+" then have the following word be an additional search like `+symba +BSYM` / `symba||BSYM` or even regex (possibly also triggered by a leading character that is unexpected to occur symbols like %l.

**Describe alternatives you've considered**
The current option is only to either filter the uninteresting symbols out (getting to the same result an extended/regex search would provide), or using the search multiple times and inspect each time separately.

PR #676, opened 2024-09-17
feat: support regex search

lievenhey:
Since we use QSortFilterProxyModel we get this for free. This patch adds support for regex search by not escaping the search term if it is prefix with %.
For the flamegraph there is a custom implementation, which changes the current QString::contains to a QRegularExpression::match call.

Closes: #666

### DIFF
--- a/src/flamegraph.cpp
+++ b/src/flamegraph.cpp
@@ -546,14 +546,13 @@ struct SearchResults
     qint64 directCost = 0;
 };
 
-SearchResults applySearch(FrameGraphicsItem* item, const QString& searchValue)
+SearchResults applySearch(FrameGraphicsItem* item, const QRegularExpression& expression)
 {
     SearchResults result;
-    if (searchValue.isEmpty()) {
+    if (expression.pattern().isEmpty()) {
         result.matchType = NoSearch;
-    } else if (item->symbol().symbol.contains(searchValue, Qt::CaseInsensitive)
-               || (searchValue == QLatin1String("??") && item->symbol().symbol.isEmpty())
-               || item->symbol().binary.contains(searchValue, Qt::CaseInsensitive)) {
+    } else if (expression.match(item->symbol().symbol).hasMatch() || expression.match(item->symbol().binary).hasMatch()
+               || (expression.pattern() == QLatin1String("\\?\\?") && item->symbol().symbol.isEmpty())) {
         result.directCost += item->cost();
         result.matchType = DirectMatch;
     }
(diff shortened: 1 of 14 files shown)


GitMensch:
That's a nice implementation! The tooltip also helps for documentation purposes (I'm still not sure if there should be a note on that in the README as well).

src/resultsbottomuppage.ui
```diff
@@ -67,7 +67,14 @@
       <item>
        <widget class="QLineEdit" name="bottomUpSearch">
         <property name="toolTip">
-         <string>Filter the call graph tree.</string>
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Filter the call graph tree.&lt;br/&gt;Prefix with '%' to turn it into an regex.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
GitMensch, line 70: @lievenhey you want to revert that line change here and in the resultstopdownpage.ui as the % is not used any more (and possibly add a note / updated screenshot for the README)

src/flamegraph.cpp
```diff
@@ -546,14 +546,13 @@ struct SearchResults
     qint64 directCost = 0;
 };
 
-SearchResults applySearch(FrameGraphicsItem* item, const QString& searchValue)
+SearchResults applySearch(FrameGraphicsItem* item, const QRegularExpression& expression)
 {
     SearchResults result;
-    if (searchValue.isEmpty()) {
+    if (expression.pattern().isEmpty()) {
```
lievenhey, line 552: I'm not sure if this should be `!expression.isValid() || expression.pattern().isEmpty()`

src/resultsbottomuppage.ui
GitMensch: Still included in resultstopdownpage.ui ...

src/flamegraph.cpp
GitMensch: if it isn't valid we should not get here, at least not when executed from the UI (the function could possibly `assert` that or do an early-exit for the invalid expression)

src/flamegraph.cpp
```diff
@@ -805,13 +804,26 @@ FlameGraph::FlameGraph(QWidget* parent, Qt::WindowFlags flags)
             searchInput->setMinimumWidth(200);
             layout->addWidget(searchInput);
 
+            auto regexCheckBox = new QCheckBox(widget);
+            regexCheckBox->setText(tr("Regex Search"));
```
milianw, line 808: can you show a screenshot please - this sounds pretty wide to me - can we somehow reduce  the width that this checkbox occupies I wonder?

milianw:
otherwise code lgtm, tests would obviously be nice to have, but I don't think we have any integration tests of that sort yet :(

src/flamegraph.cpp
lievenhey: I mean I could shorten the text to something like `^[A-Z]$` and then add a tooltip explaining that this enables the regex search.

GitMensch: Even for a programmer that looks strange.
That said - the version from vscode does look quite nice (inline toggle - could still be an outside checkbox) with three short "text symbols" and the expected tooltip:

vscode search

Regex Text Symbol `.*`

milianw: lgtm, thanks

PR #677, opened 2024-09-17
feat: detect more perf data file types in directory

lievenhey:
Currently `hotspot dir` only opens perf.data files. This patch extends this to perf.data, perf.data.old and perf.data.perfparser files.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -57,6 +57,20 @@ void Q_DECL_UNUSED initRCCIconTheme()
 }
 #endif
 
+namespace {
+QStringList findPerfDataFiles(const QDir& directory = QDir::current())
+{
+    QStringList files;
+
+    for (const auto& filename : {QStringLiteral("perf.data"), QStringLiteral("perf.data.perfparser")}) {
+        if (directory.exists(filename)) {
+            files.push_back(directory.absoluteFilePath(filename));
+        }
+    }
+    return files;
+}
+}
+
 std::unique_ptr<QCoreApplication> createApplication(int& argc, char* argv[])
 {
     const std::initializer_list<std::string_view> nonGUIOptions = {"--version", "-v", "--exportTo",
@@ -236,7 +250,12 @@ int main(int argc, char** argv)
     if (!files.isEmpty()) {
         auto file = files.constFirst();
         if (QFileInfo(file).isDir()) {
-            file.append(QLatin1String("/perf.data"));
+            // search for common perf data files in dir and open the first one found
+            // if none is found, hotspot will complain, that `file` is not a file
+            const auto perfDataFiles = findPerfDataFiles(QDir(file));
+            if (!perfDataFiles.isEmpty()) {
+                file = perfDataFiles.first();
+            }
         }
 
         if (parser.isSet(exportTo)) {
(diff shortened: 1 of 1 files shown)


GitMensch:
From the commit and PR title I've thought this to be about "open all perf.data/.perfparser files in the given directory"...

Not sure about perf.old.data... what do you think about moving it at least after perf.data.perfparser or take both out of that check and list dir directory for any files starting with "perf.data" or ending on ".perfparser"?

lievenhey:
right, I changed the title so it is less misleading.

> Not sure about perf.old.data... what do you think about moving it at least after perf.data.perfparser or take both out of that check and list dir directory for any files starting with "perf.data" or ending on ".perfparser"?

The intend of this feature is to emulate `perf report`. Therefor it doesn't make sense to open all files, only the newest one. But yes it should be not the second in the list. Lets see what @milianw has to say.

GitMensch:
Opening the newest one would make it unlikely to load perf.old.data - but yes, I'd find it very nice if specifying the directory just takes the newest file per our file open- re https://github.com/KDAB/hotspot/blob/e6ce4c460b2990a86dee0d2b180949a28cc71dc8/src/mainwindow.cpp#L287-L294

Those are more files and currently there's no list dir, order by timestamp - can you please inspect adding this in this PR?

src/main.cpp
```diff
@@ -236,7 +236,15 @@ int main(int argc, char** argv)
...
-            file.append(QLatin1String("/perf.data"));
+            // search for common perf data files in dir and open the first one found
+            // if none is found, hotspot will complain, that `file` is not a file
+            for (const auto& perfDataFile : {QStringLiteral("/perf.data"), QStringLiteral("/perf.data.old"),
+                                             QStringLiteral("/perf.data.perfparser")}) {
+                if (QFileInfo::exists(file + perfDataFile)) {
+                    file.append(perfDataFile);
+                    break;
+                }
+            }
```
GitMensch, line 247: Can you move that out to a helper function that returns a QString() with the filename?
This can then also be adjusted to use the newest matching file...

milianw:
I agree with Simon, remove the `.old` and extract that into a single function that allows us to share that code

lievenhey:
done

lievenhey:
I just noticed, that this should not be merged, until #678 is merged. That PR makes opening compressed files much easier.
I should add compressed files to `findPerfDataFiles`

src/main.cpp
```diff
@@ -267,9 +286,12 @@ int main(int argc, char** argv)
     } else {
         // open perf.data in current CWD, if it exists
         // this brings hotspot closer to the behavior of "perf report"
-        const auto perfDataFile = QStringLiteral("perf.data");
-        if (QFile::exists(perfDataFile) && window) {
-            window->openFile(perfDataFile);
+        const auto perfDataFiles = findPerfDataFiles();
+        for (const auto& perfDataFile : perfDataFiles) {
```
GitMensch, line 290: `perf report` opens only a single file and I have _lots_ of perf files in my directories...
I suggest to add either an optional integer limit or `boolean newestFileOnly` attribute to that function and use it here and above.

GitMensch (changes requested):
Note: the commit summary still says "perf.data.old" (I still suggest to actually do as the function name ` findPerfDataFiles` - not search for a list of 2 fixed entries but to search for `perf*.data perf.data.* *.perfparser *.perfparser.*`, ordered by date (newest first).

src/main.cpp
milianw: the loop breaks after the first opened file, so this is fine as-is

milianw (approved):

src/main.cpp
GitMensch: In theory this means that the caller may collect lots of file entries, just to drop all of those later.
That's not of a big issue currently as `findPerfDataFiles()` does not match our file open window (which I think it should, I'm creating a new issue to track it).

PR #678, opened 2024-09-17
Improve compressed files

lievenhey:


lievenhey:
This is a fix for the currently broken compressed files feature. The current implementation doesn't work, because I passed the compressed file to `perfparser`. This patch series fixes that and adds support for archives files so that hotspot can now open `zip` and `tar.gz` files.

GitMensch:
Does this overlap with #633?

src/parsers/perf/perfparser.cpp
```diff
@@ -1466,10 +1466,7 @@ PerfParser::PerfParser(QObject* parent)
         m_stopRequested = false;
     });
 
-    auto parsingStopped = [this] {
-        m_isParsing = false;
-        m_decompressed.reset();
-    };
+    auto parsingStopped = [this] { m_isParsing = false; };
```
milianw, line 1474: can we get a test for this patch please?

src/parsers/perf/perfparser.cpp
```diff
@@ -1991,19 +1996,110 @@ void PerfParser::exportResults(const QUrl& url)
...
+    }
+
+    auto dir = archive->directory();
+    auto entries = dir->entries();
+
+    if (entries.size() == 1) {
+        return extractFile(dir, entries[0]);
+    }
+
+    for (const auto& file : {QStringLiteral("perf.data"), QStringLiteral("perf.data.perfparser")}) {
```
milianw, line 2052: reuse utility function you are adding in the other patch

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -825,6 +825,68 @@ private slots:
 
         QCOMPARE(decompressed.readAll(), QByteArrayLiteral("Hello World\n"));
     }
+
+    void testArchive_data()
```
milianw, line 829: squash this into the previous patch

milianw (changes requested):

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -825,6 +825,68 @@ private slots:
...
+    {
+        QTest::addColumn<QByteArray>("content");
+        QTest::addColumn<QString>("filename");
+
+        QTest::addRow("uncompressed tar") << qUncompress(QByteArray::fromBase64(QByteArrayLiteral(
```
milianw, line 834: this is really ugly and opaque stuff that I feel uncomfortable with it, can we find another way?

tests/integrationtests/tst_perfparser.cpp
```diff
@@ -825,6 +825,68 @@ private slots:
...
+        QFETCH(QByteArray, content);
+        QFETCH(QString, filename);
+
+        QTemporaryFile compressed;
+        compressed.setFileTemplate(filename);
+        compressed.open();
+        compressed.write(content);
+        compressed.close();
+
+        qDebug() << qCompress(content).toBase64();
```
milianw, line 882: remove?

milianw (approved):

src/parsers/perf/perfparser.cpp
milianw: fine then, let's get this in as-is then!

lievenhey:
Ok I found a way to make it work with one function.
That patch also refactors a lot of code. Mainly the use of `#include <../utils.h>` in the model classes. This is done by moving that code to the models.

milianw (approved):

PR #686, opened 2024-12-10
Fix memory leaks

lievenhey:


src/settingsdialog.cpp
```diff
@@ -68,6 +68,7 @@ SettingsDialog::SettingsDialog(QWidget* parent)
 #if KGraphViewerPart_FOUND
     , callgraphPage(new Ui::CallgraphSettingsPage)
 #endif
+    , m_configs(std::make_unique<MultiConfigWidget>(this))
```
milianw, line 71: why is this leaking, we pass `this` - is that not considered as parent?

src/mainwindow.cpp
```diff
@@ -96,7 +96,6 @@ MainWindow::MainWindow(QWidget* parent)
     , m_startPage(new StartPage(this))
     , m_recordPage(new RecordPage(this))
     , m_resultsPage(new ResultsPage(m_parser, this))
-    , m_settingsDialog(new SettingsDialog(this))
```
milianw, line 99: dito, why is this leaking?

lievenhey:
Seems like the leaks you where wondering about (me too) were caused by a qt/plasma bug. After updating they no longer appear. (Still get wayland leaks but I can deal with them, before it were hundreds)

GitMensch:
Should this 4-line change be pulled in or is it not relevant any more (further leaks, if there are some, could always be coped with a follow-up PR)?

PR #694, opened 2025-02-28
Fix build with Qt 6.9

heirecka:
That version added some overloads to setBrush [1], making the {} ambiguous:
"hotspot/src/models/timelinedelegate.cpp:197:26: error: call of overloaded 'setBrush(<brace-enclosed initializer list>)' is ambiguous"

[1] qtbase commit: 2ad79c626d7a94e545886152ce5ac3feabf619a2


milianw:
thanks!

Issue #691, opened 2025-01-25
Hotspot should escape Caller/Callee data to be graphed


cmack:
**Describe the bug**
While inspecting perf output from a dynamic language (Common Lisp; SBCL implementation) the Callee/Caller tab will not show the call graph for certain items when clicked. 

I used graphviz's dot command to  check the tmp file and found it also could not parse some items because the double-quotes were not escaped. I *believe* this is the main problem preventing display.

**To Reproduce**

1. Use SBCL and its sb-perf package to create the debug symbols then start perf recording normally. (I don't expect hotpot devs to do this unless they really want to)
2. Start Hotspot and then click on the Callee/Caller tab
3. Click on Callee or Caller items until you have a screen like the screenshot provided where the `graph /tmp/hotspot.<blah> loading...` message is seen and doesn't change even given enough time.

**Expected behavior**
For hotspot to generate a proper digraph file in the presence of symbols that have double quotes (or anything else that should be escaped) and the expected call graph can be seen.

**Screenshots**

![Image](https://github.com/user-attachments/assets/74693859-dca8-4384-813c-498e8d45fa69)

**Version Info (please complete the following information):**

- Linux Kernel version: `Linux multivec 6.12.10 #1 SMP PREEMPT_DYNAMIC 1 x86_64 GNU/Linux`
- perf version: `perf version 6.12.10-gnu`
- hotspot version (appimage? selfcompiled?): `1.5.80` -- GUIX compiled
- if self-compiled hotspot, what version of elfutils:

**Additional context**
This is the generated  tmp file.

[hotspot.GSyuJj.txt](https://github.com/user-attachments/files/18547815/hotspot.GSyuJj.txt)

lievenhey:
Hi, thanks for the report. Can you try again with https://github.com/KDAB/hotspot/pull/698 . That should fix it.

GitMensch:
Just to add to what was posted before - you may try the appimage from the CI build, link to that one: https://github.com/KDAB/hotspot/actions/runs/13784358035/artifacts/2728434083 - see https://github.com/KDAB/hotspot?tab=readme-ov-file#for-any-linux-distro-appimage for some details (note: I commonly make that executable, then run it with `--appimage-extract`, then running it from the extracted folder as that removes additional dependency and temporary mounts).
@cmack Thanks for your report back!

lievenhey:
[...]

PR #700, opened 2025-03-14
fix: escape special symbols in node labels of children, too

cmack:
## A Warning :)
C++ is not my everyday language, nor am I super familiar with the code to know if this may have unintended consequences.

## Details
Further fixes to #691 continuing the initial fix in fd9f7d10cd00832f2212b36f07da1752cbe550d5

I was able to test this locally with a perf.data file exhibiting the problem in the original issue and have not had problems seeing the graph while browsing various items in the Callee/Caller tab that previously failed.

I wasn't sure if line 88 would have been more appropriate/canonical for this fix...


milianw:
thanks, lgtm

PR #703, opened 2025-03-21
Properly include kgraphviewer_interface.h

gin-ahirsch:
The KGraphViewerPart CMake target sets the include-directory for the compiler including the kgraphviewer parent directory. Properly linking against the KGraphViewerPart target and fixing the include path is important when not using a system-wide installation of KGraphViewerPart (or when it is not installed in the default location).


milianw:
thanks!

GitMensch:
Any ETA on when kraphviewer may be part of the AppImage?

PR #707, opened 2025-04-02
fix: Always call QToolTip::showText with a widget

kbroulik:
Under Wayland, a popup must always have a parent window. Otherwise, the tooltip will show up as proper window with title bar and all and steal focus from the main window, causing the tooltip to just close again immediately.

---

Tooltip shows properly on Wayland now:

Screenshot_20250402_171910


milianw:
thanks!

Issue #705, opened 2025-03-21
RISC-V support


dendibakh:
Are there any plans to support RISC-V?

I tried hotspot on SiFive P550 and see the following:
```
// dump file collected with simple `perf record -e branches -- <my app>`
feature not properly read PerfHeader::CACHE 8 0
feature not properly read PerfHeader::SAMPLE_TIME 16 0
feature not properly read PerfHeader::BPF_BTF 4 0
1 528384 6 1 0
feature not properly read PerfHeader::BPF_PROG_INFO 4 0
feature not properly read PerfHeader::CPU_TOPOLOGY 380 348
feature not properly read PerfHeader::CPU_PMU_CAPS 4 0
Linux version "6.6.21" detected. Switching to automatic buffering.
No information about CPU architecture found. Cannot unwind.
"The hotspot-perfparser binary exited with code 6 (invalid perf data file)."
```

An estimate of how hard it would be to add the RISC-V support would also be great.
Thank you for considering.

dendibakh:
Sorry for the delay. Please find the binary and the perf.data: [data.zip](https://github.com/user-attachments/files/19675642/data.zip)

Just in case, here is some other info...

Sources are here: https://github.com/dendibakh/perf-ninja/tree/main/labs/misc/warmup

```
perf version 6.6.21
kernel version 6.6.21-9-premier
```
Collected with:
```
$ perf record -e task-clock ./lab 
2025-04-09T23:46:11+00:00
Running ./lab
Run on (4 X 1400 MHz CPU s)
Load Average: 0.03, 0.32, 0.85
-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
bench1           1015 ns         1015 ns       678512
[ perf record: Woken up 1 times to write data ]
[ perf record: Captured and wrote 0.124 MB perf.data (3218 samples) ]
```

The output of perf report:
```
Samples: 3K of event 'task-clock:u', 4000 Hz, Event count (approx.): 804500000
solution(int*, int)  /home/dendibakh/perf-ninja/labs/misc/warmup/build/lab [Percent: local period]
Percent│
       │
       │
       │      Disassembly of section .text:
       │
       │      00000000000120cc <solution(int*, int)>:
       │      solution(int*, int):
       │
       │      #include "solution.h"
       │
       │      int solution(int *arr, int N) {
       │      int res = 0;
       │      for (int i = 0; i < N; i++) {
[...]

PR #709, opened 2025-04-14
feat: support risc-v

lievenhey:
Requires https://github.com/KDAB/perfparser/pull/39 to be merged. Fixes: #705


lievenhey:
@dendibakh I added some support for risc-v. I can't really test this, but the result looks sane.

dendibakh:
Thanks @lievenhey! Is there an easy way for me to test this? IIRC, building hotspots from sources is not trivial. But if I can update a few libs and test, then I can test the fix.

lievenhey:
@dendibakh there is a ci job running which builds an appimage. You can download the current appimage here https://github.com/KDAB/hotspot/actions/runs/14440980603?pr=709

dendibakh:
Thanks @lievenhey. The appimage is x86, so I can't install it on my P550 board. However, I installed it on a Linux x86 system and opened the profile there. I can see the correct function names, which is good.
image
I can't see the disassembly for the `solution` function though -- it shows a blank window.

lievenhey:
Yeah the disassembly is a little bit funky but I am working on a better solution.

lievenhey:
We are using `objdump` for the disassembly. If you have a `objdump` version which supports riscv you can set that in Settings -> Unwind -> objdump path. Then it should work.

dendibakh:
I installed riscv binutils with `sudo apt install binutils-riscv64-unknown-elf` and added `/usr/bin/riscv64-unknown-elf-objdump` in `Settings -> Unwind -> objdump path`. Now when I try to show disassembly for a function, the application crashes.
```
/tmp/.mount_hotspoe5cYDy/AppRun.wrapped: line 6:  8792 Segmentation fault      (core dumped) LD_LIBRARY_PATH="$d/usr/lib":$LD_LIBRARY_PATH "$bin/hotspot" "$@"
```

GitMensch:
> I tried getting a stack trace but it seems the image provided by lievenhey doesn't have symbols. So, when I run it under gdb it shows no useful data.

@dendibakh Been there before - but don't be afraid, the CI run also has debug info and there's a documentation how to use it (please suggest changes if you stumble over something as this was written after doing it more than once, sometimes you leave important pieces out then...) https://github.com/KDAB/hotspot/blob/master/HACKING.md#debugging-the-appimage

PR #710, opened 2025-04-16
Tracefs location

adamtassier:
Hi,

This PR changes the file system location that is checked for `tracefs`. At [least since linux v4.1](https://www.kernel.org/doc/Documentation/trace/ftrace.txt) the default location has been `/sys/kernel/tracing/` instead of `/sys/kernel/debug/tracing/` (but the latter was kept for compatability).

I also changed the tooltip of the 'off-cpu profiling' checkbox to better reflect the conditions under which it can be checked as I had to go look into the source code.

This was an issue for me when setting up rootless 'off-cpu profiling' (ie not using 'elevate privileges'). Turns out the problem was that `/sys/kernel/debug/tracing/` was not readable on my machine but `/sys/kernel/tracing/` was (via `tracing` group).

### DIFF
--- a/src/recordhost.cpp
+++ b/src/recordhost.cpp
@@ -116,10 +116,12 @@ bool privsAlreadyElevated()
         struct stat buf;
         return stat(path, &buf) == 0 && ((buf.st_mode & 07777) & required) == required;
     };
-    static const auto paths = {"/sys/kernel/", "/sys/kernel/tracing"};
-    isElevated = std::all_of(paths.begin(), paths.end(), checkPerms);
 
-    return isElevated;
+    if (checkPerms("/sys/kernel/tracing")) {
+        return true;
+    }
+    // tracing used to be mounted on /sys/kernel/debug/ on older systems
+    return checkPerms("/sys/kernel/debug/tracing");
 }
 
 RecordHost::PerfCapabilities fetchLocalPerfCapabilities(const QString& perfPath)
--- a/src/recordhost.cpp
+++ b/src/recordhost.cpp
@@ -69,7 +69,7 @@ QByteArray perfBuildOptions(const QString& perfPath)
 
 bool canTrace(const QString& path)
 {
-    const QFileInfo info(QLatin1String("/sys/kernel/debug/tracing/") + path);
+    const QFileInfo info(QLatin1String("/sys/kernel/tracing/") + path);
     if (!info.isDir() || !info.isReadable()) {
         return false;
     }
@@ -116,7 +116,7 @@ bool privsAlreadyElevated()
         struct stat buf;
         return stat(path, &buf) == 0 && ((buf.st_mode & 07777) & required) == required;
     };
-    static const auto paths = {"/sys/kernel/debug", "/sys/kernel/debug/tracing"};
+    static const auto paths = {"/sys/kernel/", "/sys/kernel/tracing"};
     isElevated = std::all_of(paths.begin(), paths.end(), checkPerms);
 
     return isElevated;
--- a/src/recordhost.cpp
+++ b/src/recordhost.cpp
@@ -111,13 +111,9 @@ bool privsAlreadyElevated()
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/recordhost.cpp
```diff
@@ -116,7 +116,7 @@ bool privsAlreadyElevated()
         struct stat buf;
         return stat(path, &buf) == 0 && ((buf.st_mode & 07777) & required) == required;
     };
-    static const auto paths = {"/sys/kernel/debug", "/sys/kernel/debug/tracing"};
+    static const auto paths = {"/sys/kernel/", "/sys/kernel/tracing"};
```
adamtassier, line 119: Checking `/sys/kernel` is probably redundant here right?

milianw: I think so, yes - the old code was needed back then (Linux 3.x I think), because there tracing was mounted into debug, and thus needed to be checked separately

milianw: that said, I wouldn't mind if we could rework this code: 

check first if `sys/kernel/tracing` is usable, if so return early. otherwise keep backwards compat and check the two old locations?

milianw (changes requested):
ping? will you attend the review comment I gave you?

adamtassier:
@milianw see latest commit [262cd7d](https://github.com/KDAB/hotspot/pull/710/commits/262cd7d6db6db6350e675e1ce231db42b54ba76d)

I did remove the permission check on `/sys/kernel/debug/` as we dont need to directly use it and `stat` will fail anyway if anything in the path to `/sys/kernel/debug/tracing` is not executable.

src/recordhost.cpp
```diff
@@ -69,7 +69,7 @@ QByteArray perfBuildOptions(const QString& perfPath)
 
 bool canTrace(const QString& path)
 {
-    const QFileInfo info(QLatin1String("/sys/kernel/debug/tracing/") + path);
+    const QFileInfo info(QLatin1String("/sys/kernel/tracing/") + path);
```
milianw, line 72: this here would ideally retain the fallback to the old location too - otherwise we can remove the fallback from the other location - either both or none

milianw (changes requested):

src/recordpage.ui
```diff
@@ -265,7 +265,8 @@
       <item row="2" column="0">
        <widget class="QLabel" name="offCpuLabel">
         <property name="toolTip">
-         <string>Record scheduler switch events. This enables off-CPU profiling to measure sleep times etc. This requires elevated privileges.</string>
+         <string>Record scheduler switch events. This enables off-CPU profiling to measure sleep times etc.
+This requires elevated privileges or both /sys/kernel/tracing to be readable and /proc/sys/kernel/perf_event_paranoid == -1.</string>
```
milianw, line 269: considering this only talks about the new location, maybe let's drop the old fallback - you said this new location is pretty old by now anyhow, right? would this be available already on say ubuntu 22.04 lts or similar?

adamtassier: According to https://www.kernel.org/doc/Documentation/trace/ftrace.txt it has been mounted on `/sys/kernel/tracing` since linux v4.1. Which means ubuntu 15.04 (v3.19) is the last version using the old mount point.

adamtassier: That said the old one is always mounted too. Just to reiterate, my initial problem was that there was a mismatch between the permissions of these folders.

src/recordhost.cpp
adamtassier: Given that the debugfs mount point is 10 years old at this point I removed the fallback.

milianw (approved):

GitMensch:
@adamtassier can you rebase, please? That would run the checks (again?)
Otherwise I may give that a try... for finally getting it integrated.

milianw:
hey all, sorry once again for my long absence. I've rebased and merged this now. much appreciated <3

PR #719, opened 2025-04-28
Fix error msg

### LINKED ISSUE #711: When `perf` isn't in `$PATH`, the "Start Recording" button in "File" > "Record Data" is disabled. However, the reason isn't communicated to the user.
#### **Is your feature request related to a problem? Please describe.**

> When `perf` isn't in `$PATH`, the "Start Recording" button in "File" > "Record Data" is disabled. However, the reason isn't communicated to the user.

#### **Describe the solution you'd like**

Indicate to the user that the problem is that `perf` isn't in `$PATH` (usually because it's not installed).

#### **Describe alternatives you've considered**

Add `perf` as a dependency. It wouldn't make much sense as a mandatory one, but would as an optional one.

lievenhey:


milianw:
how do you ensure that the error message is cleared once a valid value is entered?

lievenhey:
Good question, with the current flow it is not possible. We would need a `isValid` check or something like that. Then we could also push all messages on a stack and show them all at once.

milianw (changes requested):
see previous comment

milianw:
@lievenhey  I believe this can be closed and will do so now. If you disagree, feel free to reopen and revive

Issue #718, opened 2025-04-27
Register as the default file association for `.perfparser` and `.data`.


RokeJulianLockhart:
#### **Is your feature request related to a problem? Please describe.**

By default, `hotspot-1.5.1-5.fc42` does not register as the default file association for `.perfparser` or `.data`:

![Image](https://github.com/user-attachments/assets/31e99f20-5b54-4ed5-954c-e5193a8bb198)

#### **Describe the solution you'd like**

Register `hotspot` as the default file association, like I can in `dolphin-25.04.0-3`:

1. ![Image](https://github.com/user-attachments/assets/82dfecb6-0c6f-4eed-b978-1159958c6533)

1. ![Image](https://github.com/user-attachments/assets/19023fdc-6aff-419b-a16c-a48702366a09)

1. ![Image](https://github.com/user-attachments/assets/7465dab5-0653-4080-a333-553910fd9138)

PR #720, opened 2025-04-28
feat: associate application/x-perf-data and application/x-hotspot wit…

lievenhey:
…h hotspot

This patch adds a custom mimetype for *.perfparser files called application/x-hotspot and registers hotspot as the default application to open it.

Fixes: #718

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,22 @@ if(NOT APPIMAGE_BUILD)
 endif()
 
 install(
+    FILES com.kdab.hotspot.xml
+    DESTINATION ${KDE_INSTALL_MIMEDIR}
+)
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -165,6 +165,22 @@ if(NOT APPIMAGE_BUILD)
     add_subdirectory(tests)
 endif()
 
+install(
+    FILES com.kdab.Hotspot.xml
```
milianw, line 169: is the uppercase `Hotspot` needed, or could this remain lowercase?

milianw (changes requested):

CMakeLists.txt
lievenhey: I tried to match the naming of the desktop file, but I looked at the appdata file.

PR #725, opened 2025-05-26
fix: don't try to use a cross-architecture objdump on the native arch…

pcc:
…itecture

If the host architecture is aarch64, aarch64-linux-gnu-objdump may not exist, which is the case on at least Fedora and NixOS, so disassembly doesn't work by default. Similar logic likely also applies to arm-linux-gnueabi-objdump. Fix the problem by only trying to use the cross-architecture objdump if we are not building for the given architecture.

### DIFF
--- a/src/resultsdisassemblypage.cpp
+++ b/src/resultsdisassemblypage.cpp
@@ -574,10 +574,18 @@ void ResultsDisassemblyPage::showDisassembly()
             return objdump;
 
         if (m_arch.startsWith(QLatin1String("armv8")) || m_arch.startsWith(QLatin1String("aarch64"))) {
-            return QStringLiteral("aarch64-linux-gnu-objdump");
+            if (auto aarch64Objdump = QStandardPaths::findExecutable(QStringLiteral("aarch64-linux-gnu-objdump"));
+                !aarch64Objdump.isEmpty())
+                return QStringLiteral("aarch64-linux-gnu-objdump");
         }
-        const auto isArm = m_arch.startsWith(QLatin1String("arm"));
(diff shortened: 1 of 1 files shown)


src/resultsdisassemblypage.cpp
```diff
@@ -573,11 +573,17 @@ void ResultsDisassemblyPage::showDisassembly()
...
 
+#ifndef __aarch64__
         if (m_arch.startsWith(QLatin1String("armv8")) || m_arch.startsWith(QLatin1String("aarch64"))) {
             return QStringLiteral("aarch64-linux-gnu-objdump");
         }
-        const auto isArm = m_arch.startsWith(QLatin1String("arm"));
-        return isArm ? QStringLiteral("arm-linux-gnueabi-objdump") : QStringLiteral("objdump");
+#endif
+#ifndef __arm__
+        if (m_arch.startsWith(QLatin1String("arm"))) {
```
milianw, line 582: I assume `__arm__` will be defined together with `__aarch64__`, right?

generally though, I wonder if we should instead just do something like this instead:

```
      if (m_arch.startsWith(QLatin1String("armv8")) || m_arch.startsWith(QLatin1String("aarch64"))) {
            if (auto aarch64Objdump = QStandardPaths::findExecutable(QStringLiteral("aarch64-linux-gnu-objdump")); !aarch64Objdump.isEmpty()))
                return aarch64Objdump;
        }
```

and dito for the other one - this should always work and is independent of the compile time macros? what do you think?

GitMensch:
Is there any reason to not merge that in its current state?

PR #726, opened 2025-05-26
fix: display disassembly correctly on aarch64

pcc:
On aarch64 (and possibly other architectures) objdump splits the instruction mnemonic from the operands with a tab so the existing code would only show the mnemonic. Change the code so that it ignores tabs after the first two so that the whole instruction is displayed.

### DIFF
--- a/src/models/disassemblyoutput.cpp
+++ b/src/models/disassemblyoutput.cpp
@@ -226,9 +226,16 @@ DisassemblyOutput::ObjectdumpOutput DisassemblyOutput::objdumpParse(const QByteA
         // [spaces]addr:\t [branch visualization] [hexdump]\tdiassembly
         // we can simplify parsing by splitting it into three parts
 
-        const auto parts = asmLine.split(QLatin1Char('\t'));
+        auto firstTab = asmLine.indexOf(QLatin1Char('\t'));
+        if (firstTab == -1) {
+            firstTab = asmLine.length();
+        }
+        auto secondTab = asmLine.indexOf(QLatin1Char('\t'), firstTab + 1);
+        if (secondTab == -1) {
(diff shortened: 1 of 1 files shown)


GitMensch:
Sounds good. Mind to share a screen before/after?

Note: there's ongoing work to use a disassembler library, so the fix may only be temporary...

pcc:
Sure, here is before/after

Screenshot from 2025-05-26 15-36-47
Screenshot from 2025-05-26 15-39-52

milianw (approved):
lgtm

milianw:
hm the test failures are related, can you please have a look at those?

```
FAIL!  : TestDisassemblyOutput::testParse(objdump2.txt) 'line.branchVisualisation.isEmpty()' returned FALSE. ()
```

milianw (changes requested):
see previous comment

pcc:
> hm the normal build should also include tests, I run those via `ctest --output-on-failure -j8` or similar myself. the CI clazy/clang-tidy jobs are sadly broken, please ignore those for now, but otherwise it should just work?

Figured it out. Right, the tests are built by default, but I was building with Nix which was disabling them unbeknownst to me. I figured out why they were failing (sent #729), sent them a PR to enable tests (https://github.com/NixOS/nixpkgs/pull/413058) and finally fixed the failing tests here.

> generally though, if you figure something out that you would have liked to see documented more clearly, by all means please create a PR to add the missing documentation after you figured it out!

I guess the only thing worth documenting is that `ctest` should be used to run tests, since this might not be obvious to everyone (some CMake using projects like LLVM have their own testing tools). Sent #730 for that.

milianw (approved):

PR #733, opened 2025-08-04
fix: formatTimeString ms format

mactul:
formatTimeString(123'456'789, false) was returning "123.123ms" instead of "123.456ms" because the µs part was missing a modulo.

This was resulting in wrong "On CPU Time" and "Off CPU Time" display for programs that executes in a little less than a second

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -294,34 +294,34 @@ QString Util::formatCostRelative(quint64 selfCost, quint64 totalCost, bool addPe
     return ret;
 }
 
-QString Util::formatTimeString(quint64 nanoseconds, bool shortForm)
+QString Util::formatTimeString(quint64 totalNanoseconds, bool shortForm)
 {
-    if (nanoseconds < 1000) {
-        return QString::number(nanoseconds) + QLatin1String("ns");
+    if (totalNanoseconds < 1000) {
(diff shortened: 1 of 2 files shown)


src/util.cpp
```diff
@@ -315,10 +315,11 @@ QString Util::formatTimeString(quint64 nanoseconds, bool shortForm)
 
     const auto milliseconds = (nanoseconds / 1000000) % 1000;
```
milianw, line 316: I think this is then equally wrong in the usage below in loc 340?

please move this variable into the conditional, and do the same for the `microseconds` above in loc 307 to ensure they are not misused. then you can use the same long name `microseconds` here.

milianw (changes requested):
great catch! can you please also add a unit test for this for various scales to verify we handle them properly? I.e. I suspect the issue you describe also hits us for the millis after seconds in the final fallback?

milianw (approved):
thanks a lot for your great contribution! much appreciated

milianw:
there's a minor clang-format issue, and could you also please simply squash your three commits into one? I don't consider these changes separable.

milianw (changes requested):
see comment above

mactul:
Oups, my bad for the clang-format.
I've fixed it and squashed everything !

milianw (approved):

PR #753, opened 2026-01-14
Update build instructions for Qt6 on Debian/Ubuntu

erijo:
Also add support for the QCustomPlot library name used on Debian.

### DIFF
--- a/cmake/FindQCustomPlot.cmake
+++ b/cmake/FindQCustomPlot.cmake
@@ -30,7 +30,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 # THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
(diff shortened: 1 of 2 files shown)


GitMensch (changes requested):

cmake/FindQCustomPlot.cmake
```diff
@@ -30,7 +30,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 # THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-find_library(QCustomPlot_LIBRARY NAMES qcustomplot-qt${QT_MAJOR_VERSION} qcustomplot)
+find_library(QCustomPlot_LIBRARY NAMES qcustomplot-qt${QT_MAJOR_VERSION} qcustomplot QCustomPlotQt${QT_MAJOR_VERSION})
```
milianw: for the future: these are two patches - one fixes the QCustomPlot library name as used on Debian, the other updates the build instructions.

I'll let it pass, but please think of such splits to make patches atomic

PR #754, opened 2026-01-14
fix: use full path in window title

erijo:
To make it easier to see which window is which when you have multiple hotspot windows showing different files all named perf.data.


GitMensch:
Not sure about that - why would you have a lot of perf.data instead of giving them names during recording?
But I guess that change wouldn't be "bad" either, so...

milianw (changes requested):

PR #758, opened 2026-01-29
feat(README): add section for rust's v0 mangling scheme

### LINKED ISSUE #751: Support rust `v0` mangling scheme
As far as i can tell it doesn't look like `hotspot` currently supports the soon-to-be default mangling scheme that rust will use: https://blog.rust-lang.org/2025/11/20/switching-to-v0-mangling-on-nightly/

It would be nice to be able to take advantage of the added info while parsing through `hotspot`.

`perf` parses and decodes the symbols as expected, so when I need to get some extra context I open with `perf report`, but this requires to recompile when I need different symbols. 

**Additional context**
This is what it looks like currently (compiled from `master`):
Image

RoloEdits:
As discussed in #751, this adds a section to the end of the `Using` section of the `README`. It adds some extra information to hopefully reduce some friction when using `hotspot` with rusts `v0` mangling scheme. This scheme is currently the default on `nightly`, and in the future will become the default for `stable`.

Closes: #751


crepererum (approved):
thank you

PR #764, opened 2026-03-01
feat: Drop Qt5 support

dantti:


dantti:
Qt6 libraries are included in the appimage, that what's usually done, I guess I should try to run it on an almalinux 8 then?

GitMensch:
Hm at least the appimage from the release has libQt6 and libKF6 - I'll check using it again

src/mainwindow.cpp
```diff
@@ -36,10 +36,8 @@
 #include <KShell>
 #include <KStandardAction>
 
-#include <kio_version.h>
-#if KIO_VERSION >= QT_VERSION_CHECK(5, 69, 0)
 #include <KIO/CommandLauncherJob>
-#endif
+#include <kio_version.h>
```
milianw: do we still need this?

milianw (approved):
lgtm, but potentially some more version checks or header includes could be cleaned up? happy to take that incrementally though.

cheers!

PR #776, opened 2026-06-10
let me help you to release static link hotspot

fifbroman8:
static link hotspot for different cpu


iamsergio:
For safety reasons we can't publish binaries built with your toolchain.

`This branch is 313 commits ahead of and 1771 commits behind buildroot/buildroot:master`

This is too much burden for us to audit, and even if it's safe now, we have no guarantees https://github.com/fifbroman8/buildroot/releases/download/v1.3.0/${{ matrix.cpu }}-buildroot-linux-musl_sdk-buildroot.tar.gz wouldn't change in the future, we'd need to host it ourselves.

I suggest we instead invest in vcpkg static builds.

PR #780, opened 2026-08-20
fall back to system path for hotspot-perfparser

GitMensch:
... if HOTSPOT_PERFPARSER not set and not found in libexec

this eases especially the use of static hotspot + hotspot-perfparser, as in that case those are the only two distributed elements

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -239,12 +239,17 @@ QString Util::findLibexecBinary(const QString& name)
 QString Util::perfParserBinaryPath()
 {
     auto parserBinary = QString::fromLocal8Bit(qgetenv("HOTSPOT_PERFPARSER"));
-    if (parserBinary.isEmpty()) {
-        parserBinary = Util::findLibexecBinary(QStringLiteral("hotspot-perfparser"));
-    } else {
-        parserBinary = QStandardPaths::findExecutable(parserBinary);
+    // prefer configured one (returns an error in the caller if invalid)
+    if (!parserBinary.isEmpty()) {
+        return QStandardPaths::findExecutable(parserBinary);
     }
-    return parserBinary;
+    // common option: find binary by configured libexec
+    parserBinary = Util::findLibexecBinary(QStringLiteral("hotspot-perfparser"));
+    if (!parserBinary.isEmpty()) {
+        return parserBinary;
+    }
+    // if both are empty: try system paths
+    return QStandardPaths::findExecutable(QStringLiteral("hotspot-perfparser"));
 }
 
 QString Util::formatString(const QString& input, bool replaceEmptyString)


fifbroman8:
Never mind. I will complie perfparser into hotspot. See https://github.com/fifbroman8/hotspot-bin/blob/main/demangle.patch

GitMensch:
I don't think that perfparser should be a GUI application also not in your fork.
his initializes X11 and other display envs - which is bad if you do a  `hotspot --exportTo your.perfparser perf.data` 

Also this PR is not only relevanbt for your fork but intended to easily use the binary in other cases as well.

GitMensch:
... also: I'd like to see a PR here that takes the "initial" demangler approach: using libiberty, when possible, and only if this isn't the case fallback to demangler plugins (not sure if the existing ones should be completely removed and libiberty made a hard dependency)

src/util.cpp
```diff
@@ -239,12 +239,17 @@ QString Util::findLibexecBinary(const QString& name)
...
-    if (parserBinary.isEmpty()) {
-        parserBinary = Util::findLibexecBinary(QStringLiteral("hotspot-perfparser"));
-    } else {
-        parserBinary = QStandardPaths::findExecutable(parserBinary);
+    // prefer configured one (returns an error in the caller if invalid)
+    if (!parserBinary.isEmpty()) {
+        return QStandardPaths::findExecutable(parserBinary);
+    }
+    // common option: find binary by configured libexec
+    parserBinary = Util::findLibexecBinary(QStringLiteral("hotspot-perfparser"));
```
milianw, line 247: the string literal won't be deduped by the compiler, so please introduce a local. since you'll then have to touch this PR anyhow, please consider changing the code in the following way, which I subjectively find nicer:

```
// prefer configured one (returns an error in the caller if invalid)
if (const auto perfparser = qEnvironmentVariable("HOTSPOT_PERFPARSER"); !perfparser.isEmpty()) {
    return QStandardPaths::findExecutable(perfparser);
}

const auto hotspotPerfparser = QStringLiteral("hotspot-perfparser");

// common option: find binary by configured libexec
if (const auto parserBinary = Util::findLibexecBinary(hotspotPerfparser); !parserBinary.isEmpty()) {
    return parserBinary;
}

// fallback: try system paths
return QStandardPaths::findExecutable(hotspotPerfparser);
```

milianw (changes requested):

fifbroman8:
> I don't think that perfparser should be a GUI application also not in your fork. his initializes X11 and other display envs - which is bad if you do a `hotspot --exportTo your.perfparser perf.data`
> 
> Also this PR is not only relevanbt for your fork but intended to easily use the binary in other cases as well.

When linking into one executable, perfparser should have GUI because some kf6 modules has auto-initialization and will crash with QCoreApplication only.

GitMensch:
> When linking into one executable, perfparser should have GUI because some kf6 modules has auto-initialization and will crash with QCoreApplication only.

Then I highly suggest to keep both binaries separate - and with this PR which already has the "ok feature-wise" from Milian it will be even easier to have both directly next to each other.

